PR #9, opened 2018-06-19
Revert "Remove unused targets openj9-{jdk,jre}-image"

groeges:
This reverts commit 6d29baf3d7557c280f0fcfc1aaedab46fdbb680f.

Signed-off-by: Steve Groeger GROEGES@uk.ibm.com

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -180,6 +180,33 @@ $(foreach file,$(OPENJ9_MANAGEMENT_LIBRARIES), \
 
 endef
 
+# generated_target_rules
+# ----------------------
+# param 1 = The make goal
+# param 2 = The jdk/jre directory to add openj9 content
+define generated_target_rules
+
+.PHONY : $1
+
+$(foreach file,$(OPENJ9_SHARED_LIBRARIES), \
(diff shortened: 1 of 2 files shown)


pshipton:
I expect @keithc-ca will make some comments about this.

keithc-ca:
Simply reverting this will leave things in a state where builds will fail due to the absence of the `jre-image` target which was a prerequisite of `openj9-jre-image` which in turn is required by `product-images`.

I suspect we only need to add whatever is missing to `generated_target_rules_build` or perhaps add the dependency `product-images : j9vm-build` in closed/Main.gmk.

groeges:
Even if I revert the changes the issue still occurs. 
Closing this PR and will open another one later when we know what the actual issue is.

keithc-ca:
I think https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/10 fixes the problem this tried to address.

andrew-m-leonard:
@keithc-ca one thing to verify once this is fixed up, is that jlink still works correctly...? the original targets for separate jre/jdk image targets was related to jlink not creating valid JREs.., it would be worth validating that still works too...

keithc-ca:
I did validate that with this fix, jlink yields a working package.

PR #15, opened 2018-07-09
Prevent the ClassLoader object being GCd in the middle of a method

adamfarley:
In ResourceBundle, the private method "getBundleFromModule" takes
a ClassLoader parameter, but doesn't appear to use it directly.
If a GC cycle occurs during the method execution, we can lose that
ClassLoader, causing a bug whose most visible symptom is us being
unable to find a locale-specific entity that definitely exists.

E.g. If we have Stuff_fr.class and Stuff_fr_CA.class, and we ask for
"Stuff" with the locale "fr, CA", then a garbage collection here
may cause us to return Stuff_fr.class (as if Stuff_fr_CA.class did not
exist).

Using the loader parameter for an operation at the very end of the
method prevents GC from collecting it, and prevents this problem.

Signed-off-by: Adam Farley <adam.farley@uk.ibm.com>

### DIFF
--- a/src/java.base/share/classes/java/util/ResourceBundle.java
+++ b/src/java.base/share/classes/java/util/ResourceBundle.java
@@ -25,7 +25,7 @@
 
 /*
  * (C) Copyright Taligent, Inc. 1996, 1997 - All Rights Reserved
- * (C) Copyright IBM Corp. 1996 - 1999 - All Rights Reserved
+ * (C) Copyright IBM Corp. 1996 - 2018 - All Rights Reserved
  *
  * The original version of this source code and documentation
  * is copyrighted and owned by Taligent, Inc., a wholly-owned
@@ -1696,7 +1696,12 @@ public abstract class ResourceBundle {
         Reference.reachabilityFence(callerModule);
         Reference.reachabilityFence(module);
 
-        return bundle;
+        //The OpenJ9 GC might collect the loader before we return here. This prevents that.
+        try {
+            return bundle;
+        } finally {
+            Reference.reachabilityFence(loader);
+        }
     }
 
     /**


src/java.base/share/classes/java/util/ResourceBundle.java
```diff
@@ -1696,7 +1696,13 @@ private static ResourceBundle getBundleImpl(Module callerModule,
         Reference.reachabilityFence(callerModule);
         Reference.reachabilityFence(module);
 
-        return bundle;
+        //The OpenJ9 GC might collect the loader before we return here. This prevents that.
+        try {
+            return bundle;
+        } finally {
+            //Should never be true. Using the loader here prevents it being GC'd.
+            if (baseName.equals(loader)) throw new Error("Unexpected error.");
```
pshipton: This could use == instead of .equals(), which should be a faster test.

DanHeidinga: This should use reachabilityFence in JDK9+: https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Reference.html#reachabilityFence-java.lang.Object-

Aka `keepAlive`

adamfarley: Incomparable object. Will try Dan's suggestion of reachabilityFence.

src/java.base/share/classes/java/util/ResourceBundle.java
```diff
@@ -1696,7 +1696,12 @@ private static ResourceBundle getBundleImpl(Module callerModule,
         Reference.reachabilityFence(callerModule);
         Reference.reachabilityFence(module);
 
-        return bundle;
+        //The OpenJ9 GC might collect the loader before we return here. This prevents that.
+        try {
+            return bundle;
+        } finally {
+            Reference.reachabilityFence(loader);
```
pshipton: Seems the reachabilityFence could be added above in the same place where its used for the modules, not sure there is any benefit to using try finally around the return statement.

adamfarley:
There's a chance this fix isn't needed for JDK10 and up. Verifying.

PR #23, opened 2018-08-13
Fix git tag pattern

keithc-ca:
Should search for Java 12 tags.
Also remove advice to 'contact support team'.


keithc-ca:
https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/25 is a better, longer-term solution.

PR #42, opened 2018-10-18
Allow configuring distinct compilers for OpenJ9

keithc-ca:
* `--with-openj9-cc`  : select C compiler
* `--with-openj9-cxx` : select C++ compiler
* `--with-openj9-developer-dir` : select Xcode version (MacOS)

This is a replay of https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/147.


DanHeidinga:
Merging as is as we don't have PR builds on this repo yet

PR #43, opened 2018-10-19
Add OSX Support

babsingh:
1) Set OPENJ9_BUILDSPEC to osx_x86-64_cmprssptrs for compressedrefs
build.

2) Setting MACOSX_DEPLOYMENT_TARGET environment variable will let OpenJ9
executables to run on OSX versions older than the OSX version on which
OpenJ9 was built. MACOSX_DEPLOYMENT_TARGET acts similar to 
-mmacosx-version-min=version compiler option. If both the compiler option is
specified and the environment variable is set, then the compiler option will take
precedence. Here, MACOSX_DEPLOYMENT_TARGET environment variable and
the compiler option will point to the same version. The environment
variable is defined to support dependencies where the compiler option
is not applied. Issue: eclipse/openj9#3244.

3) Append "-pagezero_size 0x1000" to LDFLAGS_JDKEXE on OSX 64-bit
compressedrefs build. This allows memory to be mapped below 4GB.

### DIFF
--- a/closed/autoconf/custom-hook.m4
+++ b/closed/autoconf/custom-hook.m4
@@ -249,7 +249,11 @@ AC_DEFUN_ONCE([OPENJ9_PLATFORM_SETUP],
       fi
     elif test "x$OPENJDK_BUILD_OS" = xmacosx; then
       OPENJ9_PLATFORM_CODE=oa64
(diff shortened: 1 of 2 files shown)


babsingh:
fyi - @pshipton @DanHeidinga

PR #46, opened 2018-11-09
WIP: Add new runtimes copyrightChecker jenkins script

groeges:
Signed-off-by: Steve Groeger <GROEGES@uk.ibm.com>


AdamBrousseau:
I took a quick scan, not too in depth. I wondered if this could be merged into the existing script in OpenJ9? I also noticed you added some ignore logic. Which maybe could be combined with Eclipse/openj9#1879?
Thoughts?

PR #47, opened 2018-11-12
Adapt to recent Java 12 changes

keithc-ca:
Requires https://github.com/eclipse/openj9/pull/3638

* openj9 now produces a redirector that is right for the current build; no need to reference a version-specific copy
* adopt the new JCL library for Java 12
* put the static JVM library where openjdk expect it to be and remove the (now unnecessary) adjustments to `LDFLAGS_JDKLIB`
* disable CDS archive generation


closed/OpenJ9.gmk
```diff
@@ -208,14 +208,15 @@ $(call openj9_add_jdk_shlibs, java.base, \
...
 	)
 
 ifeq (windows,$(OPENJDK_TARGET_OS))
-$(eval $(call openj9_add_jdk_rules, \
-	$(OUTPUTDIR)/vm/lib/$(call STATIC_LIBRARY,jvm), \
-	$(OUTPUTDIR)/vm/j9vm_jdk12/$(call STATIC_LIBRARY,jdk12_jvm)))
+$(foreach subdir, j9vm server, \
+	$(eval $(call openj9_add_jdk_rules, \
+		$(JDK_OUTPUTDIR)/bin/$(subdir)/$(call STATIC_LIBRARY,jvm), \
+		$(OUTPUTDIR)/vm/j9vm_jdk12/$(call STATIC_LIBRARY,jdk12_jvm))))
```
pshipton: Should this be jvm instead of jdk12_jvm?

keithc-ca: No, the redirectors are all the same after https://github.com/eclipse/openj9/pull/3638, but the main JVM libraries are still version-specific.

pshipton:
nope, lgtm

PR #63, opened 2019-01-09
Add permissions required by module openj9.gpu & openj9.cuda

JasonFengJ9:
<details><summary>Initial PR description (for background info)</summary>
<p>
Add `openj9.gpu` to boot module list

This allows module `openj9.gpu` to access `java.base/com.ibm.gpu.spi.GPUAssist.Provider`.

closes: https://github.com/eclipse/openj9/issues/3213

`java.base/java.util.GPUAssistHolder` attempts to load `openj9.gpu/com.ibm.gpu.internal.CudaGPUAssistProvider` as per following stacktrace:
```
	at java.base/java.util.ServiceLoader.loadProvider(ServiceLoader.java:862)
	at java.base/java.util.ServiceLoader$ModuleServicesLookupIterator.hasNext(ServiceLoader.java:1076)
	at java.base/java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1299)
	at java.base/java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1384)
	at java.base/java.util.GPUAssistHolder.gpuAssist(GPUAssistHolder.java:33)
	at java.base/java.util.GPUAssistHolder.<clinit>(GPUAssistHolder.java:28)
	at java.base/java.util.Arrays.sort(Arrays.java:226)
```
`openj9.gpu/com.ibm.gpu.internal.CudaGPUAssistProvider` is loaded via reflection but doesn't have permission `accessClassInPackage.com.ibm.gpu.spi` as per following stacktrace:
```
Caused by: java.security.AccessControlException: Access denied ("java.lang.RuntimePermission" "accessClassInPackage.com.ibm.gpu.spi")
	at java.base/java.security.AccessController.throwACE(AccessController.java:176)
	at java.base/java.security.AccessController.checkPermissionHelper(AccessController.java:237)
	at java.base/java.security.AccessController.checkPermission(AccessController.java:373)
	at java.base/java.lang.SecurityManager.checkPermission(SecurityManager.java:322)
	at java.base/java.lang.SecurityManager.checkPackageAccess(SecurityManager.java:1238)
	at java.base/java.lang.J9VMInternals$2.run(J9VMInternals.java:313)
```
This is because module `openj9.gpu` is not a boot module hence can't access `java.base/com.ibm.gpu.spi`. 

Manually verified that this PR passes the test in https://github.com/eclipse/openj9/issues/3213
</p>
</details>
<br><br>

Add explicit permissions required by modules `openj9.gpu` and `openj9.cuda`

Modules `openj9.gpu` and `openj9.cuda` are not boot modules and require these permissions.
Otherwise an `java.security.AccessControlException` might be thrown as following:
```
Exception in thread "main" java.util.ServiceConfigurationError: com.ibm.gpu.spi.GPUAssist$Provider: Unable to load com.ibm.gpu.internal.CudaGPUAssistProvider
	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:581)
	at java.base/java.util.ServiceLoader.loadProvider(ServiceLoader.java:862)
         ...
	at java.base/java.util.GPUAssistHolder.gpuAssist(GPUAssistHolder.java:34)
	at java.base/java.util.GPUAssistHolder.<clinit>(GPUAssistHolder.java:31)
	at java.base/java.util.Arrays.sort(Arrays.java:152)
         ...
Caused by: java.security.AccessControlException: Access denied ("java.lang.RuntimePermission" "accessClassInPackage.com.ibm.gpu.spi")
	at java.base/java.security.AccessController.throwACE(AccessController.java:176)
	at java.base/java.security.AccessController.checkPermissionHelper(AccessController.java:237)
``` 

Note: `JTReg java/util/concurrent/tck/JSR166TestCase.java` won't be fixed by this PR and will be excluded instead as per discussion within this PR conversation.

Reviewer: @DanHeidinga 
FYI: @pshipton 

### DIFF
--- a/src/java.base/share/lib/security/default.policy
+++ b/src/java.base/share/lib/security/default.policy
@@ -198,6 +198,20 @@ grant codeBase "jrt:/jdk.zipfs" {
     permission java.util.PropertyPermission "os.name", "read";
 };
 
+grant codeBase "jrt:/openj9.cuda" {
+    permission java.util.PropertyPermission "com.ibm.oti.vm.library.version", "read";
+    permission java.lang.RuntimePermission "loadLibrary.cuda4j29";
+};
+
+grant codeBase "jrt:/openj9.gpu" {
+    permission java.lang.RuntimePermission "accessClassInPackage.com.ibm.gpu.spi";
+    permission com.ibm.gpu.GPUPermission "access";
+    permission java.util.PropertyPermission "com.ibm.gpu.verbose", "read";
+    permission java.util.PropertyPermission "com.ibm.gpu.enforce", "read";
+    permission java.util.PropertyPermission "com.ibm.gpu.enable", "read";
+    permission java.util.PropertyPermission "com.ibm.gpu.disable", "read";
+};
+
 // permissions needed by applications using java.desktop module
 grant {
     permission java.lang.RuntimePermission "accessClassInPackage.com.sun.beans";


pshipton:
We can grant the openj9.gpu module the `accessClassInPackage.com.ibm.gpu.spi` permission and anything else it needs. This is the preferred approach. See the lib/security/default.policy file.

@keithc-ca fyi

JasonFengJ9:
Will find out a complete list of permission required by this test (tried before, there were quite a few).

JasonFengJ9:
Aha, okay, will do then. 
Just for what it worths, here was the permission identified.
```
grant codeBase "jrt:/openj9.gpu" {
    permission java.lang.RuntimePermission "accessClassInPackage.com.ibm.gpu.spi";
    permission com.ibm.gpu.GPUPermission "access";
    permission java.util.PropertyPermission "com.ibm.gpu.verbose", "read";
    permission java.util.PropertyPermission "com.ibm.gpu.enforce", "read";
    permission java.util.PropertyPermission "com.ibm.gpu.enable", "read";
    permission java.util.PropertyPermission "com.ibm.gpu.disable", "read";
//    permission java.security.AllPermission;
};
```

pshipton:
The shorter list is preferred over AllPermission, and seems a reasonable list.

pshipton:
Also note, once this is finalized, the https://github.com/ibmruntimes/openj9-openjdk-jdk12/ repo will also need the fix.

JasonFengJ9:
Just realize couple of problems:
1. Any modification or addition of security related property/permission within module `openj9.gpu` will break a hard-coded list like above;
2. Application could provide its own security policy just like eclipse/openj9#3213 and it still fails with modified `default.policy` (I was running a smaller standalone testcase without custom security policy and passed with `default.policy`).

Hence existing PR is a way without application influence.

pshipton:
1. is correct. We'll just need to keep the list of permissions up to date. Since the JTREG test will fail, that will be a good reminder.

For 2 I'm surprised the default.policy isn't still in effect even when a custom policy is provided. The default.policy is needed for all the modules included there to work. There are two policies in affect normally, the default.policy and the user modifiable conf/security/java.policy. Perhaps we can fix the test and contribute the fix back.

JasonFengJ9:
There are two ways to supply a custom security policy, one way is to append to existing policy settings, the other way is to override existing version. Test result appears JTReg performs override though I have to dig further to confirm it.
For user overriding the policy, one argument will be that application takes the responsibility.

keithc-ca:
This puzzles me: java.base says `exports com.ibm.gpu.spi to openj9.gpu;` so openj9.gpu should have access to that package. Has the meaning of `exports` changed?

JasonFengJ9:
Re: _openj9.gpu should have access to that package_
This `java.security.AccessControlException` is not about the module `openj9.gpu` accesses the package `com.ibm.gpu.spi`. 
The code within `openj9.gpu` is not part of boot module list, and is loaded with less privileged ProtectionDomain (along with permissions)  hence doesn't have permission to access the code within protected packaged `com.ibm.gpu.spi` within module `java.base` even though the code in question is in package `com.ibm.gpu` (from module `openj9.gpu`).
This is about the permissions (in form of `ProtectionDomain`) assigned to a `CodeSource`.
Putting module `openj9.gpu` into boot module list or assigning it Permissions required (or `AllPermission`) in a `policy` file will grand `openj9.gpu` permissions needed.

pshipton:
I think Keith's question is a good one. However it does seem to be expected behavior. Looking at lib/security/default.policy, java.xml.crypto has permission to access sun.security.util which is also a packaged exported to this module.

I think we need a better reason to add openj9.gpu to the boot modules list other than passing a test which is likely somewhat bogus. The reason to not use a boot module is to avoid unexpected security vulnerabilities. I think we should add the explicit permissions required, and either modify the test to work properly, or exclude this test and write a new one that will fail if openj9.gpu starts needing additional permissions to work.

JasonFengJ9:
Note: this `AccessControlException` can be reproduced via a call to `Arrays.sort()` which invokes `GPUAssistHolder.instance.trySort()` when a `SecurityManager` is installed. 

Regarding the approach _add the explicit permissions required, and either modify the test to work properly, or exclude this test and write a new one that will fail if openj9.gpu starts needing additional permissions to work_, my concerns are:
1. When application overrides JVM existing `policy` (and install a `SecurityManager`), the user has to add these `openj9.gpu` specific permissions to the custom `policy` file otherwise an `AccessControlException` could be thrown. This presents a behaviour difference between `OpenJ9` and `RI`;
2. As per https://openjdk.java.net/jtreg/tag-spec.html about `policy=<file>` within `JTReg` tests, it states "In particular, the system security policy and the user's security policy will be **ignored**" . This means more `JTReg` tests in the future could be excluded due to this behaviour difference. 

Giving these consideration, my feeling is that adding module `openj9.gpu` to the boot module list is a better approach.

pshipton:
If the default.policy is overridden, this means all the modules that have permissions in this file will no longer work when the Security Manger is enabled. I don't believe this is sufficient reason to add openj9.gpu to the boot module list.

JasonFengJ9:
Re: _If the default.policy is overridden, this means all the modules that have permissions in this file will no longer work when the Security Manger is enabled._
This is not true if the custom policy adds permissions the application (or JTreg tests) required. The custom policy file could add module `openj9.gpu` specific permissions as well however this probably won't happen for `JTreg` tests.
This is about a behaviour difference between `OpenJ9` and `RI`, is it necessary? might it cause some `OpenJ9` _customer defect_ down the road?

Having said that, I am fine to modify this PR to take the approach @pshipton recommended, i.e., _add the explicit permissions required, and either modify the test to work properly, or exclude this test and write a new one that will fail if openj9.gpu starts needing additional permissions to work_.
We can re-visit this decision if same `AccessControlException` occurs in the future.

JasonFengJ9:
Just a quick note:
The initial `AccessControlException` reported within https://github.com/eclipse/openj9/issues/3213 blocked the actual JTReg test `java/util/concurrent/tck/JSR166TestCase.java` which has to be excluded due to this `OpenJ9` and `RI` difference.
I assume the new added test is to ensure `openj9.gpu` works with permissions introduced into `java.policy` to prevent regression if there are additional permissions required by `openj9.gpu`.
This PR won't duplicate the test coverage of `java/util/concurrent/tck/JSR166TestCase.java`.

pshipton:
Agree with the previous comment. What is involved in fixing the JSR166 test case? Does the test pass once it gets past the permission issues?

JasonFengJ9:
`JSR166` test passes when the permission issue is resolved. There was no other fix required for this testcase besides this PR.

pshipton:
There is no jdk testing atm. jdk11 tested in https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/89, and this change is the same.

PR #107, opened 2019-05-13
Rework static initialization of useNativeChaCha20Cipher

keithc-ca:
* use a helper method so the field can be final


pshipton:
jenkins compile plinux jdknext

PR #135, opened 2019-09-18
Set OpenJ9 ProdDefns requires class

sophia-guo:
Some hotspot tests apply to Openj9, which we'd like to run them regularly in test builds or run manually. In that case unset Hotspot VMProps ProdDefns requires class and replace with optional OpenJ9 class is necessary.

Signed-off-by: Sophia Guo sophiag@ca.ibm.com

### DIFF
--- a/test/hotspot/jtreg/TEST.ROOT
+++ b/test/hotspot/jtreg/TEST.ROOT
@@ -71,6 +71,9 @@ requires.properties= \
     docker.support \
     test.vm.gc.nvdimm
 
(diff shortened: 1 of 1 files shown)


sophia-guo:
This is the same PR for openj9-openjdk-jdk11 ibmruntimes/openj9-openjdk-jdk11#208
Problematic test build: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/538/
Fix build: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/539/

PR #149, opened 2019-10-25
Fix LUDCL optimization

theresa-m:
Port from ibmruntimes/openj9-openjdk-jdk8#341
Doc issue: eclipse/openj9-docs#413

The assumption that the ludcl will not change during a call to readObject/readUnshared can only be made if user code is not invoked. This change triggers a refresh of the cachedLudcl when a user method is invoked.
Re-enable ludcl caching by default (com.ibm.enableClassCaching)

Signed-off-by: Theresa Mammarella Theresa.T.Mammarella@ibm.com

### DIFF
--- a/src/java.base/share/classes/java/io/ObjectInputStream.java
+++ b/src/java.base/share/classes/java/io/ObjectInputStream.java
@@ -336,12 +336,12 @@ public class ObjectInputStream
     }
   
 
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Jenkins test sanity zlinux jdk

PR #157, opened 2020-01-04
Add jitserver launcher in OpenJ9 builds

ashu-mehra:
If --enable-jitserver configuration option is set, then enable
building of jitserver launcher and include it in the jdk.

Signed-off-by: Ashutosh Mehra <mehra.ashutosh@ibm.com>


ashu-mehra:
I guess this PR can now be merged.

PR #189, opened 2020-04-08
Fix J9DDR compile commands

keithc-ca:
This is a replay of ibmruntimes/openj9-openjdk-jdk11#291 for Java next.


pshipton:
Testing in https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/291 is sufficient for this identical change.

PR #199, opened 2020-04-30
Fix DDR compile error due to ibm.jzos

keithc-ca:
A replay of ibmruntimes/openj9-openjdk-jdk11#302 for jdknext.


pshipton:
Tested by https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/302

PR #201, opened 2020-05-02
JAVAC_WARNINGS is replaced by JAVA_WARNINGS_ARE_ERRORS

pshipton:
Must disable -Werror for OpenJ9 to compile with javac.

See https://github.com/eclipse/openj9/issues/9328#issuecomment-622637910

Note this is a merge to the openj9-staging branch.


closed/custom/common/JavaCompilation.gmk
```diff
@@ -18,5 +18,5 @@
 # 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
 # ===========================================================================
 
-# OpenJ9 requires more lenient javac warnings
-JAVAC_WARNINGS := -Xlint:-removal
+# OpenJ9 requires allowing warnings
+JAVA_WARNINGS_ARE_ERRORS :=
```
keithc-ca, line 22: Perhaps it would be better to put this (temporarily) in `custom-spec.gmk.in` until we can fix those warnings, then there would be no need to modify `make/common/JavaCompilation.gmk`.

pshipton: It doesn't work. I tried https://github.com/ibmruntimes/openj9-openjdk-jdk/compare/openj9-staging...pshipton:werror2?expand=1

pshipton: I hadn't reconfigured, it's working after doing so. I'll update the PR.

pshipton:
Updated and rebased.

keithc-ca:
This doesn't cover all the changes we'll need to make, but it's a start and I don't see any reason to hold this part back.

PR #202, opened 2020-05-06
Update makefiles for recent openjdk changes

keithc-ca:
* Restore customization post hooks removed by:
  - 8244044: Refactor phase makefiles to be structured per module 
* Adapt organization of makefiles for:
  - 8244044: Refactor phase makefiles to be structured per module 
*  Rework how `jdk.jcmd` launchers are built for openj9.
* Ignore java compile warnings only in certain modules.

These changes together with eclipse/openj9#9292 and eclipse/openj9#9471,
and using the `--with-build-jdk` configure option specifying a jdk15 re-enable building jdknext. The resulting image doesn't run: further work anticipated by eclipse/openj9#9328 will be necessary.

Note this is a merge to the openj9-staging branch.


pshipton:
Why is jdk15 needed as the boot jdk instead of jdk14?

keithc-ca:
> Why is jdk15 needed as the boot jdk instead of jdk14?

The _boot_ jdk can be jdk14: it's the _build_ jdk that must be 15 (else it will try to use the VM being built which does not yet work).

pshipton:
@hangshao0 fyi. Although I expect you are better off sticking with an older version than trying to chase the latest.

PR #215, opened 2020-06-24
Do not use rebind in StringConcatFactory

babsingh:
OpenJ9 does not support MethodHandle.rebind. So, its usage has been temporarily
removed from StringConcatFactory until OpenJ9 supports OpenJDK MethodHandles.

Fixes: eclipse/openj9#9771

Identical to https://github.com/ibmruntimes/openj9-openjdk-jdk15/pull/4.


babsingh:
fyi @pshipton

PR #229, opened 2020-08-31
Include openj9.dataaccess in JRE

keithc-ca:
See: eclipse/openj9#10030.

### DIFF
--- a/closed/custom/Images-pre.gmk
+++ b/closed/custom/Images-pre.gmk
@@ -18,4 +18,8 @@
 # 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
 # ===========================================================================
 
-JRE_MODULES += openj9.cuda openj9.gpu
+JRE_MODULES += \
+	openj9.cuda \
+	openj9.dataaccess \
+	openj9.gpu \
+	#


closed/custom/Images-pre.gmk
```diff
@@ -18,4 +18,8 @@
 # 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
 # ===========================================================================
 
-JRE_MODULES += openj9.cuda openj9.gpu
+JRE_MODULES += \
+	openj9.cuda \
+	openj9.dataaccess \
+	openj9.gpu \
+	#
```
DanHeidinga, line 25: Why the final separator and `#` line?

keithc-ca: It's a pattern common in openjdk. I think the motivation is that to add to the list you don't have to change any existing lines.

DanHeidinga:
My assumption is this won't be tested by a PR build as it relates to a JRE package - can you confirm that?  If that's the case, I won't bother with PR builds

keithc-ca:
That's my understanding as well.

PR #252, opened 2020-11-19
Add OpenJ9 modules as platform modules

pshipton:
Referring to https://openjdk.java.net/jeps/261 modules which are part of
the java platform, and not tools, should be platform modules. In
particular this is a problem for the ibm.jzos platform module which
requires openj9.dataaccess.

Signed-off-by: Peter Shipton <Peter_Shipton@ca.ibm.com>


keithc-ca (approved):

keithc-ca:

PR #256, opened 2020-12-08
Do not rely on usingSharedClasses() to add url to loaderURLs

hangshao0:
When reaching URLClassPath.getLoader() before module system is
initialized, usingSharedClasses() always returns false. It is
likely that SCC support is not initialized for BuiltinClassLoaders
yet, which is done during module system initialization. We still need
to update loaderURLs in this case.

issue eclipse/openj9#11342

Signed-off-by: Hang Shao <hangshao@ca.ibm.com>


pshipton:
jenkins compile zlinux jdknext

pshipton (approved):

PR #265, opened 2021-01-11
Export jdk.internal.org.objectweb.asm for ClassGenerator

keithc-ca:
Required when using a Java 16+ boot jdk (where the default `--illegal-access=deny` applies).
See eclipse/openj9#11613.


pshipton:
I won't bother with PR testing here since this is the same change as https://github.com/ibmruntimes/openj9-openjdk-jdk16/pull/5

pshipton (approved):

PR #266, opened 2021-01-12
Enhancing GCM ByteBuffer Implementation

samasri:
Fixing [OpenJ9's Issue 11390](https://github.com/eclipse/openj9/issues/11390). Specifically, some of the AEAD tests were failing in the jtreg test harness. This PR fixes them by:

* Adding necessary functions to `NativeGaloisCounterMode` to support processing ByteBuffer input and output.
* Editing the jtreg test _GCMBufferTest.java_ to not check for the output length after `cipher.update`. This is because `cipher.update` doesn't have to return the encrypted input length. In our JDK, the `cipher.update` behaves differently depending on whether the machine used does have native crypto or not:
   * If native crypto is used, the `cipher.update` processes the input by adding it to a buffer, which will be processed when `cipher.doFinal` is called. Hence, the `cipher.update` returns zero.
   * However, when native crypto is not used, `cipher.update` returns 16.

### DIFF
--- a/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
+++ b/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2019 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2021 All Rights Reserved
  * ===========================================================================
  */
 
@@ -36,7 +36,9 @@ import java.security.*;
 import javax.crypto.*;
 import static com.sun.crypto.provider.AESConstants.AES_BLOCK_SIZE;
 import com.sun.crypto.provider.AESCrypt;
+import sun.security.util.ArrayUtil;
 
+import java.nio.ByteBuffer;
 import jdk.crypto.jniprovider.NativeCrypto;
 
 /**
@@ -99,6 +101,9 @@ final class NativeGaloisCounterMode extends FeedbackCipher {
     }
 
     private static void checkDataLength(int processed, int len) {
+        if (len < 0) {
+            throw new ProviderException("input size cannot be negative");
+        }
         if (processed > MAX_BUF_SIZE - len) {
             throw new ProviderException("SunJCE provider only supports " +
                 "input size up to " + MAX_BUF_SIZE + " bytes");
@@ -298,10 +303,6 @@ final class NativeGaloisCounterMode extends FeedbackCipher {
      * @return the number of bytes placed into the <code>out</code> buffer
      */
     int encrypt(byte[] in, int inOfs, int len, byte[] out, int outOfs) {
-        if ((len % blockSize) != 0) {
-            throw new ProviderException("Internal error in input buffering");
-       }
-
         checkDataLength(ibuffer_enc.size(), len);
 
         if (len > 0) {
@@ -312,6 +313,22 @@ final class NativeGaloisCounterMode extends FeedbackCipher {
         }
 
         return 0;
+    }
+
+    /*
+     * This method is for CipherCore to insert the remainder of its buffer
+     * into ibuffer_enc before a encryptFinal(ByteBuffer, ByteBuffer) operation
+     */
+    int encrypt(byte[] in, int inOfs, int len) {
+        // store internally until encryptFinal
+        ArrayUtil.nullAndBoundsCheck(in, inOfs, len);
+        if (len > 0) {
+            if (ibuffer_enc == null) {
+                ibuffer_enc = new ByteArrayOutputStream();
+            }
+            ibuffer_enc.write(in, inOfs, len);
+        }
+        return len;
     }
 
     /**
@@ -363,6 +380,46 @@ final class NativeGaloisCounterMode extends FeedbackCipher {
     }
 
     /**
+     * Performs encryption operation for the last time.
+     *
+     * @param src the input buffer with the data to be encrypted
+     * @param dst the output buffer with encrypted data
+     * @return the number of bytes placed into the <code>dst</code> buffer
+     */
+    int encryptFinal(ByteBuffer src, ByteBuffer dst)
+        throws IllegalBlockSizeException, ShortBufferException {
+
+        // Get array from source
+        byte[] src_arr = getBbArray(src);
+        int src_ofs = src.hasArray() ? (src.position() + src.arrayOffset()) : 0;
+
+        // Get array from destination
+        byte[] dst_arr;
+        int dst_offset;
+        if (dst.hasArray()) {
+            dst_arr = dst.array();
+            dst_offset = dst.position() + dst.arrayOffset();
+        } else {
+            dst_arr = new byte[dst.remaining()];
+            dst_offset = 0;
+        }
+
+        int len = encryptFinal(src_arr, src_ofs, src.remaining(), dst_arr, dst_offset);
+
+        // Advance source buffer position
+        src.position(src.limit());
+
+        // Update destination buffer
+        if (dst.hasArray()) {
+            dst.position(dst.position() + len);
+        } else {
+            dst.put(dst_arr, 0, len);
+        }
+
+        return len;
+    }
+
+    /**
      * Performs decryption operation.
      *
      * <p>The input cipher text <code>in</code>, starting at
@@ -380,10 +437,6 @@ final class NativeGaloisCounterMode extends FeedbackCipher {
      * @return the number of bytes placed into the <code>out</code> buffer
      */
     int decrypt(byte[] in, int inOfs, int len, byte[] out, int outOfs) {
-        if ((len % blockSize) != 0) {
-            throw new ProviderException("Internal error in input buffering");
-       }
-
         checkDataLength(ibuffer.size(), len);
 
         if (len > 0) {
@@ -416,16 +469,18 @@ final class NativeGaloisCounterMode extends FeedbackCipher {
                      byte[] out, int outOfs)
         throws IllegalBlockSizeException, AEADBadTagException,
         ShortBufferException {
-        if (len < tagLenBytes) {
+        if (len < 0) {
+            throw new ProviderException("Input length is negative");
+        }
+        if (len < tagLenBytes - ibuffer.size()) {
             throw new AEADBadTagException("Input too short - need tag");
         }
+        if (len > MAX_BUF_SIZE - ibuffer.size()) {
+            throw new ProviderException("SunJCE provider only supports "
+                + "a positive input size up to " + MAX_BUF_SIZE + " bytes");
+        }
 
-        // do this check here can also catch the potential integer overflow
-        // scenario for the subsequent output buffer capacity check.
-        checkDataLength(ibuffer.size(), (len - tagLenBytes));
-
-
-        if (out.length - outOfs < ((ibuffer.size() + len) - tagLenBytes)) {
+        if (out.length - outOfs < len  + ibuffer.size() - tagLenBytes) {
             throw new ShortBufferException("Output buffer too small");
         }
 
@@ -458,6 +513,69 @@ final class NativeGaloisCounterMode extends FeedbackCipher {
 
 
         return ret;
+    }
+
+
+    /**
+     * Performs decryption operation for the last time.
+     *
+     * @param src the input buffer with the data to be decrypted
+     * @param dst the output buffer with the decrypted data
+     * @return the number of bytes placed into the <code>dst</code> buffer
+     */
+    int decryptFinal(ByteBuffer src, ByteBuffer dst)
+        throws IllegalBlockSizeException, AEADBadTagException,
+        ShortBufferException {
+
+        // Get array from source
+        byte[] src_arr = getBbArray(src);
+        int src_ofs = src.hasArray() ? (src.position() + src.arrayOffset()) : 0;
+
+        // Get array from destination
+        byte[] dst_arr;
+        int dst_offset;
+        if (dst.hasArray()) {
+            dst_arr = dst.array();
+            dst_offset = dst.position() + dst.arrayOffset();
+        } else {
+            dst_arr = new byte[dst.remaining()];
+            dst_offset = 0;
+        }
+
+        int len = decryptFinal(src_arr, src_ofs, src.remaining(), dst_arr, dst_offset);
+
+        // Advance source buffer position
+        src.position(src.limit());
+
+        // Update destination buffer
+        if (dst.hasArray()) {
+            dst.position(dst.position() + len);
+        } else {
+            dst.put(dst_arr, 0, len);
+        }
+
+        return len;
+    }
+
+    /**
+     * Gets the byte array behind a buffer.
+     * Tries to use ByteBuffer.array(). If this is not available, the function uses ByteBuffer.get()
+     *
+     * @param src the buffer whose byte[] is needed
+     * @return the byte array with the buffer's content
+     */
+    private static byte[] getBbArray(ByteBuffer src) {
+        byte[] arr;
+        if (src.hasArray()) {
+            arr = src.array();
+        } else {
+            ByteBuffer cpy = src.duplicate();
+            arr = new byte[cpy.remaining()];
+            if (arr.length > 0) {
+                cpy.get(arr, 0, arr.length);
+            }
+        }
+        return arr;
     }
 
     // return tag length in bytes
--- a/src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
+++ b/src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2019 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2021 All Rights Reserved
  * ===========================================================================
  */
 
@@ -1308,7 +1308,11 @@ final class CipherCore {
             return cipher.decryptFinal(src, dst);
         } else {
             if (buffered > 0) {
-                ((GaloisCounterMode)cipher).encrypt(buffer, 0, buffered);
+                if (useNativeGCM) {
+                    ((NativeGaloisCounterMode) cipher).encrypt(buffer, 0, buffered);
+                } else {
+                    ((GaloisCounterMode) cipher).encrypt(buffer, 0, buffered);
+                }
             }
             return cipher.encryptFinal(src, dst);
         }
--- a/test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
+++ b/test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
@@ -20,6 +20,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+ * ===========================================================================
+ */
 
 /*
  * @test
@@ -62,7 +67,9 @@ public class GCMBufferTest implements Cloneable {
     int[] sizes;
     boolean incremental = false;
     // In some cases the theoretical check is too complicated to verify
-    boolean theoreticalCheck;
+    boolean theoreticalCheckFinal;
+    // This variable is never set to true in order to disable length verification after cipher.update for GCM Buffer Tests
+    boolean theoreticalCheckUpdate;
     List<Data> dataSet;
     int inOfs = 0, outOfs = 0;
 
@@ -128,7 +135,8 @@ public class GCMBufferTest implements Cloneable {
     GCMBufferTest(String algo, List<dtype> ops) {
         this.algo = algo;
         this.ops = ops;
-        theoreticalCheck = true;
+        theoreticalCheckUpdate = false;
+        theoreticalCheckFinal = true;
         dataSet = datamap.get(algo);
     }
 
@@ -416,7 +424,7 @@ public class GCMBufferTest implements Cloneable {
                     default -> throw new Exception("Unknown op: " + v.name());
                 }
 
-                if (theoreticalCheck) {
+                if (theoreticalCheckUpdate) {
                     pbuflen += plen - rlen;
                     if (encrypt && rlen != theoreticallen) {
                         throw new Exception("Wrong update return len (" +
@@ -473,7 +481,7 @@ public class GCMBufferTest implements Cloneable {
                     default -> throw new Exception("Unknown op: " + v.name());
                 }
 
-                if (theoreticalCheck && rlen != olen - outOfs) {
+                if (theoreticalCheckFinal && rlen != olen - outOfs) {
                     throw new Exception("Wrong doFinal return len (" +
                         v.name() + "):  " + "rlen=" + rlen +
                         ", expected output len=" + (olen - outOfs));
@@ -569,7 +577,7 @@ public class GCMBufferTest implements Cloneable {
                 }
 
                 // Check that the theoretical return value matches the actual.
-                if (theoreticalCheck && encrypt && rlen != theorticallen) {
+                if (theoreticalCheckUpdate && encrypt && rlen != theorticallen) {
                     throw new Exception("Wrong update return len (" +
                         v.name() + "):  " + "rlen=" + rlen +
                         ", expected output len=" + theorticallen);


closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -348,7 +372,7 @@ int encryptFinal(byte[] in, int inOfs, int len, byte[] out, int outOfs)
         len = in.length;
         ibuffer_enc.reset();
 
-        byte[] aad = ((aadBuffer == null || aadBuffer.size() == 0) ? emptyAAD : aadBuffer.toByteArray());
+        byte[] aad = ((aadBuffer.size() == 0 || aadBuffer == null) ? emptyAAD : aadBuffer.toByteArray());
```
pshipton, line 375: Calling `aadBuffer.size()` before `aadBuffer == null` may cause NPE.

src/java.base/share/classes/com/sun/crypto/provider/FeedbackCipher.java
```diff
@@ -140,6 +140,11 @@ abstract void init(boolean decrypting, String algorithm, byte[] key,
      */
     abstract int encrypt(byte[] plain, int plainOffset, int plainLen,
                          byte[] cipher, int cipherOffset);
+
+    int encrypt(byte[] plain, int plainOffset, int length) {
```
pshipton, line 144: Is it really necessary to modify this class and add this?

test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
```
pshipton, line 2: The Oracle copyright shouldn't be modified, add an IBM copyright below.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -99,7 +101,7 @@
     }
 
     private static void checkDataLength(int processed, int len) {
-        if (processed > MAX_BUF_SIZE - len) {
+        if (processed + len > MAX_BUF_SIZE) {
```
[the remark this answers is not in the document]
samasri: This is because `MAX_BUF_SIZE` is set to `Integer.MAX_VALUE`. Hence, when `len` is negative, `MAX_BUG_SIZE + len` overflows, which triggers the if statement and throws an exception. Hence, I thought it would make sense to move the `len` calculation to the other side.

A side note here: `len` can be negative because of [this line](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/d328410ecd9a3ea77cdd67f97ea69d96842ac6dc/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java#L487). When `len` (in the referenced line) is zero, the second argument passed to `checkDataLength` is negative.

However, looking at your comment, it makes sense that `processed + len` can also overflow. In this case, I think it's best to add a check for negative values to handle this case. So that would become:

```java
if(len < 0) return;
if (processed > MAX_BUF_SIZE - len) throw new ProviderException("...");
```

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: I reverted this change.

test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
samasri: I reverted this change.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2021, Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
samasri: I reverted this change.

src/java.base/share/classes/com/sun/crypto/provider/FeedbackCipher.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2021, Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
samasri: I reverted this change.

src/java.base/share/classes/com/sun/crypto/provider/FeedbackCipher.java
samasri: That was necessary at some point, however I tried it now and it seems to compile fine without it. Hence, I removed it.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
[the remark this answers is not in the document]
samasri: @pshipton Please let me know if you have any concerns about this new implementation (I updated the branch with the fix).

pshipton:
Pls update the IBM copyrights in the files changed.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -416,7 +475,7 @@ int decryptFinal(byte[] in, int inOfs, int len,
                      byte[] out, int outOfs)
         throws IllegalBlockSizeException, AEADBadTagException,
         ShortBufferException {
-        if (len < tagLenBytes) {
+        if (ibuffer.size() + len < tagLenBytes) {
```
pshipton, line 478: This should deal with integer overflow.

test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
```diff
@@ -128,7 +129,8 @@
     GCMBufferTest(String algo, List<dtype> ops) {
         this.algo = algo;
         this.ops = ops;
-        theoreticalCheck = true;
+        theoreticalCheckUpdate = false;
```
pshipton, line 132: Please add a comment indicating the tests which use this variable are being disabled.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -314,6 +319,22 @@ int encrypt(byte[] in, int inOfs, int len, byte[] out, int outOfs) {
...
     }
 
+    /*
+     * This method is for CipherCore to insert the remainder of its buffer
+     * into the ibuffer before a encryptFinal(ByteBuffer, ByteBuffer) operation
+     */
+    int encrypt(byte[] in, int inOfs, int len) {
+        if (len > 0) {
+            // store internally until encryptFinal
+            ArrayUtil.nullAndBoundsCheck(in, inOfs, len);
```
keithc-ca, line 329: I think this check should precede the `if` statement.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -314,6 +319,22 @@ int encrypt(byte[] in, int inOfs, int len, byte[] out, int outOfs) {
...
+            if (ibuffer == null) {
+                ibuffer = new ByteArrayOutputStream();
```
keithc-ca, line 331: The comment on `ibuffer` says 'not used for encryption'; perhaps you meant `ibuffer_enc`?

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -362,6 +383,47 @@ int encryptFinal(byte[] in, int inOfs, int len, byte[] out, int outOfs)
...
+    /**
+     * Performs encryption operation for the last time.
+     *
+     * @param src the input buffer with the data to be encrypted
+     * @param dst the offset in <code>in</code>
+     * @return the number of bytes placed into the <code>out</code> buffer
+     */
+    int encryptFinal(ByteBuffer src, ByteBuffer dst)
+        throws IllegalBlockSizeException, ShortBufferException {
+        byte[] dst_arr = null;
```
keithc-ca, line 395: Please remove the redundant initialization.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -362,6 +383,47 @@ int encryptFinal(byte[] in, int inOfs, int len, byte[] out, int outOfs)
...
+        }
+        int dst_offset = dst.position() + (dst.hasArray() ? dst.arrayOffset() : 0);
+
+        int len = encryptFinal(src_arr, src_ofs, src.remaining(), dst_arr, dst_offset);
+
+        // Update destination buffer
+        if (dst.hasArray()) {
+            dst.position(dst.position() + len);
+        }
+        else {
```
keithc-ca, line 416: No line break here; please write `} else {` (everywhere).

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -362,6 +383,47 @@ int encryptFinal(byte[] in, int inOfs, int len, byte[] out, int outOfs)
...
+        // Get array from destination
+        if (dst.hasArray()) {
+            dst_arr = dst.array();
+        }
+        else {
+            dst_arr = new byte[dst.capacity()];
+        }
+        int dst_offset = dst.position() + (dst.hasArray() ? dst.arrayOffset() : 0);
+
+        int len = encryptFinal(src_arr, src_ofs, src.remaining(), dst_arr, dst_offset);
```
keithc-ca, line 410: I expect a primary motivation for supporting `ByteBuffer` is for performance of the case where it is not backed by an array (e.g. direct buffers). Allocating and copying to an array sacrifices that performance expectation.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -460,6 +519,68 @@ int decryptFinal(byte[] in, int inOfs, int len,
...
+        byte[] dst_arr = null;
+        if (dst.hasArray()) {
+            dst_arr = dst.array();
+        }
+        else {
+            dst_arr = new byte[dst.capacity()];
+        }
+        int dst_offset = dst.position() + (dst.hasArray() ? dst.arrayOffset() : 0);
+
+        int len = decryptFinal(src_arr, src_ofs, src.remaining(), dst_arr, dst_offset);
```
keithc-ca, line 548: Same comment here about performance for the non-array-backed case.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -460,6 +519,68 @@ int decryptFinal(byte[] in, int inOfs, int len,
...
+        if (dst.hasArray()) {
+            dst.position(dst.position() + len);
+        }
+        else {
+            dst.put(dst_arr, dst.position(), len);       
+        }
+
+        if (dst.position() < dst.limit()) {
+            dst.limit(dst.position());
+        }
```
keithc-ca, line 560: This seems inappropriate.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -460,6 +519,68 @@ int decryptFinal(byte[] in, int inOfs, int len,
...
+     * Gets the byte array behind a buffer.
+     * Tries to use ByteBuffer.array(). If this is not available, the function uses ByteBuffer.get()
+     *
+     * @param src the buffer whose byte[] is needed
+     * @return the byte array with the buffer's content
+     */
+    private byte[] getBbArray(ByteBuffer src) {
+        byte[] arr;
+        if (src.hasArray()) arr = src.array();
+        else {
```
keithc-ca, line 574: Please put body of `if` into a block:
```
        if (src.hasArray()) {
            arr = src.array();
        } else {
```
Likewise, line 579 below.

src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
```diff
@@ -1308,7 +1308,11 @@ int gcmDoFinal(ByteBuffer src, ByteBuffer dst) throws ShortBufferException,
             return cipher.decryptFinal(src, dst);
         } else {
             if (buffered > 0) {
-                ((GaloisCounterMode)cipher).encrypt(buffer, 0, buffered);
+                if (useNativeGCM) {
+                    ((NativeGaloisCounterMode) cipher).encrypt(buffer, 0, buffered);
+                } else {
+                    ((GaloisCounterMode) cipher).encrypt(buffer, 0, buffered);
+                }
```
keithc-ca, line 1315: Please undo this change, or explain why it is necessary.

src/java.base/share/classes/com/sun/crypto/provider/FeedbackCipher.java
```diff
@@ -140,6 +140,7 @@ abstract void init(boolean decrypting, String algorithm, byte[] key,
      */
     abstract int encrypt(byte[] plain, int plainOffset, int plainLen,
                          byte[] cipher, int cipherOffset);
+
```
keithc-ca, line 143: Please revert this change.

test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
keithc-ca: Shouldn't the changes being proposed satisfy the tests?

keithc-ca (changes requested):
Please also remove all trailing whitespace in any changes.

src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
keithc-ca: Nevermind, I see the reason now.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Done

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Done

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Done

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Yes, I meant `ibuffer_enc`. Fixed it.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Isn't it faster to compare integers (`if (len > 0)`) than doing `ArrayUtil.nullAndBoundsCheck`? I'm just thinking that the common case would be that the user passes zero as the third argument (`len`) when passing `null` for the first argument (`in`). So it makes sense to first check the length.

keithc-ca: If the length is negative, you've suppressed an exception that should have been thrown.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -362,6 +383,45 @@ int encryptFinal(byte[] in, int inOfs, int len, byte[] out, int outOfs)
...
+        // Get array from source
+        byte[] src_arr = getBbArray(src);
+        int src_ofs = src.position() + (src.hasArray() ? src.arrayOffset() : 0);
+
+        // Get array from destination
+        if (dst.hasArray()) {
+            byte[] dst_arr = dst.array();
+        } else {
+            byte[] dst_arr = new byte[dst.capacity()];
+        }
```
keithc-ca, line 406: This won't compile: `dst_arr` will not be visible to line 409.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
keithc-ca: If `len` and `tagLenBytes` have both been validates as non-negative, then Peter's comment was (I believe) that you should write this:
```
if (ibuffer.size() < tagLenBytes - len) {
```

keithc-ca:
I think you should use the same author for all commits (eventually squashed?). So far we have two:
* Samer Almasri <almasri@ualberta.ca>
* Samer AL Masri <samasri@ibm.com>

pshipton:
Checking what testing is resolved and what still fails with this change.

jenkins test sanity.openjdk zlinux jdk16

pshipton:
PR build is https://ci.eclipse.org/openj9/job/PullRequest-OpenJDKnext/71/

pshipton:
jenkins test sanity.openjdk zlinux jdknext

pshipton:
PR build is https://ci.eclipse.org/openj9/job/PullRequest-OpenJDKnext/72/

pshipton:
```
[2021-01-19T14:12:47.871Z] /home/jenkins/workspace/Build_JDKnext_s390x_linux_Personal/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java:403: error: variable dst_arr is already defined in method encryptFinal(ByteBuffer,ByteBuffer)
[2021-01-19T14:12:47.871Z]             byte[] dst_arr = dst.array();
[2021-01-19T14:12:47.871Z]                    ^
[2021-01-19T14:12:47.872Z] /home/jenkins/workspace/Build_JDKnext_s390x_linux_Personal/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java:405: error: variable dst_arr is already defined in method encryptFinal(ByteBuffer,ByteBuffer)
[2021-01-19T14:12:47.872Z]             byte[] dst_arr = new byte[dst.capacity()];
[2021-01-19T14:12:47.872Z]                    ^
```

samasri:
Yes, I have a commit for this ready, but I realized that some tests are not passing, so I'm fixing them. Will update with a fix here soon.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Right, fixed.

samasri:
The source code should compile and tests should be passing again

pshipton:
jenkins test sanity.openjdk zlinux jdknext

pshipton:
https://ci.eclipse.org/openj9/job/PullRequest-OpenJDKnext/73/

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -380,9 +440,6 @@ int encryptFinal(byte[] in, int inOfs, int len, byte[] out, int outOfs)
      * @return the number of bytes placed into the <code>out</code> buffer
      */
     int decrypt(byte[] in, int inOfs, int len, byte[] out, int outOfs) {
-        if ((len % blockSize) != 0) {
-            throw new ProviderException("Internal error in input buffering");
-       }
```
keithc-ca, line 385: Why are these lines removed from `decrypt`, but not `encrypt`?

pshipton:
test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java  needs to have an IBM copyright added.

keithc-ca:
> test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java needs to have an IBM copyright added.

I'd prefer that we don't modify the test: that file should be reverted.

pshipton:
> I'd prefer that we don't modify the test: that file should be reverted.

Agreed that solution is preferable.

pshipton:
com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java is failing.

Also com/sun/crypto/provider/Cipher/ChaCha20/ChaCha20Poly1305ParamTest.java I think is a new problem, probably not related to this PR.

pshipton:
I don't think the PR testing is running the correct test material.

pshipton:
Created a new issue https://github.com/eclipse/openj9/issues/11689 for the ChaCha20Poly1305ParamTest failure.

pshipton:
If GCMBufferTest.java continues to be modified, it needs to be validated via grinder.
https://ci.eclipse.org/openj9/view/Test/job/Grinder/1434/ - passed

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Okay, I changed it.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -99,6 +101,9 @@
     }
 
     private static void checkDataLength(int processed, int len) {
+        if (len < 0) {
+            return;
+        }
```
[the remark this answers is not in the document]
samasri: I reset the `checkDataLength` to the way it was, and changed the usage of that function in `decryptFinal` to not pass any negative values to it.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: I removed those lines from encrypt also.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -314,6 +312,22 @@ int encrypt(byte[] in, int inOfs, int len, byte[] out, int outOfs) {
         return 0;
     }
 
+    /*
+     * This method is for CipherCore to insert the remainder of its buffer
+     * into the ibuffer before a encryptFinal(ByteBuffer, ByteBuffer) operation
```
keithc-ca, line 317: Please correct the comment: it works with `ibuffer_enc`.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -460,6 +508,65 @@ int decryptFinal(byte[] in, int inOfs, int len,
...
+    private byte[] getBbArray(ByteBuffer src) {
+        byte[] arr;
+        if (src.hasArray()) {
+            arr = src.array();
+        } else {
+            ByteBuffer cpy = src.duplicate();
+            arr = new byte[cpy.capacity()];
+            cpy.limit(cpy.capacity());
+            cpy.position(0);
+            if (arr.length > 0) cpy.get(arr, 0, cpy.capacity());
```
keithc-ca, line 565: I don't think this is correct. The length of the resulting array should be `src.remaining()`.
```
            ByteBuffer cpy = src.duplicate();
            arr = new byte[cpy.remaining()];
            if (arr.length > 0) {
                cpy.get(arr, 0, arr.length);
            }
```
This helper method can also be `static`.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -362,6 +376,42 @@ int encryptFinal(byte[] in, int inOfs, int len, byte[] out, int outOfs)
...
+
+        int len = encryptFinal(src_arr, src_ofs, src.remaining(), dst_arr, dst_offset);
+
+        // Update destination buffer
+        if (dst.hasArray()) {
+            dst.position(dst.position() + len);
+            dst.limit(dst.position());
+        } else {
+            dst.put(dst_arr, dst.position(), len);
+        }
```
keithc-ca, line 410: The destination array does not need to be so large; it need be no larger than `dst.remaining()`.
When a new array is created, `dst_offset` should be zero.
The position of the source buffer must be updated to reflect consumption of input bytes.
The destination buffer limit should not be modified.
```
        // Get array from source
        byte[] src_arr = getBbArray(src);
        int src_ofs = src.hasArray() ? (src.position() + src.arrayOffset()) : 0;

        // Get array from destination
        byte[] dst_arr;
        int dst_offset;

        if (dst.hasArray()) {
            dst_arr = dst.array();
            dst_offset = dst.position() + dst.arrayOffset();
        } else {
            dst_arr = new byte[src.remaining()];
            dst_offset = 0;
        }

        int len = encryptFinal(src_arr, src_ofs, src.remaining(), dst_arr, dst_offset);

        // Advance source buffer position
        src.position(src.position() + len); // FIXME len may be too large

        // Update destination buffer content and position
        if (dst.hasArray()) {
            dst.position(dst.position() + len);
        } else {
            dst.put(dst_arr, 0, len);
        }
```

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
keithc-ca: Changes similar to those proposed for `encryptFinal` are required here.

keithc-ca (changes requested):

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: `ibuffer.size()` is added to `len` in multiple places. Would it make sense to make a new `long` variable to hold the sum?

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: The Cipher API requires handling ByteBuffer inputs, which is the main motivation behind this PR. In terms of performance, the only way to send the ByteBuffer data from Java to OpenSSL is by converting it to a byte[]. Looking further at how to optimize that is beyond the scope of this PR.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: I removed this from here and merged it with the specific case where this is needed.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: The purpose of this function is to get an array similar to the one obtained from `buffer.array()` when the buffer is array-backed. Then I control the memory used in this array using the offset and length. The reason I do this is to get the same array structure regardless of whether the buffer is array-backed or not

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
keithc-ca: That doesn't address the potential for overflow (the conversion from `int` to `long` occurs after the addition).
`len` needs to be validated somewhere that it is non-negative, after which, `tagLenBytes - len` cannot overflow, hence my suggestion.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
keithc-ca: I thought about this more: I don't think it's reasonable to create a byte array of length `buffer.remaining()`; imagine the case where the `ByteBuffer` is mapped to a large input file.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -460,6 +508,65 @@ int decryptFinal(byte[] in, int inOfs, int len,
...
+            dst_arr = new byte[dst.capacity()];
+        }
+        int dst_offset = dst.position() + (dst.hasArray() ? dst.arrayOffset() : 0);
+
+        int len = decryptFinal(src_arr, src_ofs, src.remaining(), dst_arr, dst_offset);
+        
+        // Update destination buffer
+        if (dst.hasArray()) {
+            dst.position(dst.position() + len);
+            dst.limit(dst.position());
```
keithc-ca, line 541: The `limit()` of any user-supplied `ByteBuffer` should not be updated.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Shouldn't the src position become the src limit?
Because we are reading all of the `src.remaining()` and encrypting them. So I think we could simple add afterwards:
```java
src.position(src.limi());
```

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Removed the `dst.limit()` calls.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
keithc-ca: Yes, that's reasonable assuming `encryptFinal` has succeeded and consumed all the bytes it was given.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Done.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Done.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: The suggested solution would work when adding `ibuffer.size()` to `len` in this if statement. However, since I will be adding these two variables in 3 different places in that function, I am looking into a more generic solution. This is how I ended up solving it (in [commit 9cbcfc2](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/266/commits/9cbcfc23e5e72be110c20579657695f469c466b4)):

I replace `len += ibuffer.size()` with:

```java
try {
    len = Math.addExact(len,ibuffer.size());
}
catch(ArithmeticException e) {
    throw new ProviderException("SunJCE provider only supports " +
        "input size up to " + MAX_BUF_SIZE + " bytes");
}
```
I throw an exception if an integer overflow happens because this is the intended behaviour. In `checkDataLength`, if `len + proc > Integer.MAX_VALUE`, the same exception is thrown. The check I add here is essentially the same check, but I add it earlier in the code.
Having said that, one might ask why do I keep `checkDataLength` if the same check is essentially happening earlier. This is because if we later decide to change the `MAX_BUF_SIZE` to a value smaller than `Integer.MAX_VALUE`, the check would not be redundant anymore. But for our current case, the second check does not add any value.

test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
samasri: The reason I edited this test separating the check for update and final is because according to the [java documentation](https://download.java.net/java/early_access/jdk16/docs/api/java.base/javax/crypto/Cipher.html#update(byte%5B%5D,int,int,byte%5B%5D)), `cipher.update` must return `the number of bytes stored in output`. This number, in our case, is zero given that we just stored the plain text in a buffer and did not encrypt anything yet. However, other implementations can return a different number here.

Before my changes, the `theoreticalCheck` was already there ([here](https://github.com/samasri/openj9-openjdk-jdk/blob/ac04db2e39441f285acebb657aa388043a894770/test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java#L64-L65)) because it was expected that in some cases, the theoretical check is not straight forward to verify. I am turning off the check for update here because our implementation is complying with the cipher.update expectations in the docs, but is different than what other implementations return.

keithc-ca: If the test is modified, an IBM copyright must be added to this file.

samasri:
@keithc-ca @pshipton  Just a head's up that I've addressed all the issues above. Please let me know if I need to change anything else.

test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
samasri: Done.

samasri: Done.

keithc-ca: The copyright should list two dates (even when they're the same):
```
* (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
```

keithc-ca: > Please add a comment indicating the tests which use this variable are being disabled.

I don't see the comment Peter requested.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -362,6 +377,46 @@ int encryptFinal(byte[] in, int inOfs, int len, byte[] out, int outOfs)
         return (len + tagLenBytes);
     }
 
+    /**
+     * Performs encryption operation for the last time.
+     *
+     * @param src the input buffer with the data to be encrypted
+     * @param dst the offset in <code>in</code>
+     * @return the number of bytes placed into the <code>out</code> buffer
```
keithc-ca, line 385: Please fix this comment: the parameter is named `dst`.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -362,6 +377,46 @@ int encryptFinal(byte[] in, int inOfs, int len, byte[] out, int outOfs)
...
+        int src_ofs = src.hasArray() ? (src.position() + src.arrayOffset()) : 0;
+
+        // Get array from destination
+        byte[] dst_arr;
+        int dst_offset;
+        if (dst.hasArray()) {
+            dst_arr = dst.array();
+            dst_offset = dst.position() + dst.arrayOffset();
+        } else {
+            dst_arr = new byte[dst.capacity()];
```
keithc-ca, line 401: This is larger than necessary; it should be:
```
dst_arr = new byte[dst.remaining()];
```

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -416,16 +467,23 @@ int decryptFinal(byte[] in, int inOfs, int len,
                      byte[] out, int outOfs)
         throws IllegalBlockSizeException, AEADBadTagException,
         ShortBufferException {
+        try {
+            len = Math.addExact(len,ibuffer.size());
+        }
+        catch(ArithmeticException e) {
+            throw new ProviderException("SunJCE provider only supports " +
+                "input size up to " + MAX_BUF_SIZE + " bytes");
+        }
```
keithc-ca, line 476: I would prefer to see explicit range checking:
```
  (0 <= len) && (len <= MAX_BUF_SIZE - ibuffer.size())
```
As written, there's no clear connection between the situations in which `addExact` would throw an exception and the value of `MAX_BUF_SIZE`.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -460,6 +519,67 @@ int decryptFinal(byte[] in, int inOfs, int len,
...
+     * @return the byte array with the buffer's content
+     */
+    private static byte[] getBbArray(ByteBuffer src) {
+        byte[] arr;
+        if (src.hasArray()) {
+            arr = src.array();
+        } else {
+            ByteBuffer cpy = src.duplicate();
+            arr = new byte[cpy.remaining()];
+            if (arr.length > 0) cpy.get(arr, 0, arr.length);
```
keithc-ca, line 578: Please use braces (`{}`) with the body on a separate line:
```
if (arr.length > 0) {
    cpy.get(arr, 0, arr.length);
}
```

test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
```diff
@@ -128,7 +134,9 @@
     GCMBufferTest(String algo, List<dtype> ops) {
         this.algo = algo;
         this.ops = ops;
-        theoreticalCheck = true;
+        theoreticalCheckUpdate = false; // Disabling length verification after cipher.update for 
+                                        // GCM Buffer Tests, the variable is never set to true
```
keithc-ca, line 138: I think a better place for the comments is with the field declarations.

keithc-ca (changes requested):

test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
pshipton: The comment text contains what I wanted to see (the variable is never set to true). I'm fine with having it on the declaration.

samasri: Moved comment to the declaration.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: I edited comments in both: `encryptFinal` and `decryptFinal` to match the actual parameters.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Fixed.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: The code I add here is not intended to be connected to `MAX_BUF_SIZE`. I need to execute: `len += ibuffer.size()`. However, since this might overflow, I add the code above to handle the possibility of overflow.

The range checking you mention int he above comment is found in the function `checkDataLength`, which is called on line 484. The only part of the range checking you mention above that is not currently there is `len >= 0`. Hence, I added this check at the beginning of `decryptFinal`:

```java
if (len < 0) {
  throw new ProviderException("Input len is negative"); }
```

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Fixed.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -314,6 +313,22 @@ int encrypt(byte[] in, int inOfs, int len, byte[] out, int outOfs) {
         return 0;
     }
 
+    /*
+     * This method is for CipherCore to insert the remainder of its buffer
+     * into the ibuffer_enc before a encryptFinal(ByteBuffer, ByteBuffer) operation
```
keithc-ca, line 318: Please remove 'the' here.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
keithc-ca: Please open an issue at https://github.com/eclipse/openj9/issues to track such improvements.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
keithc-ca: I suggest `len` should not be modified at all in this method. Instead, after the check you added that `len` is not negative, the expression `ibuffer.size() < tagLenBytes - len` is safe from overflow and does the necessary validation.

Parameters `inOfs` and `outOfs` should also be validated.

test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
```diff
@@ -62,7 +67,9 @@
     int[] sizes;
     boolean incremental = false;
     // In some cases the theoretical check is too complicated to verify
-    boolean theoreticalCheck;
+    boolean theoreticalCheckUpdate; // This variable is never set to true in order to disable
+                                    // length verification after cipher.update for GCM Buffer Tests
+    boolean theoreticalCheckFinal;
```
keithc-ca, line 72: Please put comments before declarations (following the existing code):
```
    // In some cases the theoretical check is too complicated to verify
    boolean theoreticalCheckFinal;
    // This variable is never set to true in order to disable
    // length verification after cipher.update for GCM Buffer Tests
    boolean theoreticalCheckUpdate;
```

samasri: Done.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Done.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Sure, I can remove it. I was just thinking because in the original code we also have the following:

```java
if (out.length - outOfs < ((ibuffer.size() + len) - tagLenBytes)) {
  throw new ShortBufferException("Output buffer too small");
}
```

This was already a potential integer overflow. But I can change that condition as well to: `out.length - outOfs - ibuffer.size() < len - tagLenBytes`

keithc-ca: This need not change: the test above that `ibuffer.size() < tagLenBytes - len` ensures that `ibuffer.size() + len` cannot overflow.

samasri: Makes sense. So I changed the code to have the following at the beginning:
```java
if (len < 0 && (len > MAX_BUF_SIZE - ibuffer.size() + tagLen)) {
  throw new ProviderException("SunJCE provider only supports " +
  "a positive input size up to " + MAX_BUF_SIZE + " bytes");    
}
```
This enforces the condition you suggested initially in this thread. I also removed the `checkDataLength` call since its redundant now (with the extra check above).

pshipton:
I would be good to get this finalized and merged before the code split occurs this weekend.

pshipton:
Once it is finalized, it needs to be backported to jdk16.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -416,16 +467,16 @@ int decryptFinal(byte[] in, int inOfs, int len,
                      byte[] out, int outOfs)
         throws IllegalBlockSizeException, AEADBadTagException,
         ShortBufferException {
-        if (len < tagLenBytes) {
-            throw new AEADBadTagException("Input too short - need tag");
+        if (len < 0 && (len > MAX_BUF_SIZE - ibuffer.size() + tagLenBytes)) {
+            throw new ProviderException("SunJCE provider only supports " +
+            "a positive input size up to " + MAX_BUF_SIZE + " bytes");
```
keithc-ca, line 472: I think you meant to use `||` rather than `&&` here, but it would be better to separate the negative test because I think it deserves a different exception (at least the message should be different).

The second test `(len > MAX_BUF_SIZE - ibuffer.size() + tagLenBytes)` still might overflow (e.g. if `ibuffer` is empty). I don't think the `tagLenBytes` should be there.

I think the result should be something like this:
```
        if (len < 0) {
            throw new ProviderException("Input length is negative");
        }
        if (len < tagLenBytes - ibuffer.size()) {
            throw new AEADBadTagException("Input too short - need tag");
        }
        if (len > MAX_BUF_SIZE - ibuffer.size()) {
            throw new ProviderException("SunJCE provider only supports "
                + "a positive input size up to " + MAX_BUF_SIZE + " bytes");
        }
```

keithc-ca (changes requested):

pshipton:
Note the openjdk tests failing due to this problem have been excluded, and will need to be re-included when this is ready to merge.
https://github.com/AdoptOpenJDK/openjdk-tests/pull/2249

test/jdk/com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
```diff
@@ -62,7 +67,9 @@
     int[] sizes;
     boolean incremental = false;
     // In some cases the theoretical check is too complicated to verify
-    boolean theoreticalCheck;
+    // This variable is never set to true in order to disable length verification after cipher.update for GCM Buffer Tests
+    boolean theoreticalCheckUpdate;
+    boolean theoreticalCheckFinal;
```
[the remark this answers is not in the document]
samasri: Done.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -36,7 +36,10 @@
 import javax.crypto.*;
 import static com.sun.crypto.provider.AESConstants.AES_BLOCK_SIZE;
 import com.sun.crypto.provider.AESCrypt;
+import sun.security.util.ArrayUtil;
+import java.lang.ArithmeticException;
```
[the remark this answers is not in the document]
samasri: Done.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -380,10 +435,6 @@ int encryptFinal(byte[] in, int inOfs, int len, byte[] out, int outOfs)
      * @return the number of bytes placed into the <code>out</code> buffer
      */
     int decrypt(byte[] in, int inOfs, int len, byte[] out, int outOfs) {
-        if ((len % blockSize) != 0) {
-            throw new ProviderException("Internal error in input buffering");
-       }
-
         checkDataLength(ibuffer.size(), len);
```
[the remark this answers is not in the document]
samasri: Okay, I added the condition in the body of `checkDataLength`.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Okay, I did the suggested change, and the tests (my local run of them) are passing.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -416,16 +469,22 @@ int decryptFinal(byte[] in, int inOfs, int len,
...
+                + "a positive input size up to " + MAX_BUF_SIZE + " bytes");
+        }
 
-        // do this check here can also catch the potential integer overflow
-        // scenario for the subsequent output buffer capacity check.
-        checkDataLength(ibuffer.size(), (len - tagLenBytes));
-
+        if (ibuffer.size() < tagLenBytes - len) {
+            throw new AEADBadTagException("Input too short - need tag");
+        }
```
keithc-ca, line 485: This is redundant (a repeat of the test on line 475).

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -460,6 +519,69 @@ int decryptFinal(byte[] in, int inOfs, int len,
...
+        src.position(src.limit());
+        
+        // Update destination buffer
+        if (dst.hasArray()) {
+            dst.position(dst.position() + len);
+        } else {
+            dst.put(dst_arr, 0, len);       
+        }
+        
+        return len;
```
keithc-ca, line 561: Please remove the trailing whitespace.

keithc-ca (changes requested):

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Done.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
samasri: Removed.

keithc-ca:
Please squash and rebase on the head of the openj9 branch.

keithc-ca:
Please fix the commit message: the author and committer should be the same.
You may want to use `git commit --amend --author='Samer AL Masri <samasri@ibm.com>'`.
Please also remove the extra `#Signed-off-by:` lines.

samasri:
Sorry about that, done.

keithc-ca:
Jenkins test sanity,extended plinux jdknext

keithc-ca:
Jenkins test sanity.openjdk zlinux jdknext

pshipton:
Tested the previously failing (and excluded) openjdk tests via grinders. All passed.

test/jdk/com/sun/crypto/provider/Cipher/AEAD/Encrypt.java
https://ci.eclipse.org/openj9/view/Test/job/Grinder/1520 - passed

com/sun/crypto/provider/Cipher/AEAD/GCMBufferTest.java
https://ci.eclipse.org/openj9/view/Test/job/Grinder/1521 - passed

com/sun/crypto/provider/Cipher/AEAD/OverlapByteBuffer.java
https://ci.eclipse.org/openj9/view/Test/job/Grinder/1522 - passed

com/sun/crypto/provider/Cipher/AEAD/SameBuffer.java
https://ci.eclipse.org/openj9/view/Test/job/Grinder/1523 - passed

Create PR to un-exclude the tests
https://github.com/AdoptOpenJDK/openjdk-tests/pull/2266

Created PRs to port this change to jdk16, assuming the current commit is the final one. In draft state until this is confirmed.
https://github.com/ibmruntimes/openj9-openjdk-jdk16/pull/25
https://github.com/ibmruntimes/openj9-openjdk-jdk16/pull/26

keithc-ca:
The sanity.openjdk failure is eclipse/openj9#11930.

keithc-ca (approved):

keithc-ca:
Thanks, @samasri for being persistent with this.

PR #271, opened 2021-01-19
Add accelerated zlibNX to LIBPATH for AIX P9 or newer systems and set system zlib as default on AIX

rmnattas:
zlibNX is an accelerated replacement for zlib, it make use of the NX accelerator in P9 and newer systems.
This PR enable the use of zlibNX for AIX systems by adding it to the LIBPATH if the system is P9 or newer and has the NX accelerator enabled.

This PR also sets system zlib as default instead of bundled zlib on AIX systems to allow zlibNX
For reference, the change to bundle zlib on AIX is https://bugs.openjdk.java.net/browse/JDK-8223685

Signed-off-by: Abdulrahman Alattas rmnattas@gmail.com

### DIFF
--- a/make/autoconf/lib-bundled.m4
+++ b/make/autoconf/lib-bundled.m4
@@ -23,6 +23,10 @@
 # questions.
 #
 
+# ===========================================================================
+# (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+# ===========================================================================
+
 ################################################################################
 # Setup bundled libraries.
 #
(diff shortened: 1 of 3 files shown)


rmnattas:
@pshipton @keithc-ca

keithc-ca:
Jenkins compile aix,win jdknext

keithc-ca:
Jenkins copyright check

keithc-ca (approved):

PR #280, opened 2021-02-05
Fix null nonce crash in ChaCha20Poly1305Param Test

shanchao95:
* issue eclipse/openj9#11689

Signed-off-by: Chao Shan <shanchao95@gmail.com>

### DIFF
--- a/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
+++ b/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2020 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2021 All Rights Reserved
  * ===========================================================================
  */
 package com.sun.crypto.provider;
@@ -193,7 +193,7 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
      */
     @Override
     protected byte[] engineGetIV() {
-        return nonce.clone();
+        return (nonce == null)? null : nonce.clone();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


shanchao95:
@pshipton Code is updated for issue eclipse/openj9#11689

pshipton (approved):

pshipton:
Same change as jdk16, I'm not going to re-run testing.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -211,13 +211,14 @@ protected int engineGetOutputSize(int inputLen) {
     @Override
     protected AlgorithmParameters engineGetParameters() {
         AlgorithmParameters params = null;
+        byte[] nonceData = (initialized || nonce != null) ? nonce : createRandomNonce(null);
```
keithc-ca, line 214: What is the motivation for the test of `initialized` here? If it is true and `nonce` is `null` we'll get a `NullPointerException` below. Why don't we write:
```
  byte[] nonceData = (nonce != null) ? nonce : createRandomNonce(null);
```
I have the same question about ibmruntimes/openj9-openjdk-jdk16#13.

keithc-ca: Same question for ibmruntimes/openj9-openjdk-jdk11#383.

keithc-ca: @shanchao95 Perhaps you meant to write:
```
byte[] nonceData = (initialized && nonce != null) ? nonce : createRandomNonce(null);
```
so that even if `nonce != null` a random nonce is used after a 'final' operation, like [here](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java#L709)?

keithc-ca: I opened PRs for Java 11, 16 and 17 with my proposed change.

pshipton: @shanchao95 @ashbm5 can you please take a look at Keith's suggested update, and comment if it looks like what was intended. You can either comment here or in one of the PRs below.

These are all the same change for the different versions:
https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/385
https://github.com/ibmruntimes/openj9-openjdk-jdk16/pull/27
https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/288

PR #284, opened 2021-02-09
Enable cmake for Windows by default

pshipton:
Signed-off-by: Peter Shipton <Peter_Shipton@ca.ibm.com>

PR #285, opened 2021-02-10
Enable cmake for AIX by default

pshipton:
This is so far untested, although jdk16 looks good https://github.com/ibmruntimes/openj9-openjdk-jdk16/pull/20#issue-570744853


pshipton:
jenkins test sanity,sanity.openjdk aix jdknext

keithc-ca (approved):

keithc-ca:
The test failure seems to be eclipse/openj9#11930 which according to https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/284#issuecomment-776131012 is unrelated to cmake.

PR #288, opened 2021-02-11
Access nonce only when non-null

keithc-ca:
This should resolve the discussion started here: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/280#pullrequestreview-584651273.

### DIFF
--- a/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
+++ b/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
@@ -193,7 +193,7 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
      */
     @Override
     protected byte[] engineGetIV() {
-        return (nonce == null)? null : nonce.clone();
+        return (nonce != null) ? nonce.clone() : null;
     }
 
     /**
@@ -211,7 +211,7 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
     @Override
     protected AlgorithmParameters engineGetParameters() {
         AlgorithmParameters params = null;
-        byte[] nonceData = (initialized || nonce != null) ? nonce : createRandomNonce(null);
+        byte[] nonceData = (nonce != null) ? nonce : createRandomNonce(null);
         if (mode == MODE_AEAD) {
             try {
                 // Force the 12-byte nonce into a DER-encoded OCTET_STRING
@@ -225,7 +225,6 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
                 throw new RuntimeException(exc);
             }
         }
-
         return params;
     }
 


closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -211,7 +211,7 @@ protected int engineGetOutputSize(int inputLen) {
     @Override
     protected AlgorithmParameters engineGetParameters() {
         AlgorithmParameters params = null;
-        byte[] nonceData = (initialized || nonce != null) ? nonce : createRandomNonce(null);
+        byte[] nonceData = (initialized && nonce != null) ? nonce : createRandomNonce(null);
```
JasonFengJ9, line 214: This change introduces a scenario in which `createRandomNonce(null)` is invoked when `initialized` is `false` but `nonce != null` which seems not desirable.
It appears that `initialized` is set to `true` after `nonce` is assigned with `newNonce` assuming not `null`.
Maybe the code can be `byte[] nonceData = (nonce != null) ? nonce : createRandomNonce(null)` instead?

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -193,7 +193,7 @@ protected int engineGetOutputSize(int inputLen) {
      */
     @Override
     protected byte[] engineGetIV() {
-        return (nonce == null)? null : nonce.clone();
+        return (initialized && nonce != null) ? nonce.clone() : null;
```
JasonFengJ9, line 196: If `initialized` is `false`, a `null` will be returned even `nonce` is already set to a non-null value which is not this API intent.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
keithc-ca: `initialized` is set to `false` as part of a 'final' operation: https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java#L709

We don't want to use `nonce` after that even if it is not `null`.

JasonFengJ9: `engineGetParameters()` is to get the algorithm parameters used/generated by this Cipher engine. This API could be invoked before or after `engineUpdate()/engineDoFinal()`, I think the cipher algorithm parameters is not affected by the actual data processing. On the other words, if `initialized` is `true` or `nonce` is already set to a non-null value, current `nonce` value should be returned, otherwise a random nonce will be generated.
I think original code can stay.

shanchao95: After `final` operation, I believe the cipher instantiation is expected to have the same functionalities as a new cipher instantiation. 
A new cipher instantiation without initialization should have a different nonce for every `engineGetParameters` when it gets invoked multiple times. So the `initialized` should be the primary condition. Once `initialized == false`, we need to generate a new nonce. The change looks good to me

JasonFengJ9: As per https://docs.oracle.com/javase/8/docs/api/javax/crypto/Cipher.html#getParameters--

public final AlgorithmParameters getParameters()
Returns the parameters used with this cipher.
**The returned parameters may be the same that were used to initialize this cipher**, or may contain a combination of default and random parameter values used by the underlying cipher implementation if this cipher requires algorithm parameters but was not initialized with any.

> After final operation, I believe the cipher instantiation is expected to have the same functionalities as a new cipher instantiation.

Can this be verified and compare with RI behaviours?

keithc-ca: I adapted some test code found in https://github.com/mkyong/core-java to compare with the RI: I'll report back here once I have made any necessary adjustments.

keithc-ca: ChaCha20 is new in Java 11; the [javadocs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/javax/crypto/Cipher.html) note special requirements for that cipher regarding initialization.

In the RI, both `getIV()` and `getParameters()` throw `NullPointerException` if called before `init()`. Any time after a call to `init()`, even after a 'final' operation, those methods in the RI return the most recent parameter information. I will update this to do likewise.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
keithc-ca: I will remove the `initialized` portion of the condition here.

keithc-ca:
Updated. Thanks, @JasonFengJ9.

pshipton:
@JasonFengJ9 I'll wait for your approval, or you can merge them.

pshipton:
Note testing https://github.com/ibmruntimes/openj9-openjdk-jdk16/pull/27#issuecomment-781553511 passed.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
JasonFengJ9: Yeah, this is the specific requirement for `ChaCha20` and its friend.
_The ChaCha20 and ChaCha20-Poly1305 algorithms have a similar requirement for unique nonces with a given key. After each encryption or decryption operation, callers should re-initialize their ChaCha20 or ChaCha20-Poly1305 ciphers with parameters that specify a different nonce value._

JasonFengJ9 (approved):
LGTM

JasonFengJ9:
Merging as per https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/288#issuecomment-781594430

PR #289, opened 2021-02-16
Fix help for '--with-openssl'

keithc-ca:
Correct imbalance of `<` and `>` and use language consistent with other feedback in that option.


pshipton (approved):

pshipton:
jenkins check line endings

pshipton:
jenkins line endings check

PR #290, opened 2021-02-22
Merging master HEAD into openj9-staging

JasonFengJ9:
Resolved a conflict at `src/java.base/share/classes/sun/nio/cs/US_ASCII.java`;
Not accept this OpenJDK change which can be applied when `OpenJ9` switches to `RI` `java.lang.String`.


JasonFengJ9:
jenkins test sanity plinux jdknext

JasonFengJ9:
> 09:21:56  /home/jenkins/workspace/Build_JDKnext_ppc64le_linux_Personal/build/linux-ppc64le-server-release/support/j9jcl/java.base/share/classes/java/lang/Access.java:61: error: Access is not abstract and does not override abstract method decodeASCII(byte[],int,char[],int,int) in JavaLangAccess
09:21:56  final class Access implements JavaLangAccess {
09:21:56        ^
09:21:56  Note: Some input files additionally use or override a deprecated API.
09:21:56  Note: Some input files additionally use or override a deprecated API that is marked for removal.
09:21:56  Note: Some input files additionally use unchecked or unsafe operations.
09:21:56  1 error

I will fix this.

keithc-ca:
I think it might make sense to accept the upstream changes
* 8261744: Implement CharsetDecoder ASCII and latin-1 fast-paths

We'll also need to implement the new methods defined by `java.lang.Access`.

JasonFengJ9:
> I think it might make sense to accept the upstream changes

The modification to `java.lang.Access` will be applied which is in OpenJ9 repo.
The change to `US_ASCII.java` can be skipped, cause we are investigating replace OpenJ9 `java.lang.String` with RI version and will pause any change to OpenJ9 String for JDK17 for now.

keithc-ca:
> > I think it might make sense to accept the upstream changes
> 
> The modification to `java.lang.Access` will be applied which is in OpenJ9 repo.
> The change to `US_ASCII.java` can be skipped, cause we are investigating replace OpenJ9 `java.lang.String` with RI version and will pause any change to OpenJ9 String for JDK17 for now.

The change to `US_ASCII.java` is independent of a switch to using the reference implementation of `java.lang.String`: I see no reason to keep the difference. The change from upstream seems to be an optimization similar to the one for OpenJ9.

JasonFengJ9:
> The change to US_ASCII.java is independent of a switch to using the reference implementation of java.lang.String: I see no reason to keep the difference. The change from upstream seems to be an optimization similar to the one for OpenJ9.

The change to `US_ASCII.java` requires implementation of `java.lang.String.decodeASCII()` invoked via `JLA.decodeASCII(sa, sp, da, dp, Math.min(sl - sp, dl - dp))`.
Current JCL patch already has an improvement in place (comparing with previous OpenJDK implementation).
OpenJ9 JDK17 can have a stub for `decodeASCII` until switching to RI `String`.

keithc-ca:
> The change to US_ASCII.java requires implementation of java.lang.String.decodeASCII()

The only reference to `String.decodeASCII()` is in the reference implementation `java.lang.System` which OpenJ9 does not use. OpenJ9 must implement `JavaLangAccess.decodeASCII()`, but that doesn't imply anything about `java.lang.String`.

JasonFengJ9:
> The only reference to String.decodeASCII() is in the reference implementation java.lang.System which OpenJ9 does not use. OpenJ9 must implement JavaLangAccess.decodeASCII(), but that doesn't imply anything about java.lang.String.

`JavaLangAccess` is an interface implemented by OpenJ9 `Access` which is expected to call into `String.decodeASCII()` for this `decodeASCII` usage (technically could be implemented in other places).

P.S. just noticed that there are other JCL callers to `decodeASCII` so OpenJ9 has to provide an implementation.

JasonFengJ9:
**Update:**

Resolved latest conflicts:
```
      make/Main.gmk
      src/java.base/share/classes/sun/nio/cs/US_ASCII.java
```
Adopt RI change to use `JLA.decodeASCII`.

JasonFengJ9:
jenkins test sanity plinux jdknext

keithc-ca:
I would have liked to see `US_ASCII` adopted entirely from upstream, but we can defer the discussion about removing changes to 
`US_ASCII.Encoder` until later.

JasonFengJ9:
> I would have liked to see US_ASCII adopted entirely from upstream

There are still `OpenJ9-perf_converter` changes around `encodeASCII`.
Yeah, it is the goal to use upstream w/o JCL patches.

src/java.base/share/classes/sun/nio/cs/US_ASCII.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2000, 2017, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2000, 2021, Oracle and/or its affiliates. All rights reserved.
```
keithc-ca, line 2: Since we're keeping some OpenJ9 changes to this file, its copyright date should be updated.

JasonFengJ9: Updated IBM copyright

keithc-ca:
PR build: https://ci.eclipse.org/openj9/job/PullRequest-OpenJDKnext/97/ should not be invalidated by copyright date change.

keithc-ca:
The build was successful, but testing failed due to eclipse/openj9#12027; that need not hold this up.

keithc-ca (approved):

PR #291, opened 2021-02-23
Remove many unnecessary classes from openj9.dtfj

keithc-ca:
All of the classes in the `com.ibm.j9ddr.vm29.structure` package are derived at runtime from the related core file and so do not need to be included in the `openj9.dtfj` module.

The `module-info` class file lists `com/ibm/j9ddr/vm29/structure` as a package defined by that module, so we retain one class (`DDRAlgorithmVersions`) to avoid an inconsistency.

The net effect of this change is to reduce the size of the `modules` file by about 2MB which may reduce the physical memory required by the VM.

### DIFF
--- a/closed/custom/CreateJmods.gmk
+++ b/closed/custom/CreateJmods.gmk
@@ -1,0 +1,29 @@
+# ===========================================================================
+# (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+# ===========================================================================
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins signed off by check

pshipton:
jenkins copyright check

pshipton:
jenkins line endings check

pshipton (approved):

PR #295, opened 2021-03-12
OpenJ9 Adopts RI java.lang.String

JasonFengJ9:
Add OpenJ9 specific fields/methods into RI `java.lang.String`;
Make `StringCoding.implEncodeISOArray()` `public` which is required by RI `String`;
A separated PR (https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/296) is to remove legacy `StringCoding` code.

Note: this change to `java.lang.String` has no impact on current JDK17 build cause it is not included by OpenJ9 yet.

This unlocks https://github.com/eclipse/openj9/pull/12209.

### DIFF
--- a/src/java.base/share/classes/java/lang/String.java
+++ b/src/java.base/share/classes/java/lang/String.java
@@ -23,9 +23,16 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+ * ===========================================================================
+ */
+
 package java.lang;
 
 import java.io.ObjectStreamField;
+import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.lang.annotation.Native;
 import java.lang.invoke.MethodHandles;
@@ -51,6 +58,8 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
+
+import jdk.internal.misc.Unsafe;
 import jdk.internal.vm.annotation.IntrinsicCandidate;
 import jdk.internal.vm.annotation.Stable;
 import sun.nio.cs.ArrayDecoder;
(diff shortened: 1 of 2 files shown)


JasonFengJ9:
`11:25:54  ERROR - should have LF line terminators: 'src/utils/IdealGraphVisualizer/ServerCompiler/nbproject/project.xml' type: 'XML 1.0 document, ASCII text, with CRLF line terminators'`

jenkins line endings check

JasonFengJ9:
jenkins copyright check

pshipton:
@fjeremic any concerns with this?

pshipton:
jenkins compile osx jdknext

JasonFengJ9:
> @fjeremic any concerns with this?

Somehow I can't request Filip as a reviewer. (DM him)

src/java.base/share/classes/java/lang/String.java
```diff
@@ -217,11 +224,389 @@
...
+
+			// Copy in s1 contents
+			decompressedArrayCopy(s1.value, 0, value, 0, s1len);
+
+			if (COMPACT_STRINGS) {
+				initCompressionFlag();
+			}
+		}
+	}
+
```
fjeremic, line 595: I think we're missing a private constructor accelerated by the JIT:
https://github.com/eclipse/openj9/blob/a9838e74bba0f5935103bedcda6fc84c881804ca/jcl/src/java.base/share/classes/java/lang/String.java#L1156-L1337

JasonFengJ9: Good catch, just wondering if JIT falls back to a less-optimized call or we were just lucky not hit by `NoSuchMethodError`?

fjeremic: We have an optimization which looks for String concatenations and replaces them with a call to a private String constructor. If that constructor is not found we just don't do the optimization (StringPeepholes).

JasonFengJ9: Filip, could you shine some lights on how the fallback works when this constructor [1] is not found? A quick experiment shows that missing a similar 3 parameter private constructor [2] caused build failure
```
Assertion failed at /Users/ivm/osx-openj9-builds/jdk17-0316-jdk17ristrV2/openj9-openjdk-jdk/omr/compiler/compile/OMRSymbolReferenceTable.cpp:979: method
VMState: 0x000535ff
	methodSymRefFromName: method must exist: java/lang/String.<init>(Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)V
compiling java/util/Arrays.copyOfRange([BII)[B at level: warm
``` 

[1] https://github.com/eclipse/openj9/blob/a9838e74bba0f5935103bedcda6fc84c881804ca/jcl/src/java.base/share/classes/java/lang/String.java#L1156-L1159
[2] https://github.com/eclipse/openj9/blob/a9838e74bba0f5935103bedcda6fc84c881804ca/jcl/src/java.base/share/classes/java/lang/String.java#L1007-L1010

fjeremic: This looks like the code which tries to find the symbol reference:
https://github.com/eclipse/openj9/blob/a9838e74bba0f5935103bedcda6fc84c881804ca/runtime/compiler/optimizer/StringPeepholes.cpp#L157-L214

I did not know that call to `methodSymRefFromName` will assert if it cannot find the symbol. So it looks like it is a bug not to have these methods.

JasonFengJ9: Added `private String(int v1, String s1, int v2, String s2, String s3)` for https://github.com/eclipse/openj9/blob/a9838e74bba0f5935103bedcda6fc84c881804ca/runtime/compiler/optimizer/StringPeepholes.cpp#L198 
Also added `String(byte[] data, int start, int length, boolean compressed)` for https://github.com/eclipse/openj9/blob/a9838e74bba0f5935103bedcda6fc84c881804ca/runtime/compiler/optimizer/StringPeepholes.cpp#L197

@fjeremic please have another look.

JasonFengJ9:
jenkins compile zlinux jdknext

fjeremic:
Functionally this is ok. Performance wise it is questionable since it is missing things like hashcode acceleration, etc. We should discuss this on Monday on the call you scheduled before proceeding.

JasonFengJ9:
Rebasing to resolve the conflict.
It appears the modification to `equals()` & `hashCode()` has more chances of conflict.

JasonFengJ9:
@fjeremic could you take another look at this? thanks.

fjeremic (approved):

JasonFengJ9:
@pshipton could you review & merge if approve?

JasonFengJ9:
jenkins compile zlinux jdknext

pshipton:
jenkins test sanity zlinux jdknext

src/java.base/share/classes/java/lang/StringCoding.java
```diff
@@ -600,7 +606,7 @@ private static Result decodeLatin1(byte[] ba, int off, int len) {
     }
 
     @IntrinsicCandidate
-    private static int implEncodeISOArray(byte[] sa, int sp,
+    public static int implEncodeISOArray(byte[] sa, int sp,
```
pshipton, line 609: Is this change needed?

JasonFengJ9: Yes, this is to match `StringCoding` in `master` branch to be referred by `String`.

pshipton (approved):

PR #313, opened 2021-05-12
DDR: do test compile with minimal set of fields

keithc-ca:
This updates the build steps for DDR code to properly handle optional fields.
This depends on eclipse-openj9/openj9#12676.


keithc-ca:
Jenkins line endings check

pshipton (approved):

PR #329, opened 2021-06-11
[JDK17] Fix the bug with the return type of the omitted finalizers in loop

ChengJin01:
The change is to fix the bug with the return type of the omitted
finalizers as OpenJDK doesn't follow the Spec to set the loop
return type.

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+ * ===========================================================================
+ */
+
 package java.lang.invoke;
 
 import jdk.internal.access.SharedSecrets;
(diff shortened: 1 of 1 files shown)


ChengJin01:
The fix is the same as https://github.com/ibmruntimes/openj9-openjdk-jdk16/pull/58 given the code involved never changes for finalizers in the loop code.

Reviewer: @pshipton, @tajila
FYI: @DanHeidinga, @fengxue-IS, @babsingh

src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -6671,7 +6677,7 @@ public static MethodHandle loop(MethodHandle[]... clauses) {
                 pred.set(i, dropArguments0(constant(boolean.class, true), 0, commonParameterSequence));
             }
             if (fini.get(i) == null) {
-                fini.set(i, empty(methodType(t, commonParameterSequence)));
+                fini.set(i, empty(methodType(loopReturnType, commonParameterSequence)));
```
JasonFengJ9, line 6680: As per https://github.com/eclipse-openj9/openj9/issues/11937#issuecomment-856086725
> We will need to raise an OpenJDK bug report to request them to fix the issue to macth the expected behavior in the Spec.

fyi @andrew-m-leonard

tajila: A request has been made to open a bug report

tajila (approved):

PR #330, opened 2021-06-14
Fix UMA criu enablement

tajila:
Fix UMA criu enablement

Signed-off-by: Tobi Ajila <atobia@ca.ibm.com>

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -249,9 +249,9 @@ endif
 
 # Adjust CRIU Support enablement flags.
 ifeq (true,$(OPENJ9_ENABLE_CRIU_SUPPORT))
-  FEATURE_SED_SCRIPT += $(call SedEnable,opt_criuSupport)
+  SPEC_SED_SCRIPT += $(call SedEnable,opt_criuSupport)
 else # OPENJ9_ENABLE_CRIU_SUPPORT
-  FEATURE_SED_SCRIPT += $(call SedDisable,opt_criuSupport)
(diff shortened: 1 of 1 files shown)


closed/OpenJ9.gmk
```diff
@@ -249,9 +249,9 @@ endif
 
 # Adjust CRIU Support enablement flags.
 ifeq (true,$(OPENJ9_ENABLE_CRIU_SUPPORT))
-  FEATURE_SED_SCRIPT += $(call SedEnable,opt_criuSupport)
+  SPEC_SED_SCRIPT += $(call SedEnable,opt_criuSupport)
 else # OPENJ9_ENABLE_CRIU_SUPPORT
-  FEATURE_SED_SCRIPT += $(call SedDisable,opt_criuSupport)
+  SPEC_SED_SCRIPT += $(call SedDisable,opt_criuSupport)
```
keithc-ca, line 254: This won't help because the file that needs to be modified is `f9.flags` which is neither `.feature` nor `.spec`.

tajila: Do you mean this? https://github.com/eclipse-openj9/openj9/blob/master/buildspecs/j9.flags#L1705

tajila: I guess I could leave it as is and add it to https://github.com/eclipse-openj9/openj9/blob/master/buildspecs/core.feature

keithc-ca: Yes, I think a default for the flag belongs in some `.feature` file, `core.feature` seems reasonable enough.

PR #342, opened 2021-07-07
Add --with-launcher-name and --with-product-name configure options

pshipton:
Port of https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/428


keithc-ca (approved):

keithc-ca:
Identical change as for jdk17; another build is unnecessary.

PR #347, opened 2021-07-14
Add libj9a2e.so to lib directory

pshipton:
Enables alternate builds such as mixed refs on z/OS without hard coding
the VM dir, compressedrefs or default.

Same change as https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/14

### DIFF
--- a/closed/custom/modules/java.base/Copy.gmk
+++ b/closed/custom/modules/java.base/Copy.gmk
@@ -76,6 +76,15 @@ $(call openj9_copy_shlibs, \
 	omrsig \
 	)
 
+ifeq ($(OPENJDK_TARGET_OS), zos)
+
+$(call openj9_copy_files_and_debuginfos, \
+	$(addsuffix /$(call SHARED_LIBRARY,j9a2e), \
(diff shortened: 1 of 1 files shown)


closed/custom/modules/java.base/Copy.gmk
```diff
@@ -80,6 +80,11 @@ $(call openj9_copy_shlibs, \
 
 ifneq (,$(filter windows zos,$(OPENJDK_TARGET_OS)))
 
+$(call openj9_copy_files_and_debuginfos, \
+	$(addsuffix /$(call SHARED_LIBRARY,j9a2e), \
+		$(OPENJ9_VM_BUILD_DIR) \
+		$(LIB_DST_DIR)))
```
keithc-ca, line 86: This is not appropriate for Windows.

pshipton: fixed

keithc-ca (approved):
No PR testing for z/OS, but similar enough to ibmruntimes/openj9-openjdk-jdk11#430.

PR #359, opened 2021-08-30
Ignore transient negative pool.ctl in getSurplusQueuedTaskCount()

keithc-ca:
See explanation in https://github.com/eclipse-openj9/openj9/issues/12073#issuecomment-908539871.

### DIFF
--- a/src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
+++ b/src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
@@ -33,6 +33,12 @@
  * http://creativecommons.org/publicdomain/zero/1.0/
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+ * ===========================================================================
+ */
+
 package java.util.concurrent;
 
 import java.lang.Thread.UncaughtExceptionHandler;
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
```diff
@@ -2277,7 +2283,8 @@ static int getSurplusQueuedTaskCount() {
             (pool = (wt = (ForkJoinWorkerThread)t).pool) != null &&
             (q = wt.workQueue) != null) {
             int p = pool.mode & SMASK;
-            int a = p + (int)(pool.ctl >> RC_SHIFT);
+            // use max to ignore transient negative
+            int a = p + Math.max((int)(pool.ctl >> RC_SHIFT), 0);
```
JasonFengJ9, line 2287: Should the ignoring transient negative be done on `a` w/ similar approach of
https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/5338a697e9b9ce138cfdc97b9c7dfd417dd31e3b/src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java#L2980-L2981

src/java.base/share/classes/java/util/concurrent/ForkJoinPool.java
```diff
@@ -2277,7 +2283,8 @@ static int getSurplusQueuedTaskCount() {
...
             int n = q.top - q.base;
             return n - (a > (p >>>= 1) ? 0 :
```
JasonFengJ9, line 2289: Did these condition checks miss the `==` corner-cases? i.e. the code should look like
```
            return n - (a >= (p >>>= 1) ? 0 :
                        a >= (p >>>= 1) ? 1 :
                        a >= (p >>>= 1) ? 2 :
                        a >= (p >>>= 1) ? 4 :
                        8);
```

JasonFengJ9 (commented):
Similar comments apply to the PRs in other extension repos.

keithc-ca:
Moved this to draft mode while we discuss the solution.

keithc-ca:
Either the test must change or the fork-join change will be different.

PR #364, opened 2021-09-27

JasonFengJ9:
Resolved conflicts:
```
	src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
```


keithc-ca:
This targets the openj9 branch, while the title says openj9-staging: I think it should be the latter.

JasonFengJ9:
> This targets the openj9 branch, while the title says openj9-staging: I think it should be the latter.

Fixed, please have another look.

keithc-ca (approved):

PR #365, opened 2021-09-28
Build AsyncGetCallTraceTest shared library on Linux

keithc-ca:
Issue: eclipse-openj9/openj9#12276.

### DIFF
--- a/closed/custom/Main.gmk
+++ b/closed/custom/Main.gmk
@@ -74,6 +74,9 @@ ALL_TARGETS += debug-image
 
 # If not cross-compiling, capture 'java -version' output.
 test-image-openj9 : exploded-image
+ifeq ($(call isTargetOs, linux), true)
+	@+$(MAKE) $(MAKE_ARGS) -f $(TOPDIR)/closed/AsyncGetCallTrace.gmk
+endif
 	@+$(OPENJ9_MAKE) openj9_test_image
 ifneq ($(COMPILE_TYPE), cross)
 	$(JDK_OUTPUTDIR)/bin/java -version 2>&1 | $(TEE) $(TEST_IMAGE_DIR)/openj9/java-version.txt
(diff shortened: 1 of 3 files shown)


closed/custom/Main.gmk
```diff
@@ -74,6 +74,9 @@ ALL_TARGETS += debug-image
 
 # If not cross-compiling, capture 'java -version' output.
 test-image-openj9 : exploded-image
+ifeq ($(call isTargetOs, linux), true)
```
JasonFengJ9, line 77: I think _AGCT is currently only available for linux x86-64_ [1].
@tajila Has this assumption been changed, and `AGCT` is available for other platforms such as `PPC` or `z`?

[1] https://github.com/eclipse-openj9/openj9/issues/12276#issuecomment-805986354

keithc-ca: It currently is implemented only for Linux x86-64, but we may as well build the test library for all Linux platforms so we're ready when other architectures are supported by OpenJ9.

JasonFengJ9: Okay, that works.
I noticed that OpenJ9 testimage already contains hotspot jtreg native `libFileUtils.so` & `libLingeredApp.so` w/o openj9 specific make file, any chance to build `libAsyncGetCallTraceTest.so` in similar fashion?

keithc-ca: Things like `libFileUtils.so` are created by a different makefile (`BuildTestLibNative.gmk`) for which all the required source is in our (filtered) source tree. The same cannot be said about `JtregNativeLibTest.gmk` which would normally build `libAsyncGetCallTraceTest.so` (it depends on source to generate `jvmti.h` and related things). I actually started down that path, but it would have been a much larger, and much more fragile, change.

pshipton (approved):

PR #366, opened 2021-09-30
Undo changes to US_ASCII for OpenJ9

keithc-ca:
Recent upstream changes should eliminate the benefit of changes specific to OpenJ9.

### DIFF
--- a/src/java.base/share/classes/sun/nio/cs/US_ASCII.java
+++ b/src/java.base/share/classes/sun/nio/cs/US_ASCII.java
@@ -23,11 +23,6 @@
  * questions.
  */
 
-/*
- * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2021 All Rights Reserved
- * ===========================================================================
- */
 package sun.nio.cs;
 
 import jdk.internal.access.JavaLangAccess;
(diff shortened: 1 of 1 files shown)


keithc-ca:
Jenkins test sanity plinux jdknext depends eclipse-openj9/openj9#13603

pshipton:
@vijaysun-omr @mpirvu fyi. We should run some converter benchmarking before and after this change to see how affects the perf.

vijaysun-omr:
@dchopra001 can you please help with this ? fyi @joransiu

vijaysun-omr:
@0xdaryl fyi

dchopra001:
`encodeASCII` is recognized by VP here: https://github.com/eclipse/omr/blob/1ec11640d4974dc68384c032b7118329bd379d7f/compiler/optimizer/ValuePropagationCommon.cpp#L2163.

However it only showed up in the Java 8 profiles. For Java 11 the encoding routine was changed and no longer calling this routine. If these changes aren't propagating to JDK8, then I don't think we will have an issue. However I can do another run with these changes implemented to see if we observe any difference.

pshipton:
These changes are only for Java 18 and later.

pshipton (approved):

PR #374, opened 2021-10-29
Add loadLibraryOnlyIfPresent to adopt NativeLibraries changes

JasonFengJ9:
Required by `NativeLibraries.loadLibraryOnlyIfPresent = ClassLoaderHelper.loadLibraryOnlyIfPresent()` which is introduced by `8275703: System.loadLibrary fails on Big Sur for libraries hidden from filesystem`.

fixes https://github.com/eclipse-openj9/openj9/issues/13827 

### DIFF
--- a/src/java.base/aix/classes/jdk/internal/loader/ClassLoaderHelper.java
+++ b/src/java.base/aix/classes/jdk/internal/loader/ClassLoaderHelper.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2020, 2020 All Rights Reserved
+ * (c) Copyright IBM Corp. 2020, 2021 All Rights Reserved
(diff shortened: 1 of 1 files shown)


pshipton:
This seems like it will be (and should be) fixed upstream. I think we should wait few days, and if it's not fixed then we can submit this to the OpenJDK mailing list. I won't merge it now because if it gets fixed upstream soon, we'll have to revert.

JasonFengJ9:
This `AIX` specific `ClassLoaderHelper.java` is not from `OpenJDK`, it was added via https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/204.

There were some related discussion at https://mail.openjdk.java.net/pipermail/ppc-aix-port-dev/2020-July/004114.html

pshipton (approved):

PR #378, opened 2021-11-24
zOS: Add 31/64-bit interoperability staging

joransiu:
- Copy the customized JNI header files to include31 directory to JDK_IMAGE base directory.
- Copy the libjvm31.so library and libjvm31.x sidedeck to server / j9vm.

Signed-off-by: Joran Siu <joransiu@ca.ibm.com>

### DIFF
--- a/closed/custom/modules/java.base/Copy.gmk
+++ b/closed/custom/modules/java.base/Copy.gmk
@@ -30,6 +30,28 @@ $(call openj9_copy_files,, \
 	$(OPENJ9_VM_BUILD_DIR)/include/jvmti.h \
 	$(INCLUDE_TARGET_DIR)/jvmti.h)
 
+ifeq (zos,$(OPENJDK_TARGET_OS))
+$(call openj9_copy_files,, \
+	$(OPENJ9_TOPDIR)/runtime/include/jni_convert.h \
+	$(INCLUDE_TARGET_DIR)/jni_convert.h)
+
+# Customized z/OS 31/64-bit interoperability JNI header files.
+# These files are generated as part of an OpenJ9 build.
+INCLUDE31_SOURCE_DIR := $(OPENJ9_VM_BUILD_DIR)/include31
+INCLUDE31_TARGET_DIR := $(INCLUDE_TARGET_DIR)/jni31
+$(call openj9_copy_files,, \
+	$(INCLUDE31_SOURCE_DIR)/jni.h \
+	$(INCLUDE31_TARGET_DIR)/jni.h)
+
+$(call openj9_copy_files,, \
+	$(INCLUDE31_SOURCE_DIR)/jni_convert.h \
+	$(INCLUDE31_TARGET_DIR)/jni_convert.h)
+
+$(call openj9_copy_files,, \
+	$(INCLUDE31_SOURCE_DIR)/jniport.h \
+	$(INCLUDE31_TARGET_DIR)/jniport.h)
+endif # zos
+
 # jitserver
 
 ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
@@ -77,14 +99,34 @@ $(call openj9_copy_shlibs, \
 	omrsig \
 	)
 
-ifeq ($(OPENJDK_TARGET_OS), zos)
+ifeq (zos,$(OPENJDK_TARGET_OS))
 
 $(call openj9_copy_files_and_debuginfos, \
 	$(addsuffix /$(call SHARED_LIBRARY,j9a2e), \
 		$(OPENJ9_VM_BUILD_DIR) \
 		$(LIB_DST_DIR)))
 
-endif
+$(call openj9_copy_files,, \
+	$(addsuffix libjsig.x, \
+		$(OPENJ9_VM_BUILD_DIR)/lib/ \
+		$(addprefix $(LIB_DST_DIR), / /j9vm/ /server/)))
+
+$(call openj9_copy_files,, \
+	$(addsuffix libjvm.x, \
+		$(OPENJ9_VM_BUILD_DIR)/lib/redirector/ \
+		$(addprefix $(LIB_DST_DIR), /j9vm/ /server/)))
+
+# 31/64-bit interoperability shim library and side deck files.
+$(call openj9_copy_files_and_debuginfos, \
+	$(addsuffix $(call SHARED_LIBRARY,jvm31), \
+		$(OPENJ9_VM_BUILD_DIR)/ \
+		$(addprefix $(LIB_DST_DIR), /j9vm/ /server/)))
+
+$(call openj9_copy_files,, \
+	$(addsuffix libjvm31.x, \
+		$(OPENJ9_VM_BUILD_DIR)/lib/ \
+		$(addprefix $(LIB_DST_DIR), /j9vm/ /server/)))
+endif # zos
 
 # static libraries that are needed on some platforms
 
--- a/closed/custom/modules/java.base/Copy.gmk
+++ b/closed/custom/modules/java.base/Copy.gmk
@@ -188,18 +188,18 @@ endif # OPENJ9_ENABLE_DDR
 # Optionally copy OpenSSL Crypto Library
 # To bundle first search for openssl 1.1.x library, if not found, search for 1.0.x
 ifneq ($(OPENSSL_BUNDLE_LIB_PATH), )
-  ifeq ($(OPENJDK_TARGET_OS), linux)
+  ifeq (linux,$(OPENJDK_TARGET_OS))
     LIBCRYPTO_NAMES := libcrypto.so.1.1 libcrypto.so.1.0.0
-  else ifeq ($(OPENJDK_TARGET_OS), macosx)
+  else ifeq (macosx,$(OPENJDK_TARGET_OS))
     LIBCRYPTO_NAMES := libcrypto.1.1.dylib libcrypto.1.0.0.dylib
-  else ifeq ($(OPENJDK_TARGET_OS), windows)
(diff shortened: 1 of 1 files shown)


joransiu:
This change depends on : https://github.com/eclipse-openj9/openj9/pull/13365 to have libjvm31.x and libjvm31.so being built in CMake enabled builds.  It has some additional JNI header and side-deck related changes that is in `openj9-openjdk-jdk11` extensions repo, which is not in this one -- I pulled them along to keep things consistent.   

Once this PR is reviewed, I'll backport to `openj9-openjdk-jdk11` and `openj9-openjdk-jdk17`.

closed/custom/Images-pre.gmk
```diff
@@ -18,6 +18,20 @@
...
+	INCLUDE31_SOURCE_DIR := $(OPENJ9_VM_BUILD_DIR)/include31
+
+	# Stage into JDK base image 'include31' subdirectory
+	$(eval $(call SetupCopyFiles, JDK_COPY_INCLUDE31_HEADERS, \
+		SRC := $(INCLUDE31_SOURCE_DIR), \
+		FILES := $(call FindFiles, $(INCLUDE31_SOURCE_DIR)/), \
+		DEST := $(JDK_IMAGE_DIR)/include31/, \
+	))
+	JDK_TARGETS += $(JDK_COPY_INCLUDE31_HEADERS)
+endif
```
keithc-ca, line 33: These lines cannot begin with tabs - please use spaces.

Please remove the trailing slashes in the call to `FindFiles` and in the value of `DEST`, and add a comment on the `endif` line.

closed/custom/modules/java.base/Copy.gmk
```diff
@@ -30,6 +30,28 @@ $(call openj9_copy_files,, \
 	$(OPENJ9_VM_BUILD_DIR)/include/jvmti.h \
 	$(INCLUDE_TARGET_DIR)/jvmti.h)
 
+ifeq ($(OPENJDK_TARGET_OS), zos)
+$(call openj9_copy_files,, \
+$(OPENJ9_TOPDIR)/runtime/include/jni_convert.h \
+$(INCLUDE_TARGET_DIR)/jni_convert.h)
+
+# Customized z/OS 31/64-bit Interoperability JNI header files.
+# These files are generated as part of OpenJ9 build.
```
keithc-ca, line 39: Why is 'Interoperability' capitalized?
Please add the missing 'an' in 'part of an OpenJ9 build'.

closed/custom/modules/java.base/Copy.gmk
```diff
@@ -30,6 +30,28 @@ $(call openj9_copy_files,, \
...
+INCLUDE31_SOURCE_DIR := $(OPENJ9_VM_BUILD_DIR)/include31
+INCLUDE31_TARGET_DIR := $(INCLUDE_TARGET_DIR)/../include31
```
keithc-ca, line 41: Have you verified that using the name `include31` works with `jlink`?

closed/custom/modules/java.base/Copy.gmk
```diff
@@ -30,6 +30,28 @@ $(call openj9_copy_files,, \
 	$(OPENJ9_VM_BUILD_DIR)/include/jvmti.h \
 	$(INCLUDE_TARGET_DIR)/jvmti.h)
 
+ifeq ($(OPENJDK_TARGET_OS), zos)
+$(call openj9_copy_files,, \
+$(OPENJ9_TOPDIR)/runtime/include/jni_convert.h \
+$(INCLUDE_TARGET_DIR)/jni_convert.h)
```
keithc-ca, line 36: Please indent lines 35, 36.

closed/custom/modules/java.base/Copy.gmk
```diff
@@ -84,6 +106,26 @@ $(call openj9_copy_files_and_debuginfos, \
...
+$(call openj9_copy_files_and_debuginfos, \
+	$(addsuffix $(call SHARED_LIBRARY,jvm31), \
+		$(OPENJ9_VM_BUILD_DIR)/ \
+		$(addprefix $(LIB_DST_DIR), /j9vm/ /server/)))
+
+$(call openj9_copy_files,, \
+	$(addsuffix libjvm31.x, \
+		$(OPENJ9_VM_BUILD_DIR)/lib/ \
+		$(addprefix $(LIB_DST_DIR), /j9vm/ /server/)))
 endif
```
keithc-ca, line 129: This should be commented:
```
endif # zos
```

keithc-ca (changes requested):

closed/custom/modules/java.base/Copy.gmk
joransiu: Thanks @keithc-ca.  This is a good point -- I hadn't realized jlink images include the headers.  I'll confirm.

keithc-ca: It's actually the `.jmod` files we are concerned about; they are inputs to `jlink` for producing (possibly custom) images. Some paths are treated specially in `.jmod` files (e.g. `classes/*`); we just need to play by the rules.

joransiu: Right. I see the following in the `java.base.jmod`, not immediately sure how to get additional directories to be included.

```sh
> jar tf java.base.jmod | grep include                                                                                                                                                                     
include/jni.h
include/jvmticmlr.h
include/zos/jni_md.h
include/ibmjvmti.h
include/jvmti.h
include/jni_convert.h
include/classfile_constants.h
```

joransiu: Looks like it's related to this: https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/make/CreateJmods.gmk#L52-L53  though not sure how `include` itself is specified for java.base.

joransiu: Looks like the module specific header files are staged into: ` $(SUPPORT_OUTPUTDIR)/modules_include/$(MODULE)` based on:

https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/make/common/modules/CopyCommon.gmk#L36-L63

joransiu: From what I can tell, jmod appears to support only `--header-files` and will stage files in the given directory into `include` directory of the output jmod file.  Looks like we'd have to modify jmod itself if we want a parallel `include31` directory.  Alternatively, we can have `include31` as a subdirectory within `include` folder by staging the files into modules_include target -- and just document the updated location in migration docs.   There will be another change  to compile /link commands of JNI natives for the libjvm31.x sidedeck file from Java 8, due to lib/s390x/j9vm -> lib/j9vm.   @keithc-ca @pshipton , thoughts?

keithc-ca: I think it would be simpler to put the new header files in a new subdirectory of `include`, e.g. `include/jni31` or similar. Applications that need to use the 31-bit shim API would compile with `{sdk}/include/jni31` on the include path (instead of `{sdk}/include31`).

keithc-ca: BTW, a `.jmod` file is not really in in `jar` format; you should use `jmod`:
```
$ ./bin/jmod list jmods/java.base.jmod | grep include | sort
include/classfile_constants.h
include/ibmjvmti.h
include/jni.h
include/jvmticmlr.h
include/jvmti.h
include/linux/jni_md.h
```

joransiu: Thanks @keithc-ca.  I got some initial feedback from 31/64-bit users... no one is against changing it to subdirectory of `include/` for Java 11+.  In some ways, `include/jni31` as you suggested is cleaner.   This simplifies this PR, as none of the `Image-pre.gmk` changes are needed.  I'll touch base with some other users who will be back from Thanksgiving next week to see if anyone raises additional concerns.

joransiu: Got feedback from users that the suggested change to include/jni31 was acceptable.  I'll update this PR with the necessary changes, and also address the other review comments.  Thanks.

joransiu:
@keithc-ca : I've removed all the Image-Pre.gmk changes, and updated the target directory of the 31-bit JNI headers to include/jni31 of java.base in the latest commit.

closed/custom/modules/java.base/Copy.gmk
```diff
@@ -30,6 +30,28 @@ $(call openj9_copy_files,, \
 	$(OPENJ9_VM_BUILD_DIR)/include/jvmti.h \
 	$(INCLUDE_TARGET_DIR)/jvmti.h)
 
+ifeq ($(OPENJDK_TARGET_OS), zos)
```
keithc-ca, line 33: nit: for consistency, the arguments should be reversed (see [OpenJ9.gmk](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/closed/OpenJ9.gmk#L94)):
```
ifeq (zos,$(OPENJDK_TARGET_OS))
```
and the endif line should just be:
```
endif # zos
```
Same goes for the conditional block which starts on line 102.

joransiu: Updated under 48c8f92. Thanks!

keithc-ca (approved):

keithc-ca:
Please repeat the second part of ibmruntimes/openj9-openjdk-jdk11#459 to reduce differences between versions.

keithc-ca:

keithc-ca:
@pshipton I'll let you merge this when you're satisfied with the changes.

pshipton (approved):

PR #382, opened 2021-12-06
GetCallerClassTest: Accept OpenJ9's InternalError Message

babsingh:
InternalError message in GetCallerClassTest differs between JVMs:
- OpenJ9: "Caller is not annotated as @sun.reflect.CallerSensitive"
- OpenJDK: "CallerSensitive annotation expected"

GetCallerClassTest has been updated to accept OpenJ9's InternalError message.

Closes: https://github.com/eclipse-openj9/openj9/issues/13998

### DIFF
--- a/test/jdk/jdk/internal/reflect/Reflection/GetCallerClassTest.java
+++ b/test/jdk/jdk/internal/reflect/Reflection/GetCallerClassTest.java
@@ -22,6 +22,12 @@
  */
 
 /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+ * ===========================================================================
+ */
+
+/*
  * @test
  * @bug 8010117
  * @summary Test if the VM enforces Reflection.getCallerClass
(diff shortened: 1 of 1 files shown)


babsingh:
While testing these changes locally, I noticed segfaults in the JVM, which was built using the latest openjdk-next, openj9 and omr repos. These test changes won't cause the segfault. Just to be safe, I will keep this PR in the WIP state until I figure out the cause of the segfaults,.

pshipton:
You could just use `startsWith() || startsWith()` to keep it simpler.

babsingh:
> You could just use `startsWith() || startsWith()` to keep it simpler.

Updated the PR to reflect the suggested changes.

pshipton:
lgtm. Regardless of the segfaults, if this is removed from draft then I would merge it.

babsingh:
The segfault is a new failure, which is observed after this fix is applied. I have opened https://github.com/eclipse-openj9/openj9/issues/14097 to investigate the segfault.

This PR should be good to merge.

pshipton (approved):

PR #383, opened 2021-12-08
[JDKnext] Set the default class name when compiling XSL templates

ChengJin01:
The changes ensure the default class name is always set
to the bytecode so as to address the corner case when
the base name converted from the system ID is null.

### DIFF
--- a/src/java.xml/share/classes/com/sun/org/apache/xalan/internal/xsltc/compiler/XSLTC.java
+++ b/src/java.xml/share/classes/com/sun/org/apache/xalan/internal/xsltc/compiler/XSLTC.java
@@ -18,6 +18,12 @@
  * limitations under the License.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+ * ===========================================================================
+ */
+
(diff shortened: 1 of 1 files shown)


ChengJin01:
Reviewer: @pshipton
FYI: @tajila

pshipton:
Pls fix the conflict, and add details about the OpenJDK mailing list posts about this change.

pshipton:
Looks like the similar change was added to OpenJDK so this is no longer necessary. Pls backport the change as-is to jdk11 and jdk17.

pshipton:
The change is https://bugs.openjdk.java.net/browse/JDK-8276657 which refers to https://bugs.openjdk.java.net/browse/JDK-8276241

pshipton:
In addition to updating https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/460 and https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/52 to backport the OpenJDK change, pls request at OpenJDK for the backport for jdk11 and jdk17.

ChengJin01:
The request was sent to OpenJDK at https://mail.openjdk.java.net/pipermail/core-libs-dev/2021-December/084523.html.

PR #386, opened 2021-12-13
CMake: Don't use separate cross compile spec

dnakamura:
Signed-off-by: Devin Nakamura <devinn@ca.ibm.com>

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -338,7 +338,7 @@ run-preprocessors-j9 : \
 ifeq (true,$(OPENJ9_ENABLE_CMAKE))
 
   CMAKE_ARGS := \
-	-C $(OPENJ9_TOPDIR)/runtime/cmake/caches/$(OPENJ9_BUILDSPEC).cmake \
+	-C $(OPENJ9_TOPDIR)/runtime/cmake/caches/$(patsubst %_cross,%,$(OPENJ9_BUILDSPEC)).cmake \
 	-DCMAKE_TOOLCHAIN_FILE="$(OUTPUTDIR)/toolchain.cmake" \
 	-DBOOT_JDK="$(BOOT_JDK)" \
 	-DBUILD_ID=$(BUILD_ID) \


closed/OpenJ9.gmk
```diff
@@ -338,7 +338,7 @@ run-preprocessors-j9 : \
 ifeq (true,$(OPENJ9_ENABLE_CMAKE))
 
   CMAKE_ARGS := \
-	-C $(OPENJ9_TOPDIR)/runtime/cmake/caches/$(OPENJ9_BUILDSPEC).cmake \
+	-C $(OPENJ9_TOPDIR)/runtime/cmake/caches/$(patsubst %_cross,%,$(OPENJ9_BUILDSPEC)).cmake \
```
keithc-ca, line 341: This seems reasonable, but how will `$(COMPILE_TYPE) == cross` be communicated to openj9/omr builds?

dnakamura: https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/334f7648e1ad3a503f033b8c93fb92f1aeae7d33/closed/autoconf/toolchain.cmake.in#L22

keithc-ca: It seems `CMAKE_SYSTEM_NAME` is used in situations when we're not cross-compiling, so I don't think that answers my question.

dnakamura: we only set `CMAKE_SYSTEM_NAME` if we are cross compiling, or compiling on windows, which CMake considers to be a cross compile (cygwin->win32). Any time that you manually set `CMAKE_SYSTEM_NAME` (even if you set it to the same as the current system name), cmake will treat the build as a cross compile and set `CMAKE_CROSSCOMPILING`.  In addition I have a PR  (eclipse/omr#6272) which defines an `OMR_CROSSCOMPILING` variable which has more intuitive behavior on windows builds.

keithc-ca: There are a number of uses of `CMAKE_SYSTEM_NAME` that don't appear to be related to cross-compiling; see [1] and [2].

`OMR_CROSSCOMPILING` seems a reasonable addition. I think this should wait until it can define that symbol properly.

[1] https://github.com/eclipse/omr/blob/master/cmake/modules/OmrDetectSystemInformation.cmake#L106
[2] https://github.com/eclipse/omr/blob/master/cmake/modules/OmrDetectSystemInformation.cmake#L160

dnakamura: By default  cmake will set `CMAKE_SYSTEM_NAME` to `CMAKE_HOST_SYSTEM_NAME` ie assuming a normal non-cross compile  (https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_NAME.html)

keithc-ca: That just seems to reinforce my assertion that `CMAKE_SYSTEM_NAME` is not by itself an indication that we're cross-compiling.

The document you reference also says:
> In this case the CMAKE_SYSTEM_VERSION variable must also be set explicitly.

which is something we're _not_ doing.

dnakamura: >That just seems to reinforce my assertion that CMAKE_SYSTEM_NAME is not by itself an indication that we're cross-compiling.

Thats correct. There seems to be a miscommunication somewhere. On the CMake side, we determine if we are cross compiling by looking at `CMAKE_CROSSCOMPILING` / `OMR_CROSSCOMPILING` . However `CMAKE_CROSSCOMPILING` is not set directly. It gets automatically by CMake when `CMAKE_SYSTEM_NAME` is explicitly set.

>In this case the CMAKE_SYSTEM_VERSION variable must also be set explicitly.
I read that as meaning that CMake will not automatically provide a value for you. However, since this variable doesn't do much on most platforms, it is safe to omit it. Indeed the official examples omit this variable https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-linux

dnakamura: @keithc-ca eclipse/omr#6272 has landed now. Is this PR good to go?

keithc-ca: I'm not sure I have the complete picture yet, but this seems safe to merge now.

keithc-ca (approved):

keithc-ca:
@dnakamura Will you be creating similar pull requests for the remaining versions (8, 11, 17 & 18)?

PR #387, opened 2021-12-16
Adopt RI java.lang.Thread/ThreadGroup w/ --enable-openjdk-thread-support

JasonFengJ9:
Apply `OpenJ9` specific fields/methods:
Additional fields:
`started/stopCalled/systemThreadGroup/tidCount/mainGroup/NO_REF/createCount/deadInterrupt`;
Additional methods:
`completeInitialization()/uncaughtException()/newName()/initialize()/Thread()/isDead()/cleanup()`
Modification to existing field modifiers and methods.
Added `--enable-openjdk-thread-support` option.

The `java.lang.Thread` code modification/addition are contributed by IBM under the GPLv2+CE license.

Note: OpenJ9 patch could be reduced further which will be addressed in separated PRs. This can enable project Loom work.

Depends on https://github.com/eclipse-openj9/openj9/pull/14167

fyi @pshipton @0xdaryl


tajila:
@JasonFengJ9 Are the existing Thread() constructors needed? If not can you please remove them or throw an exception in them so we can ensure only the j9 constructors are used.

For this I don't mean the constructs that redirect to another constructor.

JasonFengJ9:
> Are the existing Thread() constructors needed? If not can you please remove them or throw an exception in them so we can ensure only the j9 constructors are used.
For this I don't mean the constructs that redirect to another constructor.

This JCL patch has three J9 constructors:
* `private Thread(String vmName, Object vmThreadGroup, int vmPriority, boolean vmIsDaemon)` is used by the VM for the threads attached through JNI;
*  `Thread(Runnable runnable, String threadName, boolean isSystemThreadGroup, boolean inheritThreadLocals, boolean isDaemon, ClassLoader contextClassLoader)` is invoked by `J9VMInternals.completeInitialization()`
*  `private Thread(ThreadGroup group, Runnable runnable, String threadName, AccessControlContext acc, boolean inheritThreadLocals)` is used by package access constructor above.

These constructors are invoked at VM early booting stage w/ OpenJ9 specifics. Further improvement is still in progress to use RI constructors instead, or reduce J9 patches.
This PR serves an intermediate solution to enable project Loom consume RI `java.lang.Thread`. 

All RI constructors are needed from static code analysis perspective.

tajila:
Okay, thanks for explaining

tajila:
Jenkins copyright check

tajila:
jenkins test sanity zlinux jdk17

pshipton:
Before this is merged, there is a license issue. The commit message (not just the PR) needs to be updated to state that the OpenJ9 code is not copied from OpenJ9 (EPLv2) to OpenJDK (GPLv2+CPE), but being contributed by IBM under the new license.

JasonFengJ9:
Updated PR/commit messages w/ _The java.lang.Thread code modification/addition are contributed by IBM
under the GPLv2+CE license._

Note: The PR build/test via https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/387#issuecomment-1006637597 all passed.

closed/autoconf/custom-hook.m4
```diff
@@ -340,6 +341,25 @@ AC_DEFUN([OPENJ9_PLATFORM_EXTRACT_VARS_FROM_CPU],
   esac
 ])
 
+AC_DEFUN([OPENJ9_CONFIGURE_OJDKTHREAD_SUPPORT],
+[
+  AC_MSG_CHECKING([for OpenJDK Thread support])
+  AC_ARG_ENABLE([ojdkthread-support], [AS_HELP_STRING([--enable-ojdkthread-support], [enable OpenJDK Thread support @<:@disabled@:>@])])
+  OPENJ9_ENABLE_OJDKTHREAD_SUPPORT=false
```
keithc-ca, line 348: Can we use `OPENJ9_ENABLE_OPENJDK_THREAD_SUPPORT` instead of `OPENJ9_ENABLE_OJDKTHREAD_SUPPORT`, please? This would be consistent with `OPENJ9_ENABLE_OPENJDK_METHODHANDLES` we had when migrating the method handle implementation.

Ideally, the OpenJ9 JPP configuration flag should also not be so abbreviated.

closed/autoconf/custom-hook.m4
```diff
@@ -340,6 +341,25 @@ AC_DEFUN([OPENJ9_PLATFORM_EXTRACT_VARS_FROM_CPU],
   esac
 ])
 
+AC_DEFUN([OPENJ9_CONFIGURE_OJDKTHREAD_SUPPORT],
```
keithc-ca, line 344: Please move this addition after `OPENJ9_CONFIGURE_JITSERVER` so the definitions are (closer to being) sorted?

closed/autoconf/custom-hook.m4
JasonFengJ9: There weren't much precedents for migrating RI implementation requiring additional configuration options. Out of respect the naming `OPENJ9_ENABLE_OPENJDK_METHODHANDLES`,  `OPENJ9_ENABLE_OJDKTHREAD_SUPPORT` seems concise than `OPENJ9_ENABLE_OPENJDK_THREAD_SUPPORT` while avoiding double `OPEN`s. Along with JPP flag `OJDKTHREAD_SUPPORT`, these are also based on `--enable-ojdkthread-support` proposed by @tajila .

closed/autoconf/custom-hook.m4
JasonFengJ9: sure.

JasonFengJ9: Re-ordered `AC_DEFUN`s.

closed/autoconf/custom-hook.m4
keithc-ca: The project is called 'openjdk' we should not be mangling that name (just look at the name of this repo for reinforcement).

I agree the configure option name should be consistent with the macro name, so in summary, the option should be `--enable-openjdk-thread-support` and the macro should be `OPENJ9_ENABLE_OPENJDK_THREAD_SUPPORT`.

BTW, the two 'open' substrings refer to different projects (openj9 vs. openjdk).

JasonFengJ9: renamed.

JasonFengJ9:
Jenkins copyright check

closed/autoconf/custom-hook.m4
```diff
@@ -49,10 +49,185 @@ AC_DEFUN_ONCE([CUSTOM_EARLY_HOOK],
   OPENJ9_CONFIGURE_WARNINGS
   OPENJ9_CONFIGURE_JITSERVER
   OPENJ9_CONFIGURE_INLINE_TYPES
+  OPENJ9_CONFIGURE_OPENJDK_THREAD_SUPPORT
   OPENJ9_THIRD_PARTY_REQUIREMENTS
   OPENJ9_CHECK_NASM_VERSION
 ])
 
+AC_DEFUN([CONFIGURE_OPENSSL],
```
keithc-ca, line 57: This is not in its proper location (after reordering).

keithc-ca (changes requested):
Please apply suggested formatting changes throughout `Thread.java`.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -193,12 +196,12 @@ private static synchronized int nextThreadNum() {
      * not specify a stack size.  It is up to the VM to do whatever it
      * likes with this number; some VMs will ignore it.
      */
-    private final long stackSize;
+    private long stackSize = 0;
```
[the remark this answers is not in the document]
JasonFengJ9: removed.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -782,38 +771,32 @@ public Thread(ThreadGroup group, Runnable target, String name,
...
-            started = true;
-        } finally {
-            try {
-                if (!started) {
-                    group.threadStartFailed(this);
-                }
+		boolean success = false;
+
+		try {
+			synchronized(blockerLock) {
```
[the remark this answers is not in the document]
JasonFengJ9: added a space.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -782,38 +771,32 @@ public Thread(ThreadGroup group, Runnable target, String name,
...
+
+				group.add(this);
+				startImpl();
+				success = true;
+			}
+		} finally {
+			try {
+				if (!success && !started) {
+					group.threadTerminated(this);
+				}
```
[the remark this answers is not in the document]
JasonFengJ9: Updated.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -924,23 +907,45 @@ private void exit() {
...
+		if (!isDead()){
+			stopWithThrowable(new ThreadDeath());
+		}
+	}
+
+	private final synchronized void stopWithThrowable(Throwable throwable) {
+		checkAccess();
+		if (currentThread() != this) {
+			SecurityManager currentManager = System.getSecurityManager();
+			if (currentManager != null)	{
```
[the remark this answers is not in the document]
JasonFengJ9: updated.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -924,23 +907,45 @@ private void exit() {
...
+				currentManager.checkPermission(SecurityConstants.STOP_THREAD_PERMISSION);
+			}
+		}
+
+		synchronized(blockerLock) {
```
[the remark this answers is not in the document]
JasonFengJ9: added.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1078,9 +1082,16 @@ public boolean isInterrupted() {
...
     public final void suspend() {
-        checkAccess();
-        suspend0();
-    }
+		checkAccess();
+		if (currentThread() == this) suspendImpl();
+		else {
+			synchronized( blockerLock ) {
+				suspendImpl();
+			}
```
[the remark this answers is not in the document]
JasonFengJ9: updated.

closed/autoconf/custom-hook.m4
JasonFengJ9: `CONFIGURE_OPENSSL` is in right order unless changing it to `OPENJ9_CONFIGURE_OPENSSL` for `AC_DEFUN` name consistency. Renamed it.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -924,23 +907,45 @@ private void exit() {
...
-        // not-NEW because we hold the lock.
-        if (threadStatus != 0) {
-            resume(); // Wake up thread if it was suspended; no-op otherwise
-        }
-
-        // The VM can handle all thread states
-        stop0(new ThreadDeath());
-    }
+		/* the only case we don't want to do the check is if the thread has been started but is now dead */
+		if (!isDead()){
```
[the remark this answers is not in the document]
JasonFengJ9: added.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -924,23 +907,45 @@ private void exit() {
...
+		if (currentThread() != this) {
+			SecurityManager currentManager = System.getSecurityManager();
+			if (currentManager != null)	{
+				currentManager.checkPermission(SecurityConstants.STOP_THREAD_PERMISSION);
+			}
+		}
+
+		synchronized(blockerLock) {
+			if (throwable != null){
+				if (!started){
```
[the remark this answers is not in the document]
JasonFengJ9: added

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -2097,4 +2129,265 @@ public boolean equals(Object obj) {
...
+                inheritableThreadLocals = ThreadLocal.createInheritedMap(parentThread.inheritableThreadLocals);
+            }
+
+            @SuppressWarnings("removal")
+            final SecurityManager sm = System.getSecurityManager();
+            final Class implClass = getClass();
+            final Class thisClass = Thread.class;
+            if ((sm != null) && (implClass != thisClass)) {
+                boolean override = ((Boolean)AccessController.doPrivileged(new PrivilegedAction() {
+                    public Object run() {
```
keithc-ca, line 2220: No raw types, please. This should begin:
```
        boolean override = AccessController.doPrivileged(new PrivilegedAction<Boolean>() {
             @Override
             public Boolean run() {
```
A closing `)` needs to be removed on line 2237.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -2097,4 +2129,265 @@ public boolean equals(Object obj) {
...
+                        try {
+                            Method method = implClass.getMethod("getContextClassLoader", new Class[0]);
```
keithc-ca, line 2222: The second parameter can be omitted:
```
    Method method = implClass.getMethod("getContextClassLoader");
```

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -2097,4 +2129,265 @@ public boolean equals(Object obj) {
...
+                            if (method.getDeclaringClass() != thisClass) {
+                                return Boolean.TRUE;
+                            }
+                        } catch (NoSuchMethodException e) {
+                        }
+                        try {
+                            Method method = implClass.getDeclaredMethod("setContextClassLoader", new Class[]{ClassLoader.class});
```
keithc-ca, line 2229: The second parameter can be simplified:
```
    Method method = implClass.getDeclaredMethod("setContextClassLoader", ClassLoader.class);
```

keithc-ca (changes requested):

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -782,29 +772,23 @@ public Thread(ThreadGroup group, Runnable target, String name,
...
+        boolean success = false;
 
-        boolean started = false;
         try {
-            start0();
-            started = true;
+            synchronized (blockerLock) {
+                if (started) {
+                    // K0341 = Thread is already started
+                    throw new IllegalThreadStateException(com.ibm.oti.util.Msg.getString("K0341"));
```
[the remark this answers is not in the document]
JasonFengJ9: Removed the message string to match RI.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1593,24 +1627,20 @@ public void setContextClassLoader(ClassLoader cl) {
...
             }
-            // optimization so we do not call into the vm for threads that
-            // have not yet started or have terminated
+        }
+        Throwable t;
+
+        synchronized (blockerLock) {
             if (!isAlive()) {
-                return EMPTY_STACK_TRACE;
+                return new StackTraceElement[0];
```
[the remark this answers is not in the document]
JasonFengJ9: restored.

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: updated.

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: removed.

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: updated.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1656,19 +1686,15 @@ public void setContextClassLoader(ClassLoader cl) {
...
-            StackTraceElement[] stackTrace = traces[i];
-            if (stackTrace != null) {
-                m.put(threads[i], stackTrace);
-            }
-            // else terminated so we don't put it in the map
+        // Allow room for more Threads to be created before calling enumerate()
+        int count = systemThreadGroup.activeCount() + 20;
+        Thread[] threads = new Thread[count];
+        count = systemThreadGroup.enumerate(threads);
+        Map<Thread, StackTraceElement[]> result = new HashMap<>((count * 4) / 3);
```
keithc-ca, line 1693: Why the `4/3` scaling here? I suggest this should just use `new HashMap<>(count)`; that class is in a better position to deal appropriately with the capacity request.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -782,29 +772,22 @@ public Thread(ThreadGroup group, Runnable target, String name,
...
+
+                group.add(this);
+                startImpl();
+                success = true;
+            }
         } finally {
             try {
-                if (!started) {
-                    group.threadStartFailed(this);
+                if (!success && !started) {
```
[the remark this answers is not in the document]
JasonFengJ9: No, for this case `started` is set by VM within `startImpl()` which is guarded by `synchronized start()`, hence there is no need to access it in a `synchronized` block.
`!success` is not sufficient, according to the code comment, `!started` was introduced to fix a deadlock when a thread fails to start and another thread is inspecting the same thread group.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -924,24 +907,48 @@ private void exit() {
...
+            @SuppressWarnings("removal")
+            SecurityManager currentManager = System.getSecurityManager();
+            if (currentManager != null) {
+                currentManager.checkPermission(SecurityConstants.STOP_THREAD_PERMISSION);
+            }
         }
 
-        // The VM can handle all thread states
-        stop0(new ThreadDeath());
+        synchronized (blockerLock) {
```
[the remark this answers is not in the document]
JasonFengJ9: This is consistent with OpenJ9 current code, `blockerLock` is an alias of `lock`.

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: This is current OpenJ9 behavour [1].
I think a change should be proposed to OpenJ9 head stream first, and ported to this afterwards.

[1] https://github.com/eclipse-openj9/openj9/blob/fc95e91365bc7a38f5abe0fd8b29b4672cb69a45/jcl/src/java.base/share/classes/java/lang/Thread.java#L1418

src/java.base/share/classes/java/lang/Thread.java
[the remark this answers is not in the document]
keithc-ca: I don't think this is correct. `start` is set fairly early in `startImpl()` and may fail afterward. In that case the thread, which has not really started is not removed from the group.

Perhaps the body of this method should be:
```
        synchronized (blockerLock) {
            if (started) {
                throw new IllegalThreadStateException();
            }
            boolean success = false;
            try {
                group.add(this);
                startImpl();
                success = true;
            } finally {
                if (!success) {
                    group.threadTerminated(this);
                }
            }
        }
```
Notice:
* the `finally` block does not have to consider the `IllegalThreadStateException`
* if the thread is not fully created it is removed from the group

src/java.base/share/classes/java/lang/Thread.java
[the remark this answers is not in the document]
keithc-ca: I mean some methods synchronize on *both* the thread object and `blockerLock` and other methods only use `blockerLock`.

src/java.base/share/classes/java/lang/Thread.java
keithc-ca: Ok: I created https://github.com/eclipse-openj9/openj9/issues/14244.

pshipton: I'm fine with the change, we can just make it here and we'll get the same change into the head stream shortly.

src/java.base/share/classes/java/lang/Thread.java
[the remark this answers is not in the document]
JasonFengJ9: At first glance, the code proposed added `group.threadTerminated(this)` into synchronized block which wasn't there and required more analysis for such addition.
I think any change to remove `!started` has to make sure no regression on the initial deadlock problem, which also should be proposed in OpenJ9 head stream first, and ported here later.

src/java.base/share/classes/java/lang/Thread.java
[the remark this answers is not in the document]
JasonFengJ9: This is current OpenJ9 behaviour [1].

[1] https://github.com/eclipse-openj9/openj9/blob/fc95e91365bc7a38f5abe0fd8b29b4672cb69a45/jcl/src/java.base/share/classes/java/lang/Thread.java#L1240-L1250

src/java.base/share/classes/java/lang/Thread.java
pshipton: i.e. https://github.com/eclipse-openj9/openj9/pull/14245

src/java.base/share/classes/java/lang/Thread.java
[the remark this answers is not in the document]
keithc-ca: Isn't the whole point of this PR to integrate with the expanded set of use-cases of this class, which means the current OpenJ9 behavior isn't necessarily sufficient.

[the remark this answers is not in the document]
JasonFengJ9: If that's the case, a change should be proposed in OpenJ9 head stream first, and ported to here afterwards.

src/java.base/share/classes/java/lang/Thread.java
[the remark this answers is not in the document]
keithc-ca: `ThreadGroup.add()` synchronizes on the `ThreadGroup` object so calling `ThreadGroup.threadTerminated()`, which does the same, should be safe from a deadlock perspective.

As I review my proposed change, I realize that the call to `ThreadGroup.add()` should precede the `try` statement (a call to `threadTerminated()` should only occur if the call to `add()` completes normally):
```
    synchronized (blockerLock) {
        if (started) {
            throw new IllegalThreadStateException();
        }
        boolean success = false;
        group.add(this);
        try {
            startImpl();
            success = true;
        } finally {
            if (!success) {
                group.threadTerminated(this);
            }
        }
    }
```
A this point, it's possible for the local variable `success` to be removed.

I created https://github.com/eclipse-openj9/openj9/issues/14260 to track this.

tajila:
What is the status of this? @JasonFengJ9

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: @keithc-ca any comments on https://github.com/eclipse-openj9/openj9/pull/14245#discussion_r782307818?

src/java.base/share/classes/java/lang/Thread.java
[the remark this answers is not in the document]
JasonFengJ9: The change can be ported here if OpenJ9 head stream accepts it.

JasonFengJ9:
> What is the status of this?

Resumed the discussion if https://github.com/eclipse-openj9/openj9/pull/14245 is to be ported here at https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/387#discussion_r788108348

src/java.base/share/classes/java/lang/Thread.java
keithc-ca: See https://github.com/eclipse-openj9/openj9/pull/14245#discussion_r788124046.

src/java.base/share/classes/java/lang/Thread.java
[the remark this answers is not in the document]
keithc-ca: That works for me.

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: Ported `OpenJ9` change here.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -782,29 +772,22 @@ public Thread(ThreadGroup group, Runnable target, String name,
...
+                group.add(this);
+                startImpl();
+                success = true;
+            }
         } finally {
             try {
-                if (!started) {
-                    group.threadStartFailed(this);
+                if (!success && !started) {
+                    group.threadTerminated(this);
```
keithc-ca, line 790: I think this should call `threadStartFailed()` instead.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -782,29 +772,22 @@ public Thread(ThreadGroup group, Runnable target, String name,
...
                 }
             } catch (Throwable ignore) {
```
keithc-ca, line 792: The comment in this `catch` block should be improved. The reference to `start0()` should be `startImpl()` and it should make it clearer that the throwable being ignored is from `ThreadGroup.threadStartFailed()`.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -782,29 +772,22 @@ public Thread(ThreadGroup group, Runnable target, String name,
...
-        boolean started = false;
         try {
-            start0();
-            started = true;
+            synchronized (blockerLock) {
+                if (started) {
+                    throw new IllegalThreadStateException();
+                }
+
+                group.add(this);
```
keithc-ca, line 783: The call to `ThreadGroup.add()` should not be within the `try` block; `ThreadGroup.threadStartFailed()` should not be called if `add()` did not succeed.

keithc-ca (changes requested):

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: This is existing OpenJ9 behaviours, a change moving `group.add(this)` out of `try` block would be accepted in `OpenJ9` head stream first.

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: `threadStartFailed ` only removes the thread from the group and update `nUnstartedThreads`, which seems not enough for `OpenJ9` failed thread.
iirc there was a hang in a test,  using `threadTerminated()` invokes `notifyAll()` when `nthreads == 0` which solved the issue. There is another similar `OpenJ9` change [1] as well.

[1] https://github.com/eclipse-openj9/openj9/blob/a8242cf13af02a7cdb11218c805163fa9f4140d8/jcl/src/java.base/share/classes/java/lang/J9VMInternals.java#L323

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: Will updated the comment.

src/java.base/share/classes/java/lang/Thread.java
keithc-ca: This is different: we're talking about a thread that failed to start.

src/java.base/share/classes/java/lang/Thread.java
keithc-ca: The list of fixes that would be deferred is getting a little too large for my taste. I take back my agreement to defer the known problems discussed here.

JasonFengJ9: Just going back a bit about this PR, it is to adopt RI `j.l.Thread` and support on-going development of project Loom, the change requires `--enable-openjdk-thread-support`, and not enabled by default.
This PR intended to keep existing OpenJ9 behaviours while consuming additional Loom related APIs w/o re-implementation in OpenJ9.
If there is a concern about existing OpenJ9 behaviours, I would like it is proposed in OpenJ9 head stream, and go through usual tests before porting here which doesn't have similar test coverage yet.
Does this make sense?

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: The OpenJ9 change was introduced at same time as this PR to solve the test failure.
Currently I am not able to reproduce the test hang w/o it, there are lots of known segmentation error in JDK19 builds such as https://github.com/eclipse-openj9/openj9/issues/14228.

JasonFengJ9:
I think we can start porting some refactoring changes in this PR back to `OpenJ9` head, which should touch the code in question, and further discussion can happen in OpenJ9. If the concern is addressed there, the change can be ported back.
This PR could be merged as-is, it has no impact on default build, and can speed up project Loom development.

@keithc-ca thoughts?

JasonFengJ9:
Created https://github.com/eclipse-openj9/openj9/pull/14355 to propose the refactoring at `OpenJ9`, and start discussion about any further concerns.

keithc-ca:
I'd prefer to fix known problems before merging.

JasonFengJ9:
> I'd prefer to fix known problems before merging.

Please raise it at https://github.com/eclipse-openj9/openj9/pull/14355, change made there can have usual test coverages.

src/java.base/share/classes/java/lang/Thread.java
keithc-ca: I don't know which failure modes you're referring to. None of what I see in eclipse-openj9/openj9#14228 seems related to `Thread.start()`.

JasonFengJ9: This is a different one, not existing OpenJ9 behaviours.
Initial RI code uses `threadStartFailed()` which caused a test hang, `threadTerminated` fixed the failure (iirc one thread was waiting for this failed thread, `notifyAll()` awoke it, this probably was due to OpenJ9 specific thread lifetime). However I can't reproduce the initial hang w/o this change as per https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/387#discussion_r788809418.

Do you see any problem w/ usage of `threadTerminated()`?

keithc-ca: `threadStartFailed()` is the right method to call if `start()` fails, but a constructed, but unstarted thread must also call `ThreadGroup.addUnstarted()` to balance the bookkeeping.

JasonFengJ9: Use `threadStartFailed()` and verified w/ personal builds.

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: Ported https://github.com/eclipse-openj9/openj9/pull/14355 here, moved `started` checking & `group.add(this)` out of the `try` block, also removed the check of `started` at the `finally` to match RI behaviours.

keithc-ca (changes requested):

closed/autoconf/custom-hook.m4
```diff
@@ -49,10 +49,50 @@ AC_DEFUN_ONCE([CUSTOM_EARLY_HOOK],
...
   OPENJ9_CONFIGURE_INLINE_TYPES
+  OPENJ9_CONFIGURE_OPENJDK_THREAD_SUPPORT
   OPENJ9_THIRD_PARTY_REQUIREMENTS
   OPENJ9_CHECK_NASM_VERSION
 ])
 
+AC_DEFUN([OPENJ9_BASIC_SETUP_FUNDAMENTAL_TOOLS],
+[
+  UTIL_REQUIRE_PROGS(M4, m4)
+])
```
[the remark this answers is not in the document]
JasonFengJ9: Defer the reordering later.

keithc-ca:
@JasonFengJ9 Please resolve the merge conflict.

JasonFengJ9:
Jenkins copyright check

JasonFengJ9:
Resolved the merge conflict.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -988,17 +995,19 @@ public void interrupt() {
                 Interruptible b = blocker;
                 if (b != null) {
                     interrupted = true;
-                    interrupt0();  // inform VM of interrupt
+                    interruptImpl();  // inform VM of interrupt
```
keithc-ca, line 998: nit: extra space after `;`

JasonFengJ9: Removed.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1127,19 +1145,30 @@ public final void resume() {
...
-                newPriority = g.getMaxPriority();
+        if ((MIN_PRIORITY <= newPriority) && (newPriority <= MAX_PRIORITY)) {
+            ThreadGroup myThreadGroup = getThreadGroup();
+            /* [PR 134167 Ignore dead threads. */
+            if (myThreadGroup != null) {
+                int finalPriority = newPriority;
+                int threadGroupMaxPriority = myThreadGroup.getMaxPriority();
+                if (threadGroupMaxPriority < newPriority) {
+                    finalPriority = threadGroupMaxPriority;
+                }
```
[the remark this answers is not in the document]
JasonFengJ9: Updated.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -2020,4 +2052,266 @@ private void dispatchUncaughtException(Throwable e) {
...
+    private Thread(ThreadGroup group, Runnable runnable, String threadName, AccessControlContext acc, boolean inheritThreadLocals) {
+        super();
+        if (threadName == null) {
+            throw new NullPointerException();
+        }
+        // We avoid the public API 'setName', since it does redundant work (checkAccess)
+        this.name = threadName;
+        // No API available here, so just direct access to inst. var.
+        this.target = runnable;
+        Thread currentThread  = currentThread();
```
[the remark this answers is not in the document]
JasonFengJ9: Removed.

JasonFengJ9:
Jenkins copyright check

keithc-ca (approved):

keithc-ca:
jenkins compile alinux64 jdknext

tajila (approved):

PR #390, opened 2022-01-13
Correctly locate directory containing JVM shared library in J9 SDKs

babsingh:
Fixes: https://github.com/eclipse-openj9/openj9/issues/14079


babsingh:
I have assumed that J9 only supports the `server` variant i.e. `libjvm.so or jvm.dll` is found in either `jdk/bin/server/jvm.dll` (Windows) or `jdk/lib/server/libjvm.so` (elsewhere).

fyi @pshipton @tajila

pshipton (approved):

pshipton:
I assume this will be backported to jdk18 and the test exclude removed?

test/lib/jdk/test/lib/Platform.java
```diff
@@ -358,7 +364,10 @@ public static Path jvmLibDir() {
     }
 
     private static String variant() {
-        if (Platform.isServer()) {
+        if (Platform.isServer()
```
keithc-ca, line 367: Might it be a better solution to modify `isServer()`? I think tests will expect at least one of the `isXXX()` methods involving `vmName` to return true.

babsingh: Yes, we will be able to support more tests. Let me draft a new PR.

pshipton: That will break isDefaultCDSArchiveSupported() so take that into account.

babsingh: Created https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/391.

PR #391, opened 2022-01-13
Update jdk.test.lib.Platform to work with J9

babsingh:
`isServer()` updated to return `true` if `java.vm.name` contains `OpenJ9` or `IBM`.

`isDefaultCDSArchiveSupported()` updated to return `false` if `java.vm.name` contains
`OpenJ9` or `IBM` since J9 does not support the default CDS archive.

This is an updated fix for eclipse-openj9/openj9#14079.

Related: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/390

### DIFF
--- a/test/lib/jdk/test/lib/Platform.java
+++ b/test/lib/jdk/test/lib/Platform.java
@@ -60,12 +60,16 @@ public class Platform {
                 PrivilegedAction<String>) () -> System.getProperty(key));
     }
 
+    private static boolean isJ9() {
+        return vmName.contains("OpenJ9") || vmName.contains("IBM");
+    }
+
     public static boolean isClient() {
         return vmName.endsWith(" Client VM");
     }
 
     public static boolean isServer() {
-        return vmName.endsWith(" Server VM");
+        return isJ9() || vmName.endsWith(" Server VM");
     }
 
     public static boolean isZero() {
@@ -364,10 +368,7 @@ public class Platform {
     }
 
     private static String variant() {
-        if (Platform.isServer()
-        || vmName.toLowerCase().contains("openj9")
-        || vmName.toLowerCase().contains("ibm")
-        ) {
+        if (Platform.isServer()) {
             return "server";
         } else if (Platform.isClient()) {
             return "client";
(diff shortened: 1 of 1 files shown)


test/lib/jdk/test/lib/Platform.java
```diff
@@ -382,14 +384,13 @@ private static String variant() {
...
-                !isZero()    &&
-                !isMinimal() &&
-                !isARM());
+        return is64bit()
+               && isServer()
+               && (isLinux() || isOSX() || isWindows())
+               && !isZero()
+               && !isMinimal()
+               && !isARM()
+               && !isJ9();
```
keithc-ca, line 393: I'd like to see this change reduced. I suggest just one line needs to change:
```
    isServer() &&
```
to
```
    isServer() && !isJ9() &&
```

test/lib/jdk/test/lib/Platform.java
```diff
@@ -60,12 +60,17 @@ private static String privilegedGetProperty(String key) {
                 PrivilegedAction<String>) () -> System.getProperty(key));
     }
 
+    private static boolean isJ9() {
+        return vmName.toLowerCase().contains("openj9")
+               || vmName.toLowerCase().contains("ibm");
```
keithc-ca, line 65: The uses of `toLowerCase()` are not necessary; just look for 'OpenJ9' or 'IBM'.

keithc-ca (changes requested):

test/lib/jdk/test/lib/Platform.java
babsingh: Updated.

test/lib/jdk/test/lib/Platform.java
babsingh: Updated.

keithc-ca (approved):

pshipton:
jenkins test sanity.openjdk jdknext

pshipton:
jenkins test sanity.openjdk alinux64 jdknext

pshipton:
There are too many failures. We don't have an exclude list for jdknext so all the tests that should be excluded are not. Pls port this change to jdk18 and I'll test it there.

babsingh:
> Pls port this change to jdk18 and I'll test it there.

Created https://github.com/ibmruntimes/openj9-openjdk-jdk18/pull/5.

PR #404, opened 2022-02-14
Fix test to check for finalizer thread after object finalized

pshipton:
In test java/lang/Object/FinalizationOption.java

Port of https://github.com/ibmruntimes/openj9-openjdk-jdk18/pull/18 to the head stream.

Issue https://github.com/eclipse-openj9/openj9/issues/14491


JasonFengJ9 (approved):
LGTM

PR #417, opened 2022-03-30
Fix @Stable annotation in MethodHandleImpl

knn-k:
This commit removes a `@Stable` annotation in MethodHandleImpl.
The field `asCollectorCache` can be overwritten after it is first
assigned.

Signed-off-by: KONNO Kazuhiro <konno@jp.ibm.com>


knn-k:
https://github.com/ibmruntimes/openj9-openjdk-jdk18/pull/34 is the PR for JDK 18 for fixing the same issue.

pshipton (approved):

PR #444, opened 2022-05-20

JasonFengJ9:
Resolved conflicts:
```
	src/java.base/unix/classes/java/lang/ProcessEnvironment.java
```


src/java.base/unix/classes/java/lang/ProcessEnvironment.java
```diff
@@ -62,10 +62,13 @@
...
 import java.io.*;
+import java.nio.charset.Charset;
 import java.util.*;
 /*[IF CRIU_SUPPORT]*/
 import openj9.internal.criu.InternalCRIUSupport;
 /*[ENDIF] CRIU_SUPPORT*/
+import jdk.internal.util.StaticProperty;
+
 
 final class ProcessEnvironment
```
keithc-ca, line 73: Can you redo this so the CRIU additions are not in the middle of the imports from upstream and without the extra blank line:
```
import jdk.internal.util.StaticProperty;
/*[IF CRIU_SUPPORT]*/
import openj9.internal.criu.InternalCRIUSupport;
/*[ENDIF] CRIU_SUPPORT*/

final class ProcessEnvironment
```

keithc-ca (changes requested):

src/java.base/unix/classes/java/lang/ProcessEnvironment.java
JasonFengJ9: Moved CRIU patch after `import jdk.internal.util.StaticProperty;`, note there are two blank lines in the upsteam code.

keithc-ca (approved):

PR #446, opened 2022-05-26
Fix SunEntries dependency

fengxue-IS:
add import sun.security.action.GetPropertyAction

### DIFF
--- a/src/java.base/share/classes/sun/security/provider/SunEntries.java
+++ b/src/java.base/share/classes/sun/security/provider/SunEntries.java
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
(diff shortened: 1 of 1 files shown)


fengxue-IS:
OpenJDK removed that import and replaced it with `GetBooleanAction` we still have patched code that uses the `GetPropertyAction` class at 
https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9-staging/src/java.base/share/classes/sun/security/provider/SunEntries.java#L429-L431

fengxue-IS:
see https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/08e03181aaeff8572f54414789c0171039710c53 for the change

pshipton (approved):

PR #448, opened 2022-05-26
Remove configure option '--enable-openjdk-thread-support'

keithc-ca:
Using openjdk threads is now required.

Note: This targets the openj9-staging branch.


pshipton (approved):

pshipton:
@fengxue-IS @JasonFengJ9 fyi

PR #451, opened 2022-06-02
Fix Thread.stop if-condition typo

fengxue-IS:
Signed-off-by: Jack Lu <Jack.S.Lu@ibm.com>

PR #452, opened 2022-06-03

JasonFengJ9:
Resolved conflicts:
```
```


keithc-ca:
Commit e918998c6d64d7a8153593e543a1d2f161ab3d1d does not have a "Signed-off-by" line.
We don't need the two merge commits (the other being 430382405aacfc24f082ee4371c150a8c03142fa).
Can you redo this to fix those things and perhaps start with the latest openj9-staging branch.

JasonFengJ9:
> Commit https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/e918998c6d64d7a8153593e543a1d2f161ab3d1d does not have a "Signed-off-by" line.

No `Signed-off-by` for `j9build` automatic merge commits as well such as https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/26d001ca158c3c6b655d6c3ae81935aa9ba8d179
This is due to no `--signoff` for `git merge` at `mergescripts/merge_master_to_staging.sh`, created a PR `runtimes/openjdk-build/pull/35`

> We don't need the two merge commits (the other being https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/430382405aacfc24f082ee4371c150a8c03142fa).

`Merge current openj9 into openj9-staging` is a separated merge command before `Merge master head into openj9-staging`, there will be two commits as well for automatic runs w/o merge conflicts requiring manual intervene.

keithc-ca:
Ok, I missed that the two merge commits were merging openj9 and master, respectively. That doesn't mean that both should not have 'Signed-off-by' annotations. If a script needs updating that's fine, but there's no reason we can't get this right.

JasonFengJ9:
> That doesn't mean that both should not have 'Signed-off-by' annotations. 

Applied updated `merge_master_to_staging.sh` though itself not merged yet, the `Signed-off-by` are already there, pls have another look.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -3094,7 +3112,7 @@ private void setPriority0(int newPriority) {
     private void stop0(Object o) {
         synchronized (interruptLock) {
             if (o != null) {
-                if (!started) {
+                if (started) {
```
keithc-ca, line 3115: Please fix this: it reverts #451.

JasonFengJ9: Good catch, fixed.

keithc-ca (approved):

Issue #422, opened 2022-04-12
Enabling CMake by default for cross-compiles


dnakamura:
The CMake code should now be able to handle cross compile builds. This issue is to track any problems / objections anybody has before making cross-compile builds use cmake by default.  CMake builds can currently be tested by passing `--with-cmake` when calling configure. In addition, where CMake is used by default, uma can still be used by passing `--without-cmake`.

janvrany:
This is process-level QEMU so it has all the memory host has. I can try to increase host memory but I'd like to know why my home-baked CMake cross-build does not suffer from this problem. 

We had similar problem with OMR tests and in the end we just increased the memory on Eclipse CI. Interestingly, on my system tests run just fine...

dnakamura:
Its odd that it dies during tracegen. Memory requirements there should be fairly minimal.  @janvrany Are the sources for your hacked cmake support available ?

janvrany:
Sorry, finally back home. 
Sure, my changes are here:

* https://github.com/janvrany/openj9/commit/d0841fa0258ce5f5e7ee0f5923f15636f2965038
* https://github.com/janvrany/openj9/commit/8c5866ee2d797649a8043610690d75c6276fdeaa
* https://github.com/janvrany/openj9-openjdk-jdk11/commit/bf01ec96ee9376d0d021daf59eae138a30d839e2
* https://github.com/janvrany/openj9-openjdk-jdk11/commit/1a2d209446c9feab200f2b38837112f039a3c7db

Note that these are quick and dirty hacks I did just to allow me to build JIT component for RISC-V JIT development - my hacks may not run tracegen at all...

As for the memory requirements, I think it is QEMU that leaks the memory, not the tracegen itself. I'll investigate.

janvrany:
Ah, this is embarrassing! It turned out that CMake picked up an ancient QEMU that was installed on my build node which suffered from some memory leak. With QEMU 6.2 or 7.0 it works just fine. 

I'm very sorry about the noise!

pshipton:
https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/454 enables cmake for cross compiling, for jdk19+.

knn-k:
I sometimes use cross compilation in my environment for a specific AArch64 Linux target.
I am OK about making CMake by default.  I guess `--with-cmake=no` still works in case of unexpected failures with cross builds.

keithc-ca:
[...]

PR #454, opened 2022-06-07
Default to using cmake on cross-compiles

dnakamura:
Issue https://github.com/ibmruntimes/openj9-openjdk-jdk/issues/422

Signed-off-by: Devin Nakamura <devinn@ca.ibm.com>


pshipton (approved):

pshipton:
Will this be added to older versions as well?

PR #456, opened 2022-06-07
Increase clone timeout to 30 minutes

keithc-ca:
Copyright checking often fails because 10 minutes isn't always enough.


keithc-ca:
Resolved conflict.

AdamBrousseau (approved):

PR #489, opened 2022-09-12
Add initCompressionFlag() to String constructors

mikezhang1234567890:
Set compression flag in constructors creating uncompressed strings when compact strings is enabled. OpenJ9 uses this flag to check if a string with uncompressed characters has been created, and not having it enabled when there has been leads to errors since the code assumes all strings are still compressed.

Port of https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/143

fyi @keithc-ca

### DIFF
--- a/src/java.base/share/classes/java/lang/String.java
+++ b/src/java.base/share/classes/java/lang/String.java
@@ -527,6 +527,7 @@ public final class String
                 this.value = val;
                 return;
             }
+            initCompressionFlag();
         }
         this.coder = UTF16;
         this.value = StringUTF16.toBytes(codePoints, offset, count);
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/lang/String.java
```diff
@@ -5301,6 +5310,7 @@ void getBytes(byte[] dst, int srcPos, int dstBegin, byte coder, int length) {
                     this.value = buf;
                     return;
                 }
+                initCompressionFlag();
```
keithc-ca, line 5313: This call should not be conditional on `asb.maybeLatin1`.

keithc-ca (changes requested):

src/java.base/share/classes/java/lang/String.java
mikezhang1234567890: Changed this so it's in its own `if(COMPACT_STRINGS)` check.

keithc-ca (approved):

keithc-ca:
jenkins compile osx jdknext

PR #514, opened 2022-10-21
[Jtreg]Fix the issue with addVarg/nextVarg of VaList on Power

ChengJin01:
The change is to resolve the Jtreg test related issue with addVarg/nextVarg of VaList on Power to ensure the struct value is correctly read from va_list and added to va_list when the struct argument size is inconsistent with its layout size.


ChengJin01:
The code is the same as https://github.com/ibmruntimes/openj9-openjdk-jdk19/pull/42.

Reviewer: @tajila
FYI: @pshipton

tajila (approved):

tajila:
jenkins test sanity plinux jdknext

src/java.base/share/classes/jdk/internal/foreign/abi/ppc64/sysv/SysVPPC64leVaList.java
```diff
@@ -155,6 +160,11 @@ private void checkNextArgument(MemoryLayout argLayout, long argByteSize) {
 		}
 	}
 
+	private static long getSmallerStructArgSize(MemorySegment structSegment, MemoryLayout structArgLayout) {
+		return (structSegment.byteSize() > structArgLayout.byteSize()) ?
+							structArgLayout.byteSize() : structSegment.byteSize();
```
keithc-ca, line 165: Am I missing something or could this just be this?
```
    return Math.min(structSegment.byteSize(), structArgLayout.byteSize());
```

ChengJin01: We expect to get the smaller size suitable for the struct, in such case `Math.min()` should work. I will create a PR for this minor adjustment.

ChengJin01: PRs are created at https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/521 and https://github.com/ibmruntimes/openj9-openjdk-jdk19/pull/49.

PR #474, opened 2022-07-19
Several JEP424 fixes

keithc-ca:
Cherry-pick https://github.com/ibmruntimes/openj9-openjdk-jdk19/pull/15.

PR #518, opened 2022-10-27
Fix the alignment issue with long & pointer on AIX

ChengJin01:
The changes aim to restore the original alignment setting for long and pointer on AIX given the padding is required in native in the case of struct [int/float,long] and [int/float,pointer].


ChengJin01:
The changes revert part of fix specific to long and pointer on AIX at https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/474, which was captured in jtreg test for upcall (also verified in native code) given the alignment for long and pointer are 64bit rather than 32bit on AIX.

Reviewer: @tajila
FYI: @pshipton, @keithc-ca

tajila:
jenkins compile aix jdknext

PR #520, opened 2022-10-31
Fix Thread.getStackTrace synchronization

fengxue-IS:
Add extra check in getStackTrace0 under interruptLock to ensure only live thread is passed to native code.

Port of: https://github.com/ibmruntimes/openj9-openjdk-jdk19/pull/47


tajila:
jenkins compile amac jdknext

PR #523, opened 2022-11-07
Add javax.management.Attribute to serialfilter

pshipton:
OpenJ9 uses generic CompositeData, which uses Attribute to transmit data, rather than having custom CompositeData implementations.

Related to https://bugs.openjdk.org/browse/JDK-8283093 - JMX connections use an ObjectInputFilter by default

Closes https://github.com/eclipse-openj9/openj9/issues/16253

Tested the equivalent change via grinders
https://openj9-jenkins.osuosl.org/job/Grinder_testList_0/179/
https://openj9-jenkins.osuosl.org/job/Grinder_testList_0/180/

### DIFF
--- a/src/jdk.management.agent/share/conf/management.properties
+++ b/src/jdk.management.agent/share/conf/management.properties
@@ -302,5 +302,5 @@
 #   If the pattern ends with "*", it matches any class with the pattern as a prefix.
 #   If the pattern is equal to the class name, it matches.
 #   Otherwise, the status is UNDECIDED.
-com.sun.management.jmxremote.serial.filter.pattern=java.lang.*;java.math.BigInteger;java.math.BigDecimal;java.util.*;javax.management.openmbean.*;javax.management.ObjectName;java.rmi.MarshalledObject;javax.security.auth.Subject;!*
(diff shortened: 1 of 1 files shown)


keithc-ca:
This exception is just a bug in the test code.
```
java.lang.NullPointerException: Cannot read the array length because "<local8>" is null
	at org.openj9.test.java.lang.management.TestOpenJ9DiagnosticsMXBean.tearDown(TestOpenJ9DiagnosticsMXBean.java:196)
```
`File.listFiles()` will return `null` instead of an empty array for a non-existent directory. Please explain why that should not be fixed instead of the change proposed here.

keithc-ca:
This should not close https://github.com/eclipse-openj9/openj9/issues/16253, the NPE should still be avoided.

keithc-ca:
I think the test code should be fixed. We would still get
```
java.lang.reflect.UndeclaredThrowableException
	...
Caused by: java.io.InvalidClassException: filter status: REJECTED
```
if the problem addressed here returns. This should not close that issue.

keithc-ca (approved):

pshipton:
If you insist. Created https://github.com/eclipse-openj9/openj9/pull/16295

keithc-ca:
Ignoring the copyright check failure: The modified file has no copyright notice.

With https://github.com/eclipse-openj9/openj9/pull/16295 merged, I'll allow this to close https://github.com/eclipse-openj9/openj9/issues/16253.

PR #536, opened 2023-01-05
Fix problem building openssl when ccache is enabled

keithc-ca:
If `ccache` is enabled and the environment contains either `CC` or `CXX`, their values (as defined in `openssl.gmk`) are propagated to the instance of `make` which will build `openssl` causing it to fail. Instead, pass along `CC` and `CXX` without the `ccache` wrapper.

Also removed the redundant test (`openssl.gmk` is only used if `$(BUILD_OPENSSL)` is `yes`).

### DIFF
--- a/closed/openssl.gmk
+++ b/closed/openssl.gmk
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2018, 2022 All Rights Reserved
+# (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
 # ===========================================================================
 # This code is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 only, as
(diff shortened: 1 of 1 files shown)


closed/openssl.gmk
```diff
@@ -88,12 +88,18 @@ ifeq (,$(OPENSSL_TARGET))
...
+ifneq (,$(CCACHE))
+  # If ccache is enabled and the environment contains either CC or CXX, their
+  # values (as defined in this make) are propagated to the instance of make
+  # which will build openssl causing it to fail. Instead, pass along CC and
+  # CXX without the ccache wrapper.
+  OPENSSL_MAKE += CC=$(ac_cv_prog_CC) CXX=$(ac_cv_prog_CXX)
+endif # CCACHE
+
 build_openssl :
-ifeq ($(BUILD_OPENSSL), yes)
```
pshipton, line 92: What's this change about?

pshipton:
jenkins compile all jdknext

PR #544, opened 2023-01-20
Java Restricted Security Mode

taoliult:
Signed-off-by: Tao Liu [tao.liu@ibm.com](mailto:tao.liu@ibm.com)

This PR is for adding the codes of Java Restricted Security Mode.

The Java Restricted Security Mode is used to restrict the providers and algorithms, by the properties configured in the java.security file.

The properties will be of the following format:

```
RestrictedSecurity(n).desc.name = This string would be presented to the user when the audit setting is turned on, used to identify the policy in a helpful manner
RestrictedSecurity(n).desc.number = Security policy number
RestrictedSecurity(n).desc.policy = Policy details link
RestrictedSecurity(n).desc.sunsetDate = Date by which the policy will sunset

RestrictedSecurity(n).tls.disabledNamedCurves = Map to jdk.disabledNamedCurves
RestrictedSecurity(n).tls.disabledAlgorithms = Map to jdk.tls.disabledAlgorithms
RestrictedSecurity(n).tls.ephemeralDHKeySize = Map to jdk.tls.ephemeralDHKeySize
RestrictedSecurity(n).tls.legacyAlgorithms = Map to jdk.tls.legacyAlgorithms

RestrictedSecurity(n).jce.certpath.disabledAlgorithms = Map to jdk.certpath.disabledAlgorithms
RestrictedSecurity(n).jce.legacyAlgorithms = Map to jdk.security.legacyAlgorithm
RestrictedSecurity(n).jce.provider.1 = Map to security.provider
RestrictedSecurity(n).jce.provider.2 = Map to security.provider
RestrictedSecurity(n).jce.provider.3 = Map to security.provider
RestrictedSecurity(n).jce.provider.4 = Map to security.provider

RestrictedSecurity(n).keystore.type = Map to keystore.type 
RestrictedSecurity(n).javax.net.ssl.keyStore = System property javax.net.ssl.keyStore

RestrictedSecurity(n).securerandom.provider = Secure random provider
RestrictedSecurity(n).securerandom.algorithm = Secure random algorithm
```

The new runtime command line arguments:

```
-Dsemeru.restrictedsecurity = (n)
    This flag will select the settings for the user specified fips policy.
-Dsemeru.restrictedsecurity= audit 
    This will list the name and number of all configured fips policies. it will NOT cause the jvm to terminate after printing the fips policies.
-Dsemeru.restrictedsecurity = trace
    This will list all properties relevant to the fips mode, including the existing default properties and the FIPS restrictions.
-Dsemeru.restrictedsecurity = help
    This will print help message.
```

The options can be combined: e.g. -Dsemeru.restrictedsecurity=1,trace,audit

During the process of reading the java.security properties file, the JDK will detect if Restricted Security mode as selected and the selected policy. This detection will be done as a check in the JCA code to see if the flag "semeru.restrictedsecurity" was used, if it is used, the specific policy number will be extracted, the code will then check the properties file to see if such a policy number exist. If a valid policy is selected, two configuration steps will occur:

1. JSSE Configuration during start up. If a valid policy was selected, the properties in java.security file will be read and mapped during the start up.

2. Cryptography available. The existing providers will be removed and only the compliant providers (defined by “SecurityRestrict(n).jce.provider”) will be added. And also, a list of permitted services that each provider may provide. A provider without any additional constraints will have all its services enabled. A provider with the constraints will only register the allowed constraints. For example:

```
RestrictedSecurity1.jce.provider.3 = SunEC [{KeyFactory, EC, ImplementedIn=Software: \
```


taoliult:
@keithc-ca Java Restricted Security Mode PR for JDKnext, please help to review and advice.

taoliult:
@keithc-ca

The pushes in last two days are for git rebase the branch, because the head had new commits, so I rebase the branch to accept the new commits from the head to make sure there is no any conflict.

Right now, the PR is ready for review. Please help to review and advice.

keithc-ca:
> The pushes in last two days are for git rebase

Please stop doing that. I'll let you know if it needs to be rebased or otherwise updated.

taoliult:
@keithc-ca Ok, sure. Right now the codes are ready for review.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,340 @@
...
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
```
keithc-ca, line 35: Please remove this import and use `Map.Entry` instead.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,769 @@
...
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Properties;
+
+import sun.security.util.Debug;
+
+public final class RestrictedSecurityProperties {
```
keithc-ca, line 39: Neither this class nor any of it's features should be `public`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,769 @@
...
+                }
+                if (userSecurityAudit) {
+                    listAudit();
+                }
+                if ((userSecurityNum == 0)) {
+                    if (userSecurityTrace) {
+                        new RuntimeException(
+                                "Unable to list the trace info without specify the security policy number.")
+                                .printStackTrace();
+                        System.exit(1);
```
keithc-ca, line 164: This pattern appears often enough that I suggest we should have these methods:
```java
    private static void printStackTraceAndExit(Exception exception) {
        exception.printStackTrace();
        System.exit(1);
    }

    private static void printStackTraceAndExit(String message) {
        printStackTraceAndExit(new RuntimeException(message));
    }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,769 @@
...
+        return descName;
+    }
+
+    public String getDescriptionNumber() {
+        return descNumber;
+    }
+
+    public String getDescriptionPolicy() {
+        return descPolicy;
+    }
```
keithc-ca, line 720: Please remove these unused methods

src/java.base/share/conf/security/java.security
```diff
@@ -85,6 +85,58 @@ security.provider.tbd=Apple
...
+     TLS_AES_128_GCM_SHA256, TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384, \
+     TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, \
+     TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, TLS_ECDH_ECDSA_WITH_AES_256_GCM_SHA384, \
+     TLS_ECDH_RSA_WITH_AES_256_GCM_SHA384, TLS_ECDH_ECDSA_WITH_AES_128_GCM_SHA256, \
+     TLS_ECDH_RSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256, \
+     TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256, TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256, \
+     TLS_EMPTY_RENEGOTIATION_INFO_SCSV
+ RestrictedSecurity1.tls.ephemeralDHKeySize =
+ RestrictedSecurity1.tls.legacyAlgorithms =
+
```
keithc-ca, line 118: Please remove the whitespace preceding property names.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,340 @@
...
+import java.util.Properties;
+
+import sun.security.util.Debug;
+
+/**
+ * Configures the security providers when in restricted security mode.
+ */
+public final class RestrictedSecurityConfigurator {
```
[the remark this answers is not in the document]
taoliult: @keithc-ca 
Change the class name from “RestrictedSecurityConfigurator” to “RestrictedSecurity”. And also update class “RestrictedSecurityProperties” as a nested class of “RestrictedSecurity”.

taoliult:
@keithc-ca

Changed the class name from “RestrictedSecurityConfigurator” to “RestrictedSecurity”. And also update class “RestrictedSecurityProperties” as a nested class of “RestrictedSecurity”. 

The PR is ready for review. Please help to review and advice.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -1383,6 +1391,11 @@ protected void putService(Service s) {
             throw new IllegalArgumentException
                     ("service.getProvider() must match this Provider object");
         }
+        if (!RestrictedSecurity.isServiceAllowed(s)) {
+            // We're in restricted security mode which does not allow this service,
+            // then return without registering.
```
keithc-ca, line 1396: The word "then" doesn't belong here.

src/java.base/share/classes/java/util/ServiceLoader.java
```diff
@@ -880,6 +888,12 @@ private Provider<S> loadProvider(ServiceProvider provider) {
             fail(service, clazz + " is not public");
         }
 
+        if (!RestrictedSecurity.isProviderAllowed(clazz)) {
+            // We're in restricted security mode which does not allow this provider,
+            // then skip it.
```
keithc-ca, line 893: The word "then" doesn't belong here.

src/java.base/share/classes/java/util/ServiceLoader.java
```diff
@@ -1235,6 +1249,11 @@ private boolean hasNextService() {
                     }
 
                     if (service.isAssignableFrom(clazz)) {
+                        if (!RestrictedSecurity.isProviderAllowed(clazz)) {
+                            // We're in restricted security mode which does not allow this provider,
+                            // then skip it.
```
keithc-ca, line 1254: The word "then" doesn't belong here.

src/java.base/share/classes/sun/security/jca/ProviderConfig.java
```diff
@@ -167,6 +169,11 @@ public String toString() {
      */
     @SuppressWarnings("deprecation")
     Provider getProvider() {
+        if (!RestrictedSecurity.isProviderAllowed(provName)) {
+            // We're in restricted security mode which does not allow this provider,
+            // then return without loading.
```
keithc-ca, line 174: The word "then" doesn't belong here.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -120,7 +120,7 @@ public final class SunPKCS11 extends AuthProvider {
 
     // This is the SunPKCS11 provider instance
     // there can only be a single PKCS11 provider in
-    // FIPS mode.
+    // restricted security FIPS mode.
```
keithc-ca, line 123: I think "FIPS" shoud be omitted here.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1027 @@
...
+            System.out.println();
+        }
+
+        /**
+         * Check if the input string is null. If null return "".
+         *
+         * @param string the input string
+         * @return "" if the string is null
+         */
+        private String parseProperty(String string) {
```
keithc-ca, line 970: This and the next two helper methods can be `static`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1027 @@
...
+         * @return true if the input string is asterisk
+         */
+        private boolean isAsterisk(String string) {
+            return "*".equals(string);
+        }
+
+        /**
+         * Nested class for provider's constraints
+         */
+        private final class Constraint {
```
keithc-ca, line 1014: This class should be `static`.

keithc-ca (changes requested):

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
taoliult: @keithc-ca Because the code changes here only apply for the FIPS mode, and the FIPS mode is belong to the restricted security mode, which policy number is 1. So I changed this "FIPS mode" to "restricted security FIPS mode".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1027 @@
...
+        return securityEnabled;
+    }
+
+    /**
+     * Restricted security mode secure random provider
+     *
+     * @return the secure random provider
+     */
+    public static String getRandomProvider() {
+        return restricts.jdkSecureRandomProvider;
```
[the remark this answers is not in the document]
taoliult: @keithc-ca 
I added a new method "restrictsCheck()", and in that method, it will check the "restricts". And I also moved the SunsetDate check and SecureRandom check into this "restrictsCheck()" method. To make sure, once the "securityEnabled" is true, then the "restricts" and SecureRandom won't be null.

taoliult:
@keithc-ca The PR is ready for review. Please help to review and advice.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
keithc-ca: Unless I misunderstand, FIPS is a special kind of restricted security mode, so saying both is just redundant. If the comment only applies to FIPS mode, then don't change it; if it applies more broadly to any restricted security mode, then mentioning FIPS is misleading. Comments in several other places should be updated accordingly.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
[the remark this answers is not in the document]
keithc-ca: There's also the problem that `configure(Properties)` can be called more than once without error.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+        shouldEnableSecurity = (userEnabledFIPS || userEnabledSecurity) && isSecuritySupported;
+    }
+
+    private RestrictedSecurity() {
+        super();
+    }
+
+    /**
+     * Restricted security mode will be enabled when the semeru.restrictedsecurity
+     * system property is set.
```
keithc-ca, line 95: The first sentence should summarize what the method does; then you can elaborate like this, but it should also mention the property `semeru.fips`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+     *
+     * @return true if restricted security is enabled
+     */
+    public static boolean isEnabled() {
+        return securityEnabled;
+    }
+
+    /**
+     * Restricted security mode secure random provider
```
keithc-ca, line 104: Each sentence should end with a period. The summary should explain with the method does, i.e. "Get the restricted security mode secure random provider." It should point out that this method can only be called in restricted security mode.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+    public static String getRandomProvider() {
+        if (!securityEnabled) {
+            printStackTraceAndExit(
+                    "Restricted security mode secure random provider can only be used when restricted security mode is enabled.");
+        }
+        return restricts.jdkSecureRandomProvider;
+    }
+
+    /**
+     * Restricted security mode secure random algorithm
```
keithc-ca, line 117: See comments for `getRandomProvider()`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+
+    /**
+     * Check if the service is allowed in restricted security mode.
+     *
+     * @param service the service to check
+     * @return true if the service is allowed
+     */
+    public static boolean isServiceAllowed(Service service) {
+        if (securityEnabled) {
+            return restricts.isRestrictedServiceAllowed(service);
```
keithc-ca, line 147: I suggest the service class should be checked before delegating to `RestrictedSecurityProperties`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+        }
+
+        try {
+            if (shouldEnableSecurity) {
+                if (debug != null) {
+                    debug.println("Restricted security mode detected, loading...");
+                }
+
+                // Read and set user restricted security settings.
+                initUserSetting();
```
keithc-ca, line 198: This should check that `restricts == null` (i.e. that `configure()` has not already been called.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+
+                // Initialize the restricted security properties from java.security file.
+                restricts = new RestrictedSecurityProperties(userSecurityNum,
+                        props, userSecurityTrace, userSecurityAudit, userSecurityHelp);
+
+                // Restricted security properties checks
```
keithc-ca, line 204: Missing period.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+        if (isNullOrBlank(restricts.jdkSecureRandomProvider)
+                || isNullOrBlank(restricts.jdkSecureRandomAlgorithm)) {
+            printStackTraceAndExit("Restricted security mode secure random is null.");
+        }
+    }
+
+    /**
+     * Check if restricted security policy sunset.
+     *
+     * @param descSunsetDate the sun set date from java.security
```
keithc-ca, line 385: The summary could be improved (missing verb).
Also 'sunset' is one word, not two.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+        exception.printStackTrace();
+        System.exit(1);
+    }
+
+    private static void printStackTraceAndExit(String message) {
+        printStackTraceAndExit(new RuntimeException(message));
+    }
+
+    /**
+     * Nested class for the properties of restricted security
```
keithc-ca, line 425: It isn't helpful to note that this class is nested; missing period.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+        private String jdkTlsDisabledNamedCurves;
+        private String jdkTlsDisabledAlgorithms;
+        private String jdkTlsDphemeralDHKeySize;
+        private String jdkTlsLegacyAlgorithms;
+        private String jdkCertpathDisabledAlgorithms;
+        private String jdkSecurityLegacyAlgorithm;
+        private String keyStoreType;
+        private String keyStore;
+
+        // For Secure Random.
```
keithc-ca, line 444: `SecureRandom` is the class name (no space), unless you are referring to the configuration of source of secure random numbers (no capitals).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+        String jdkSecureRandomProvider;
+        String jdkSecureRandomAlgorithm;
+
+        // Provider with argument (provider name + optional argument).
+        private List<String> providers;
+        // Provider without argument.
+        private List<String> providersSN;
```
keithc-ca, line 451: It might be helpful to expand `SN`; is it short for `SimpleName`? If so, please don't abbreviate.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+            }
+        }
+
+        /**
+         * Load security constraints with type, algorithm, attributes.
+         *
+         * Example:
+         * RestrictedSecurity1.jce.provider.1 = SUN [{CertPathBuilder, PKIX, *},
+         * {Policy,
+         * JavaPolicy, *}, {CertPathValidator, *, *}].
```
keithc-ca, line 636: Please join the description on these two lines.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+
+                        String inType = input[0].trim();
+                        String inAlgorithm = input[1].trim();
+                        String inAttributes = input[2].trim();
+
+                        // Each attribute must includes 2 fields(key and value) or *.
+                        if (!isAsterisk(inAttributes)) {
+                            String[] attributeArray = inAttributes.split(":");
+                            for (String attribute : attributeArray) {
+                                String[] in = attribute.split("=");
```
keithc-ca, line 697: I suggest this use of `split()` (and the one on line 778) should be `split("=", 2)`, otherwise some of the input could be ignored.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+                    }
+                }
+                if (debug != null) {
+                    debug.println("Security constraints check."
+                            + " Service type: " + type
+                            + " Algorithm: " + algorithm
+                            + " is NOT allowed in provider " + providerName);
+                }
+                // Go through all the constraints for the provider,
+                // no match, then return NOT allowed.
```
keithc-ca, line 806: This comment doesn't read well; I suggest:
```java
            // No match for any constraint, return NOT allowed.
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1044 @@
...
+                } else if (ch == '}' || ch == ']' || ch == ')') {
+                    if (!deque.isEmpty()
+                            && ((deque.peekFirst() == '{' && ch == '}')
+                                    || (deque.peekFirst() == '[' && ch == ']')
+                                    || (deque.peekFirst() == '(' && ch == ')'))) {
+                        deque.removeFirst();
+                    } else {
+                        return false;
+                    }
+                }
```
keithc-ca, line 1013: I suggest simplifying this by pushing the expected closing character:
```java
                switch (ch) {
                case '{':
                    deque.addFirst('}');
                    break;
                case '[':
                    deque.addFirst(']');
                    break;
                case '(':
                    deque.addFirst(')');
                    break;
                case '}':
                case ']':
                case ')':
                    if (deque.isEmpty() || (deque.removeFirst().charValue() != ch)) {
                        return false;
                    }
                    break;
                default:
                    break;
                }
```

keithc-ca (changes requested):

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
taoliult: @keithc-ca Ok, I will remove the "restricted security" and only use the "FIPS" for those FIPS specific changes.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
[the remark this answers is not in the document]
taoliult: @keithc-ca 
I added a check in the “configure()” method. If the “securityEnabled” is true when go into this “configure()” method, that means the restricted security mode is already initialized. Then, throw a runtime exception explains the restricted security mode can't be initialized twice.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca 

In the Provider.java, before the “RestrictedSecurity.isServiceAllowed(s)” is called, there are the check for the service class, codes as below. So I did not add the check in the “RestrictedSecurity” class “isServiceAllowed()” method. 

```
        if (s == null) {
            throw new NullPointerException();
        }
        if (s.getProvider() != this) {
            throw new IllegalArgumentException
                    ("service.getProvider() must match this Provider object");
        }
        if (!RestrictedSecurity.isServiceAllowed(s)) {
            // We're in restricted security mode which does not allow this service,
            // return without registering.
            return;
        }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca 
I added a check when go into the “configure()” method, to check if the restricted security mode is already initialized or not.

taoliult:
@keithc-ca 
The PR is ready for review. For some review suggestions I updated the codes, others I replied the questions. Please help to review and advice. Thanks so much.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+            }
+        }
+    }
+
+    /**
+     * Add restricted security properties.
+     *
+     * @param props the java.security properties
+     */
+    private static void setProperties(Properties props, RestrictedSecurityProperties properties) {
```
keithc-ca, line 327: Suggest removing `RestrictedSecurityProperties properties` and using `restricts` instead.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+        exception.printStackTrace();
+        System.exit(1);
+    }
+
+    private static void printStackTraceAndExit(String message) {
+        printStackTraceAndExit(new RuntimeException(message));
+    }
+
+    /**
+     * This nested class is for save and operate the restricted security
```
keithc-ca, line 436: Again, "nested" isn't relevant to the purpose of this class; and the description doesn't read well.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+        // For SecureRandom.
+        String jdkSecureRandomProvider;
+        String jdkSecureRandomAlgorithm;
+
+        // Provider with argument (provider name + optional argument).
+        private List<String> providers;
+        // Provider without argument.
+        private List<String> providersSimpleName;
+        // The map is keyed by provider name.
+        private Map<String, Constraint[]> providerConstraints;
```
keithc-ca, line 465: These fields should be `final`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+                // Remove the provider's class package names if there are.
+                pos = providerName.lastIndexOf('.');
+                providerName = (pos < 0) ? providerName : providerName.substring(pos + 1, providerName.length());
+                // Provider without arguments and package names.
+                providersSimpleName.add(pNum - 1, providerName);
+
+                if (debug != null) {
+                    debug.println(
+                            "Loaded restricted security provider: " + providers.get(pNum - 1) + " with simple name: "
+                                    + providersSimpleName.get(pNum - 1));
```
keithc-ca, line 589: Suggest replacing `providersSimpleName.get(pNum - 1))` with just `providerName`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+            }
+        }
+
+        /**
+         * Check if the Service is allowed in restricted security mode.
+         *
+         * @param service the Service to check
+         * @return true if the Service is allowed
+         */
+        private boolean isRestrictedServiceAllowed(Service service) {
```
keithc-ca, line 739: `private` should only be used for features which are only used from within the class.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+                    return true;
+                }
+            }
+            if (debug != null) {
+                debug.println("Security constraints check."
+                        + " Service type: " + type
+                        + " Algorithm: " + algorithm
+                        + " is NOT allowed in provider " + providerName);
+            }
+            // Go through all the constraints for the provider.
```
keithc-ca, line 815: This comment doesn't read well; perhaps it should simply be deleted (I think line 816 is sufficient).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+            if (!java.security.Provider.class.isAssignableFrom(providerClazz)) {
+                if (debug != null) {
+                    debug.println("The provider class " + providerName + " does not extend java.security.Provider.");
+                }
+                // For class doesn't extend java.security.Provider, no need to
+                // check allowed or not allowed, always return true to load it.
+                return true;
+            }
+            return isRestrictedProviderAllowed(providerName);
+        }
```
keithc-ca, line 881: Suggest this be refactored into `isProviderAllowed(Class<?>)`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+         *
+         * @param string input string for checking
+         * @return true if the input string is asterisk
+         */
+        private static boolean isAsterisk(String string) {
+            return "*".equals(string);
+        }
+
+        /**
+         * Nested class for provider's constraints
```
keithc-ca, line 1038: Please remove "Nested" and expand on the purpose of this class (using full sentences and proper punctuation).

keithc-ca (changes requested):

src/java.base/share/classes/java/security/Security.java
```diff
@@ -144,10 +144,13 @@ private static void initialize() {
...
-        boolean fipsEnabled = FIPSConfigurator.configureFIPS(props);
-        if (sdebug != null) {
-            sdebug.println(fipsEnabled ? "FIPS mode enabled.": "FIPS mode disabled.");
+        // Load restricted security mode properties.
+        if (RestrictedSecurity.shouldEnable()) {
+            boolean restrictedSecurityEnabled = RestrictedSecurity.configure(props);
+            if (sdebug != null) {
+                sdebug.println(restrictedSecurityEnabled ? "Restricted security mode enabled."
+                        : "Restricted security mode disabled.");
+            }
```
keithc-ca, line 153: This seems like a step backward. In an earlier version of this change, the check that `RestrictedSecurity.configure()` was called at most once was internal to that method. I don't think that responsibility belongs here (at the call site). Please restore the behavior by updating `RestrictedSecurity.configure()` to detect a second call and fail if that occurs.

taoliult: @keithc-ca 

Ok, I will restore the previous version of this Security.java, to remove the RestrictedSecurity.shouldEnable() check.

And for detecting second call in "RestrictedSecurity.configure()". Right now, the current behaviour is, when go into the configure() method, if the securityEnabled is already true, that mean it is a second call, so it will throw the runtime exception saying "it can't be initialized twice". If the securityEnabled is false, that mean, either it is the first call, which is good. Or it is the second call but shouldEnableSecurity is false, and in this, the configure() method do nothing and just return false, means the restricted security is not support and no need to enable, or the user did not set semeru.restrictedsecurity or semeru.fips property.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca I updated the comments to mention the semeru.fips system property.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca 

For the line 182, the method "isProviderAllowed(Class<?> providerClazz)". Before this method is called from the ServiceLoader.java, it had the checks for the providerClazz. Checks from ServiceLoader.java are below: 

```
        if (clazz == null) {
            fail(service, "Provider " + cn + " not found");
        }

        int mods = clazz.getModifiers();
        if (!Modifier.isPublic(mods)) {
            fail(service, clazz + " is not public");
        }

        if (!RestrictedSecurity.isProviderAllowed(clazz)) {
            // We're in restricted security mode which does not allow this provider,
            // skip it.
            return null;
        }
```

and 

```
                    Class<?> clazz = nextProviderClass();
                    if (clazz == null)
                        return false;

                    if (clazz.getModule().isNamed()) {
                        // ignore class if in named module
                        continue;
                    }

                    if (service.isAssignableFrom(clazz)) {
                        if (!RestrictedSecurity.isProviderAllowed(clazz)) {
                            // We're in restricted security mode which does not allow this provider,
                            // skip it.
                            continue;
                        }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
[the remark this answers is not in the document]
taoliult: For the checks in "configure()", please see the latest comment.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: For the checks in "configure()", please see the latest comment.

taoliult:
@keithc-ca
The PR is ready for review. For some review suggestions I updated the codes, others I replied the questions. Please help to review and advice. Thanks so much.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1042 @@
...
+        }
+        String keyStore = restricts.keyStore;
+        if (!isNullOrBlank(keyStore)) {
+            // SSL property "javax.net.ssl.keyStore" set at the JVM level via system properties.
+            System.setProperty("javax.net.ssl.keyStore", keyStore);
+        }
+    }
+
+    /**
+     * Check the restricted security properties
```
keithc-ca, line 385: Please finish sentences with a period. Also, line 388

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1042 @@
...
+
+        // Check if the SunsetDate expired.
+        if (isPolicySunset(restricts.descSunsetDate)) {
+            printStackTraceAndExit("Restricted security policy expired.");
+        }
+
+        // Check secure random settings.
+        if (isNullOrBlank(restricts.jdkSecureRandomProvider)
+                || isNullOrBlank(restricts.jdkSecureRandomAlgorithm)) {
+            printStackTraceAndExit("Restricted security mode secure random is null.");
```
keithc-ca, line 401: Not necessarily `null`; suggest "missing".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1042 @@
...
+        System.exit(1);
+    }
+
+    private static void printStackTraceAndExit(String message) {
+        printStackTraceAndExit(new RuntimeException(message));
+    }
+
+    /**
+     * This class is used for save and operate the restricted security
+     * mode properties which are loaded from the java.security file.
```
keithc-ca, line 448: This doesn't read well; suggest:
```java
     * This class is used to save and operate on restricted security
     * properties which are loaded from the java.security file.
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1042 @@
...
+                // Remove the provider's class package names if there are.
+                pos = providerName.lastIndexOf('.');
+                providerName = (pos < 0) ? providerName : providerName.substring(pos + 1, providerName.length());
+                // Provider without arguments and package names.
+                providersSimpleName.add(pNum - 1, providerName);
+
+                if (debug != null) {
+                    debug.println(
+                            "Loaded restricted security provider: " + providers.get(pNum - 1) + " with simple name: "
+                                    + providerName);
```
keithc-ca, line 600: Suggest folding this differently:
```java
                            "Loaded restricted security provider: " + providers.get(pNum - 1)
                                    + " with simple name: " + providerName);
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1042 @@
...
+        /**
+         * Load security constraints with type, algorithm, attributes.
+         *
+         * Example:
+         * RestrictedSecurity1.jce.provider.1 = SUN [{CertPathBuilder, PKIX, *},
+         * {Policy, JavaPolicy, *}, {CertPathValidator, *, *}].
+         */
+        private void initConstraints() {
+            for (int pNum = 1; pNum <= providersSimpleName.size(); pNum++) {
+
```
keithc-ca, line 659: Please remove this blank line.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1042 @@
...
+                        // Each constraint must includes 3 fields(type, algorithm, attributes).
+                        if (input.length != 3) {
+                            printStackTraceAndExit("Constraint format is incorrect: " + providerInfo);
+                        }
+
+                        String inType = input[0].trim();
+                        String inAlgorithm = input[1].trim();
+                        String inAttributes = input[2].trim();
+
+                        // Each attribute must includes 2 fields(key and value) or *.
```
keithc-ca, line 711: Space before `(`, please.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1042 @@
...
+            providerName = (pos < 0) ? providerName : providerName.substring(0, pos);
+
+            Constraint[] constraints = providerConstraints.get(providerName);
+
+            if ((constraints == null) || (constraints.length == 0)) {
+                // Allow this provider with no constraints.
+                return true;
+            }
+
+            // Go into this provider constraints check.
```
keithc-ca, line 765: Suggest:
```java
            // Check the constraints of this provider.
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1042 @@
...
+            String type = service.getType();
+            String algorithm = service.getAlgorithm();
+
+            for (Constraint constraint : constraints) {
+
```
keithc-ca, line 770: Please remove this blank line.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1042 @@
...
+                        break;
+                    case '}':
+                    case ']':
+                    case ')':
+                        if (deque.isEmpty() || (deque.removeFirst().charValue() != ch)) {
+                            return false;
+                        }
+                        break;
+                    default:
+                        break;
```
keithc-ca, line 1010: Case labels are not indented extra relative to the enclosing `switch` statement.
```java
                switch (ch) {
                case '{':
                    deque.addFirst('}');
                    break;
...
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1042 @@
...
+         *
+         * @param string input string for checking
+         * @return true if the input string is asterisk
+         */
+        private static boolean isAsterisk(String string) {
+            return "*".equals(string);
+        }
+
+        /**
+         * This class is used for define the provider's constraints.
```
keithc-ca, line 1027: This doesn't read well; suggest:
```java
         * A class representing the constraints of a provider.
```

keithc-ca (changes requested):

taoliult:
@keithc-ca   The PR is ready for review. Please help to review and advice. Thanks so much.

taoliult:
@keithc-ca 

As we talked in slack, I added the checks in three methods of ProviderList.java.

In the ProviderList.insertAt() method, even we had the provider checks in ServiceLoader.java and ProviderConfig.java, to stop the non-FIPS provider loading. But adding the isProviderAllowed() checks in insertAt() will avoid NPE, when the customer try to add their own provider by calling ProviderList.insertAt() or ProviderList.add() methods, and then calling ProviderList.remove() to remove it. Because the provider will be added into the list but it can not be loaded, so the provider in the list will be as null.

When running the extended.openjdk tests, I can see the test case failure changed from NPE to “java.lang.Exception: Provider not added” which is more readable, in those test cases which add and remove the test dummy providers.

In the ProviderList.newList(), add the similar checks with the same reason as ProviderList.insertAt() method.

In the getService() method, and the inner class ServiceList’s tryGet() method, add the isServiceAllowed() check, to stop the non-FIPS services returned.

Right now, the PR is ready for review. Please help to review and advice. Thanks so much.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1040 @@
...
+            // e.g. SunPKCS11-NSS-FIPS, remove argument -NSS-FIPS.
+            int pos = providerName.indexOf('-');
+            providerName = (pos < 0) ? providerName : providerName.substring(0, pos);
+
+            Constraint[] constraints = providerConstraints.get(providerName);
+
+            if ((constraints == null) || (constraints.length == 0)) {
+                // Allow this provider with no constraints.
+                return true;
+            }
```
keithc-ca, line 762: This doesn't seem right. Services from a provider not mentioned in `java.security` should not be allowed.
Perhaps this should be:
```java
            if (constraints == null) {
                // Disallow unknown providers.
                return false;
            } else if (constraints.length == 0) {
                // Allow this provider with no constraints.
                return true;
            }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1040 @@
...
+
+            for (Constraint constraint : constraints) {
+                String cType = constraint.type;
+                String cAlgorithm = constraint.algorithm;
+                String cAttribute = constraint.attributes;
+
+                boolean cTypePut = isAsterisk(cType)
+                        || type.equals(cType);
+                boolean cAlgorithmPut = isAsterisk(cAlgorithm)
+                        || algorithm.equals(cAlgorithm);
```
keithc-ca, line 776: Algorithms should be case-insensitive, this should use `equalsIgnoreCase()` instead of `equals()`.

This can be simplified, avoiding the need for *Put variables that are tested repeatedly:
```java
                if (!isAsterisk(cType) && !type.equals(cType)) {
                    // The constraint doesn't apply to the service type.
                    continue;
                }
                if (!isAsterisk(cAlgorithm) && !algorithm.equalsIgnoreCase(cAlgorithm)) {
                    // The constraint doesn't apply to the service algorith.
                    continue;
                }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1040 @@
...
+                                + " Algorithm " + algorithm
+                                + " is allowed in provider " + providerName);
+                    }
+                    return true;
+                }
+
+                // For type and algorithm match, and attribute is not *.
+                // Then continue checking attributes.
+                if (cTypePut && cAlgorithmPut) {
+                    String[] cAttributeArray = constraint.attributes.split(":");
```
keithc-ca, line 793: Please reuse `cAttribute` here and on line 812.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1040 @@
...
+                    for (String attribute : cAttributeArray) {
+                        String[] input = attribute.split("=", 2);
+
+                        String cName = input[0].trim();
+                        String cValue = input[1].trim();
+                        String sValue = service.getAttribute(cName);
+                        if ((sValue == null) && !cValue.equalsIgnoreCase(sValue)) {
+                            // Any of the attribute not match, return service is not allowed
+                            return false;
+                        }
```
keithc-ca, line 805: Shouldn't a service that doesn't have the expected attribute be disallowed?
```java
                    if ((sValue == null) || !cValue.equalsIgnoreCase(sValue)) {
                        // If any attribute doesn't match, return service is not allowed.
                        return false;
                    }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1040 @@
...
+                debug.println("Checking the provider " + providerName + " in restricted security mode.");
+            }
+
+            // Remove the provider class package name if there is.
+            int pos = providerName.lastIndexOf('.');
+            providerName = (pos < 0) ? providerName : providerName.substring(pos + 1, providerName.length());
+
+            // Remove argument, e.g. -NSS-FIPS, if there is.
+            pos = providerName.indexOf('-');
+            providerName = (pos < 0) ? providerName : providerName.substring(0, pos);
```
keithc-ca, line 845: Perhaps this substring extraction should happen in the reverse order (select the portion before a `-`, _then_ choose the part after the last `.`)?

src/java.base/share/classes/sun/security/jca/ProviderList.java
```diff
@@ -117,6 +119,11 @@ public static ProviderList insertAt(ProviderList providerList, Provider p,
         if (providerList.getProvider(p.getName()) != null) {
             return providerList;
         }
+        if (!RestrictedSecurity.isProviderAllowed(p.getName())) {
```
keithc-ca, line 122: Suggest a local rather than call `p.getName()` twice.

src/java.base/share/classes/sun/security/jca/ProviderList.java
```diff
@@ -385,6 +403,11 @@ public Service getService(String type, String name) {
                 Provider p = getProvider(pList.get(i).provider);
                 Service s = p.getService(type, name);
                 if (s != null) {
+                    if (!RestrictedSecurity.isServiceAllowed(s)) {
+                        // We're in restricted security mode which does
+                        // not allow this service, return null.
+                        return null;
+                    }
```
keithc-ca, line 410: Like below, I think we want to just ignore services that are not allowed; one from a later provider may be allowed.
```java
                if ((s != null) && RestrictedSecurity.isServiceAllowed(s)) {
                    // We found a service that is allowed in restricted security mode.
                    return s;
                }
```
Same comment for 420... below.

src/java.base/share/classes/sun/security/jca/ProviderList.java
```diff
@@ -530,13 +558,23 @@ private Service tryGet(int index) {
                     // simple lookup
                     Service s = p.getService(type, algorithm);
                     if (s != null) {
+                        if (!RestrictedSecurity.isServiceAllowed(s)) {
+                            // We're in restricted security mode which does
+                            // not allow this service, skip it.
+                            continue;
+                        }
                         addService(s);
                     }
```
keithc-ca, line 567: I think this is clearer if we avoid the negation here; suggest:
```java
                    if ((s != null) && RestrictedSecurity.isServiceAllowed(s)) {
                        addService(s);
                    }
```
Like for line 572... below.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+                            }
+                        }
+
+                        Constraint constraint = new Constraint(inType, inAlgorithm, inAttributes);
+
+                        if (debug != null) {
+                            debug.println("Loading constraints for security provider: " + providerName
+                                    + " with constraints type: " + constraint.type
+                                    + " algorithm: " + constraint.algorithm
+                                    + " attributes: " + constraint.attributes);
```
keithc-ca, line 729: Please reuse locals `inType`, `inAlgorithm` and `inAttributes` here.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: I don't see a response to this question.

src/java.base/share/classes/sun/security/jca/ProviderList.java
```diff
@@ -146,6 +154,17 @@ public static ProviderList remove(ProviderList providerList, String name) {
...
     public static ProviderList newList(Provider ... providers) {
+        if (RestrictedSecurity.isEnabled()) {
+            List<Provider> allowedProviders = new ArrayList<>(Arrays.asList(providers));
+            for (Provider p : providers) {
+                if (!RestrictedSecurity.isProviderAllowed(p.getName())) {
+                    // We're in restricted security mode which does not allow this provider,
+                    // remove it from the list.
+                    allowedProviders.remove(p);
+                }
+            }
```
keithc-ca, line 165: (I thought I made this comment earlier) There's no need to put providers in the list unless they're allowed, avoiding the cost of finding those that need to be removed.
```java
            List<Provider> allowedProviders = new ArrayList<>();
            for (Provider p : providers) {
                if (RestrictedSecurity.isProviderAllowed(p.getName())) {
                    // This provider is allowed, add it the list.
                    allowedProviders.add(p);
                }
            }
```

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca 

These codes, first remove the class package name if there is, which is in the fix format using ‘.’ to separate the packages. So, after remove the package name, we can make sure only the class name left. And then remove the argument(e.g. -NSS-FIPS) from the class name if there is. 

If reverse the order, first to remove the class name argument, which start with ‘-‘, and then remove the class package name. But if there is a ‘-‘ also in the package name, for example, com.test-provider.SunPKCS11-NSS-FIPS. If I first select the portion before a ‘-‘, then the “com.test” will be returned and it is wrong.

So I suggest, we first choose the part after the last ‘.’ which is the fix format for the class packages, to make sure we get the correct class name, and then select the portion before a ‘-‘ to remove the class name argument if there is.

taoliult:
@keithc-ca
The PR is ready for review. For some review suggestions I updated the codes, others I replied the questions. Please help to review and advice. Thanks so much.

@jasonkatonica @WilburZjh 
FYI.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: The example you give (`com.test-provider.SunPKCS11-NSS-FIPS`) is not possible to write in the Java (source) language. Yes, the JVM will accept classes and package names containing `-`, but you have to write them in some other language (and you can't use them directly in Java source code).

I was thinking of a different example, say `com.sample.provider.SunPKCS11-NSS-FIPS.v2`. The current code will select `v2` which is probably not what we want.

How would you propose we deal with these (exceptional) situations?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+        private void initProviders() {
+            if (debug != null) {
+                debug.println("Loading restricted security providers.");
+            }
+
+            for (int pNum = 1;; ++pNum) {
+                String providerInfo = securityProps
+                        .getProperty(propsPrefix + ".jce.provider." + pNum);
+
+                if (providerInfo == null || providerInfo.trim().isEmpty()) {
```
keithc-ca, line 575: Missing parentheses:
```java
                if ((providerInfo == null) || providerInfo.trim().isEmpty()) {
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+         * List audit info if userSecurityAudit is true, default as false.
+         */
+        private void listAudit() {
+            System.out.println();
+            System.out.println("Restricted Security Audit Info:");
+            System.out.println("===============================");
+
+            for (int num = 1;; ++num) {
+                String desc = securityProps.getProperty("RestrictedSecurity" + num + ".desc.name");
+                if (desc == null || desc.trim().isEmpty()) {
```
keithc-ca, line 893: Missing parentheses:
```java
                if ((desc == null) || desc.trim().isEmpty()) {
```

keithc-ca (changes requested):

taoliult:
@keithc-ca
The PR is ready for review. Please help to review and advice. Thanks so much.

@jasonkatonica @WilburZjh
FYI.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+                String providerName = providersSimpleName.get(pNum - 1);
+                String providerInfo = securityProps
+                        .getProperty(propsPrefix + ".jce.provider." + pNum);
+
+                if (debug != null) {
+                    debug.println("Loading constraints for security provider: " + providerName);
+                }
+
+                // Check if the provider has constraints
+                if (providerInfo.indexOf('[') < 0 && providerInfo.indexOf(']') < 0) {
```
keithc-ca, line 668: Please finish the sentence with a period.
As we've already ensure the brackets are balanced, it's sufficient to check for just `[`:
```java
                if (providerInfo.indexOf('[') < 0) {
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+                        debug.println("No constraints for security provider: " + providerName);
+                    }
+                    providerConstraints.put(providerName, new Constraint[0]);
+                    continue;
+                }
+
+                // Remove the whitespaces in the format separator if there are.
+                providerInfo = providerInfo.trim()
+                        .replaceAll("\\[\\s*\\{", "[{")
+                        .replaceAll("\\}\\s*\\]", "}]")
```
keithc-ca, line 679: There's no need to match zero whitespace in these two lines:
```java
                        .replaceAll("\\[\\s+\\{", "[{")
                        .replaceAll("\\}\\s+\\]", "}]")
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -0,0 +1,1053 @@
...
+                        .replaceAll("\\}\\s*,\\s*\\{", "},{");
+
+                int startIndex = providerInfo.indexOf("[{");
+                int endIndex = providerInfo.indexOf("}]");
+
+                // Provider with constraints.
+                if ((startIndex > 0) && (endIndex > 0)) {
```
keithc-ca, line 686: This should ensure that there's only one `[{` and one `}]` that occur in that order.
```java
                int startIndex = providerInfo.lastIndexOf("[{");
                int endIndex = providerInfo.indexOf("}]");

                // Provider with constraints.
                if ((startIndex > 0) && (endIndex > startIndex)) {
```
(`areBracketsBalanced()` would accept `"[{ ... }], [{ ... }]"`)

keithc-ca (changes requested):

taoliult:
@keithc-ca 

Yes, I listed and compared all the security providers and their algorithms, between the Java Restricted Security Mode by using FIPS policy (-Dsemeru.restrictedsecurity=1) and the current released Semeru FIPS by using “-Dsemeru.fips=true”. The listed security providers are in the same order, offering the same algorithms.

And I also compared the 5 OpenJDK and JCK test results (sanity.openjdk, extended.openjdk, sanity.jck, extended.jck and special.jck) between the Java Restricted Security Mode by using FIPS policy (-Dsemeru.restrictedsecurity=1) and the current released Semeru FIPS by using “-Dsemeru.fips=true”. There are more expected failures since the test cases are trying to create its own providers which is not allowed in Java Restricted Security Mode FIPS. Except these expected failures, other test results are same as before. So, the tests look good.

@jasonkatonica @WilburZjh   FYI.

keithc-ca:

pshipton:
Will this be backported to jdk8?

taoliult:
@pshipton Yes, I am working on the backport to OpenJDK8.

@keithc-ca @jasonkatonica @WilburZjh fyi.

PR #531, opened 2022-12-06
Merge master HEAD into openj9-staging

JasonFengJ9:
Conflicts:
```
	make/autoconf/jdk-version.m4
	make/conf/version-numbers.conf
	src/java.base/share/classes/java/lang/String.java
	src/java.base/share/classes/java/lang/Thread.java
	src/java.base/share/classes/java/lang/foreign/VaList.java
	src/java.base/share/classes/jdk/internal/foreign/CABI.java
	src/java.base/share/classes/jdk/internal/foreign/PlatformLayouts.java
	src/java.base/share/classes/jdk/internal/foreign/SystemLookup.java
	src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
	src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
	src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
	src/java.base/share/classes/jdk/internal/foreign/abi/x64/sysv/CallArranger.java
	src/java.base/share/classes/jdk/internal/foreign/abi/x64/windows/CallArranger.java
	src/java.base/unix/classes/java/lang/ProcessEnvironment.java
	src/jdk.management.agent/share/conf/management.properties
	test/jdk/java/foreign/CallGeneratorHelper.java
	test/jdk/java/foreign/LibraryLookupTest.java
	test/jdk/java/foreign/NativeTestHelper.java
	test/jdk/java/foreign/SafeFunctionAccessTest.java
	test/jdk/java/foreign/StdLibTest.java
	test/jdk/java/foreign/TestClassLoaderFindNative.java
	test/jdk/java/foreign/TestDowncallScope.java
	test/jdk/java/foreign/TestDowncallStack.java
	test/jdk/java/foreign/TestFunctionDescriptor.java
	test/jdk/java/foreign/TestHeapAlignment.java
	test/jdk/java/foreign/TestMatrix.java
	test/jdk/java/foreign/TestNULLAddress.java
	test/jdk/java/foreign/TestNative.java
	test/jdk/java/foreign/TestNulls.java
	test/jdk/java/foreign/TestScopedOperations.java
	test/jdk/java/foreign/TestSegments.java
	test/jdk/java/foreign/TestStringEncoding.java
	test/jdk/java/foreign/TestUpcallAsync.java
	test/jdk/java/foreign/TestUpcallException.java
	test/jdk/java/foreign/TestUpcallHighArity.java
[...]

PR #545, opened 2023-02-01
Use GetExtensionEvents to derive extension event numbers

babsingh:
Currently, extension event numbers are hard-coded in the tests. This is
invalid since the extension event numbers can vary between
implementations. Instead, GetExtensionEvents should be used to derive
the extension event numbers.

Related: https://github.com/eclipse-openj9/openj9/pull/16501
Related: https://github.com/eclipse-openj9/openj9/issues/16167

### DIFF
--- a/test/hotspot/jtreg/serviceability/jvmti/vthread/BreakpointInYieldTest/libBreakpointInYieldTest.cpp
+++ b/test/hotspot/jtreg/serviceability/jvmti/vthread/BreakpointInYieldTest/libBreakpointInYieldTest.cpp
@@ -20,6 +20,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * ===========================================================================
+ */
 
 #include <string.h>
 #include "jvmti.h"
@@ -250,13 +255,13 @@ Agent_OnLoad(JavaVM *jvm, char *options, void *reserved) {
     LOG("error in JVMTI SetEventNotificationMode: %d\n", err);
   }
 
-  err = jvmti->SetEventNotificationMode(JVMTI_ENABLE, EXT_EVENT_VIRTUAL_THREAD_MOUNT, NULL);
+  err = set_ext_event_notification_mode(jvmti, JVMTI_ENABLE, "VirtualThreadMount", NULL);
   if (err != JVMTI_ERROR_NONE) {
     LOG("error in JVMTI SetEventNotificationMode: %d\n", err);
     return JNI_ERR;
   }
 
-  err = jvmti->SetEventNotificationMode(JVMTI_ENABLE, EXT_EVENT_VIRTUAL_THREAD_UNMOUNT, NULL);
+  err = set_ext_event_notification_mode(jvmti, JVMTI_ENABLE, "VirtualThreadUnmount", NULL);
   if (err != JVMTI_ERROR_NONE) {
     LOG("error in JVMTI SetEventNotificationMode: %d\n", err);
     return JNI_ERR;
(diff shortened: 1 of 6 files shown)


babsingh:
**GetExtensionEvents JVMTI spec:** https://download.java.net/java/early_access/jdk19/docs/specs/jvmti.html#GetExtensionEvents

**Ports:**
- JDK-next: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/545
- JDK20: https://github.com/ibmruntimes/openj9-openjdk-jdk20/pull/13
- JDK19: https://github.com/ibmruntimes/openj9-openjdk-jdk19/pull/65

test/lib/jdk/test/lib/jvmti/jvmti_common.h
```diff
@@ -880,6 +881,27 @@ set_event_notification_mode(jvmtiEnv* jvmti, JNIEnv* jni, jvmtiEventMode mode, j
...
+    err = JVMTI_ERROR_NOT_AVAILABLE;
+  } else {
+    err = jvmti->SetEventNotificationMode(mode, (jvmtiEvent)info->extension_event_index, event_thread);
+  }
+
+  return err;
+}
+
+static void
+set_ext_event_notification_mode(jvmtiEnv* jvmti, JNIEnv* jni, jvmtiEventMode mode, const char* ename, jthread event_thread) {
```
keithc-ca, line 900: I don't see anything that suggests it's safe to assume this is only used in a C++ context (and the extension is not `.hpp`), in which case it's not reasonable (nor necessary) to overload the name `set_ext_event_notification_mode`.

Also, the code should be written in the style of the surrounding code:
```c
static jvmtiError
set_ext_event_notification_mode(jvmtiEnv* jvmti, jvmtiEventMode mode, const char* ename, jthread event_thread) {
  jvmtiExtensionEventInfo* info = find_ext_event(jvmti, ename);
  if (NULL == info) {
    LOG("jvmti_common set_ext_event_callback: Extension event was not found: %s\n", ename);
    return JVMTI_ERROR_NOT_AVAILABLE;
  }
  return jvmti->SetEventNotificationMode(mode, (jvmtiEvent)info->extension_event_index, event_thread);
}
```

keithc-ca (changes requested):

test/hotspot/jtreg/serviceability/jvmti/vthread/BreakpointInYieldTest/libBreakpointInYieldTest.cpp
```diff
@@ -303,11 +308,9 @@ Java_BreakpointInYieldTest_check(JNIEnv *jni, jclass cls) {
   err = jvmti->SetEventNotificationMode(JVMTI_DISABLE, JVMTI_EVENT_VIRTUAL_THREAD_START, NULL);
   check_jvmti_status(jni, err, "check: error in JVMTI SetEventNotificationMode: disable VIRTUAL_THREAD_START");
 
-  err = jvmti->SetEventNotificationMode(JVMTI_DISABLE, EXT_EVENT_VIRTUAL_THREAD_MOUNT, NULL);
-  check_jvmti_status(jni, err, "check: error in JVMTI SetEventNotificationMode: disable VIRTUAL_THREAD_MOUNT");
+  set_ext_event_notification_mode(jvmti, jni, JVMTI_DISABLE, "VirtualThreadMount", NULL);
```
keithc-ca, line 311: I think this should only replace the call to `SetEventNotificationMode()`, leaving the use of `check_jvmti_status()` alone.

keithc-ca:
Why is this necessary at all? We go to some effort in [jvmti.h.m4](https://github.com/eclipse-openj9/openj9/blob/master/runtime/include/jvmti.h.m4#L746-L747) to (apparently) match the RI event numbers. Perhaps that's where the problem lies?

babsingh:
> Why is this necessary at all? We go to some effort in [jvmti.h.m4](https://github.com/eclipse-openj9/openj9/blob/master/runtime/include/jvmti.h.m4#L746-L747) to (apparently) match the RI event numbers. Perhaps that's where the problem lies?

@gacholio mentioned that the current behaviour is invalid and not supported by the JVMTI specification. More details can be found in the comments surrounding https://github.com/eclipse-openj9/openj9/pull/16501#issuecomment-1410478678.

gacholio:
Extension events are by definition not in the header file since they are not expected to be the same between VM implementations. There is a specified way to get the numbers from the data returned from the extension query.

keithc-ca:
Ok got it. Thanks.

keithc-ca:
Once we settle on a solution here, we should be proposing the fix upstream.

test/hotspot/jtreg/serviceability/jvmti/vthread/BreakpointInYieldTest/libBreakpointInYieldTest.cpp
babsingh: Updated.

test/lib/jdk/test/lib/jvmti/jvmti_common.h
babsingh: In JDK-next, the compiler used to build `jvmti_common.h` should support function overloading in C. Also, there are existing instances of function overloading here: `set_event_notification_mode`. Should we still be worried about overloading?

babsingh:
> we should be proposing the fix upstream.

by upstream, do we mean https://github.com/openjdk/jdk?

test/lib/jdk/test/lib/jvmti/jvmti_common.h
babsingh: Removed 5-argument version.

gacholio:
As this is throw-away code, the most likely way to get changes accepted upstream would be to make the smallest change possible, not the most elegant.

keithc-ca:
> by upstream, do we mean https://github.com/openjdk/jdk?

Yes.

keithc-ca:
I think this should wait for the resolution of https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/531#issuecomment-1419832430.

babsingh:
> I think this should wait for the resolution of https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/531#issuecomment-1419832430.

None of the files in this PR are updated in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/531#issuecomment-1419832430. There is no noticeable dependency. I don't see a reason to wait.

keithc-ca:
> I don't see a reason to wait.

The changes here appear to be the same as in https://github.com/ibmruntimes/openj9-openjdk-jdk20/pull/13; once I'm satisfied with the latter, I think it should be safe to merge this.

keithc-ca (approved):

PR #552, opened 2023-02-22
Fix memory leaks related to openssl EC pathways

jasonkatonica:
This update fixes a few memory leaks associated with the creation of EC private and public keys.

This openssl api `BN_bn2bin` described here https://www.openssl.org/docs/man1.0.2/man3/BN_bin2bn.html states that a new `BIGNUM` is created if the last argument is null to the API. In our case the last argument is indeed null so we are leaking memory in the EC paths since we never free the big number returned. 

A cleanup goto was also added to simplify the code and ensure that all cleanup of allocated memory is done. We make explicit calls to then free all allocated memory in the cleanup.

Signed-off-by: Jason Katonica [katonica@us.ibm.com](mailto:katonica@us.ibm.com)

### DIFF
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -2538,9 +2538,6 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECEncodeGFp
     BIGNUM *yBN = NULL;
     BIGNUM *nBN = NULL;
     BIGNUM *hBN = NULL;
-    EC_GROUP *group = NULL;
-    EC_POINT *generator = NULL;
-    BN_CTX *ctx = NULL;
     int ret = 0;
 
     nativeA = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, a, 0));
@@ -2676,9 +2673,6 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECEncodeGF2m
     BIGNUM *yBN = NULL;
     BIGNUM *nBN = NULL;
     BIGNUM *hBN = NULL;
-    EC_GROUP *group = NULL;
-    EC_POINT *generator = NULL;
-    BN_CTX *ctx = NULL;
     int ret = 0;
 
     nativeA = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, a, 0));
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
  * ===========================================================================
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -2316,6 +2316,14 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECCreatePublicKey
     (*env)->ReleasePrimitiveArrayCritical(env, x, nativeX, JNI_ABORT);
     (*env)->ReleasePrimitiveArrayCritical(env, y, nativeY, JNI_ABORT);
 
+    if (NULL != xBN) {
+        (*OSSL_BN_free)(xBN);
+    }
+
+    if (NULL != yBN) {
+        (*OSSL_BN_free)(yBN);
+    }
+
     if (0 == ret) {
         return -1;
     }
@@ -2354,6 +2362,10 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECCreatePrivateKey
 
     (*env)->ReleasePrimitiveArrayCritical(env, s, nativeS, JNI_ABORT);
 
+    if (NULL != sBN) {
+        (*OSSL_BN_free)(sBN);
+    }
+
     if (0 == ret) {
         return -1;
     }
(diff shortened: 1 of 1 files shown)


closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2354,14 +2362,138 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECCreatePrivateKey
...
 
-/* Encode an EC Elliptic Curve over a Prime Field
+/* Encode an EC Elliptic Curve over a Prime Field */
+EC_KEY* ECEncodeGFp(BIGNUM *aBN,
+                    BIGNUM *bBN,
+                    BIGNUM *pBN,
+                    BIGNUM *xBN,
+                    BIGNUM *yBN,
+                    BIGNUM *nBN,
+                    BIGNUM *hBN)
```
keithc-ca, line 2383: This function (and `ECEncodeGF2m()`) should be static with the name at the start of a new line:
```c
static EC_KEY *
ECEncodeGFp(BIGNUM *aBN,
            BIGNUM *bBN,
            BIGNUM *pBN,
            BIGNUM *xBN,
            BIGNUM *yBN,
            BIGNUM *nBN,
            BIGNUM *hBN)
{
```
with a space before `*` consistently.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2354,14 +2362,138 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECCreatePrivateKey
...
+    int ret = 0;
+
+    ctx = (*OSSL_BN_CTX_new)();
+    if (NULL == ctx) {
+        goto cleanup;
+    }
+
+    group = (*OSSL_EC_GROUP_new_curve_GFp)(pBN, aBN, bBN, ctx);
+    if (NULL == group)
+        goto cleanup;
```
keithc-ca, line 2398: Please don't omit braces (please correct throughout):
```c
    if (NULL == group) {
        goto cleanup;
    }
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2354,14 +2362,138 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECCreatePrivateKey
...
+
+    generator = (*OSSL_EC_POINT_new)(group);
+    if (NULL == generator)
+        goto cleanup;
+
+    ret = (*OSSL_EC_POINT_set_affine_coordinates_GF2m)(group, generator, xBN, yBN, ctx);
+    if (0 == ret)
+        return NULL;
```
keithc-ca, line 2467: Shouldn't this be `goto cleanup;`?

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
jasonkatonica: This update ( and the same type of update to `ECEncodeGF2m` ) is complete.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
jasonkatonica: This update should be done throughout all the code that I changed.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
jasonkatonica: Good catch here. I agree that we want to `goto cleanup` in order to cleanup any allocated objects, then return `NULL` instead of directly returning `NULL` in this spot.

keithc-ca:
In functions `ECEncodeGF2m()` and `ECEncodeGFp()`, these locals are no longer used which may lead to (fatal) compile warnings:
```c
    EC_GROUP *group = NULL;
    EC_POINT *generator = NULL;
    BN_CTX *ctx = NULL;
```
I see other uses of `convertJavaBItoBN()` that appear to leak; e.g. int these functions:
* `createRSAPublicKey()`
* `createRSAPrivateCrtKey()`

In each function, if any allocation of a BN fails, other BNs are not freed.

jasonkatonica:
> In functions `ECEncodeGF2m()` and `ECEncodeGFp()`, these locals are no longer used which may lead to (fatal) compile warnings:
> 
> ```c
>     EC_GROUP *group = NULL;
>     EC_POINT *generator = NULL;
>     BN_CTX *ctx = NULL;
> ```
> 
These unused variables were removed via commit abbc995


> I see other uses of `convertJavaBItoBN()` that appear to leak; e.g. int these functions:
> 
>     * `createRSAPublicKey()`
> 
>     * `createRSAPrivateCrtKey()`
> 
> 
> In each function, if any allocation of a BN fails, other BNs are not freed.

This is indeed true that additional work is needed on a case by case basis in order to ensure no additional leaks are found and we correctly make use of the BN methods in such a way that either openssl frees these big number OR we free them depending on the situation. I did not however attempt to go through all the uses since it will take quite a bit of time to test and ensure proper use. My goal for this particular change was to fix the reported leak impacting users for https://github.com/eclipse-openj9/openj9/issues/16451 for that particular testcase.

keithc-ca:
The changes we make here will need to be back-ported to earlier versions. I think we should fix all the leaks we know about, not just those users have encountered and reported. Even if it would mean just an extra pull request here, that gets multiplied by the number of other versions to which the fixes are to be applied.

jasonkatonica:
OK, its going to take a while to go through the source of each of the OpenSSL APIs that make use of BNs. Some APIs take on the responsibility of freeing the BNs in use, some like this case do not free the BNs. The documentation is often times missing in this area so it will involve looking through the openssl source code for each of our uses of an API that uses BNs, testing the path for memory exhaustion, then fixing each leak found. If you would like this fix to be delayed for that work to be completed, and nobody needs this particular fix, we can revisit this fix at some later date.

keithc-ca:
I'd be content if the problems I identified are fixed.

jasonkatonica:
OK sounds good I believe that all the review comments have been addressed besides the usage of BNs in other OpenSSL APIs. 

There are no other known or reported leaks in this area, that I am aware of, although I wouldn't be surprised if there are more issues there with BNs. This work will certainly need to be done to ensure there are no further issues.

pshipton:
jenkins compile xlinux jdknext

keithc-ca:
@jasonkatonica Any progress on fixing the leaks I pointed out in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/552#issuecomment-1456867151?

jasonkatonica:
Hi @keithc-ca I have not gone through and tested / fixed the other potential leaks that have similar patterns here. I am hoping myself or somebody else on my team gets a chance in the coming weeks even though we are not quite sure at this time if there are more problem areas we have to tackle or not.

pshipton:
Is it possible to merge what we have if it fixes the customer problem? "in the coming weeks" sounds like it will miss the 0.38 release, for which we want everything merged before M2, which is March 20.

jasonkatonica:
I do believe it is possible to merge this code to fix the specific reported issue at https://github.com/eclipse-openj9/openj9/issues/16451 that had a testcase available as I was able to run these specific test in a loop for extended periods of time. I also agree that there could be potential other leaks due to the way we are using `BN_bn2bin` in openssl.

pshipton:
FYI the customer agrees this resolves the issues seen in the field.
https://github.com/eclipse-openj9/openj9/issues/16451#issuecomment-1466772213

keithc-ca:
I created https://github.com/eclipse-openj9/openj9/issues/16910 to track fixing the remaining problems in the integration with OpenSSL.

keithc-ca:
Jenkins test sanity alinux64 jdknext

pshipton:
I ran some security related testing in https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/2078/, which all passed.
jdk_security1, 2, 3, 4 all in mode 1 (-XX:-UseCompressedOops)

pshipton:
I've stopped the sanity.function testing, which probably doesn't use OpenSSL in any significant way anyway. The sanity.functional testing is hung in JCL_Test / JCL_TEST_Java-Lang, which I've seen before on jdk21, but didn't get around to creating an issue for. https://github.com/eclipse-openj9/openj9/issues/16920 is created now.

keithc-ca:
I think the hanging test is a known problem: https://github.com/eclipse-openj9/openj9/issues/16779.

keithc-ca (approved):

pshipton:
@jasonkatonica ideally this should be backported this week before the 0.38 release M2.

PR #554, opened 2023-02-27
Expand OpenSSL Version granularity

WilburZjh:
Additional granularity for tracking versions is needed in Native code to make better branching decisions based upon the library versions. Currently, there is differentiation only between <=1.0.x and >=1.1.0. Since, version 1.1.1 will be supported until 2023-09-11 and versions including 1.1.0, 1.0.2, 1.0.1, 1.0.0, and 0.9.8 are no longer supported. We need to add support for separate versions to deprecate/add features in the future.

### DIFF
--- a/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -48,6 +48,10 @@ public class NativeCrypto {
     public static final int SHA5_384 = 3;
     public static final int SHA5_512 = 4;
 
+    public static final long OPENSSL_VERSION_1_0_0 = 0x1_00_00_000L;
+    public static final long OPENSSL_VERSION_1_1_0 = 0x1_01_00_000L;
+    public static final long OPENSSL_VERSION_3_0_0 = 0x3_00_00_000L;
+
     private static final Cleaner ECKeyCleaner = CleanerFactory.cleaner();
 
     private static final boolean useNativeCrypto = Boolean.parseBoolean(
@@ -60,14 +64,13 @@ public class NativeCrypto {
         private static final NativeCrypto instance = new NativeCrypto();
     }
 
-    //ossl_vers:
+    //ossl_vers will be either:
     // -1 : library load failed
-    //  0 : openssl 1.0.x
-    //  1 : openssl 1.1.x or newer
-    private final int ossl_ver;
+    // or one of the OPENSSL_VERSION_x_x_x constants
+    private final long ossl_ver;
 
-    private static int loadCryptoLibraries() {
-        int osslVersion;
+    private static long loadCryptoLibraries() {
+        long osslVersion;
 
         try {
             // load jncrypto JNI library
@@ -91,7 +94,7 @@ public class NativeCrypto {
 
     @SuppressWarnings("removal")
     private NativeCrypto() {
-        ossl_ver = AccessController.doPrivileged((PrivilegedAction<Integer>) () -> loadCryptoLibraries()).intValue();
+        ossl_ver = AccessController.doPrivileged((PrivilegedAction<Long>) () -> loadCryptoLibraries()).longValue();
     }
 
     /**
@@ -112,7 +115,7 @@ public class NativeCrypto {
      *
      * @return the OpenSSL library version if it is available
      */
-    public static final int getVersionIfAvailable() {
+    public static final long getVersionIfAvailable() {
 /*[IF CRIU_SUPPORT]*/
         if (InternalCRIUSupport.isCheckpointAllowed()) {
             return -1;
@@ -186,7 +189,7 @@ public class NativeCrypto {
 
     /* Native digest interfaces */
 
-    private static final native int loadCrypto(boolean trace);
+    private static final native long loadCrypto(boolean trace);
 
     public final native long DigestCreateContext(long nativeBuffer,
                                                  int algoIndex);
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
  * ===========================================================================
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -29,6 +29,7 @@
 #include <openssl/ecdh.h>
 #include <openssl/pkcs12.h>
 
+#include <ctype.h>
 #include <jni.h>
 #include <stdio.h>
 #include <stdint.h>
@@ -38,10 +39,15 @@
 #include "jdk_crypto_jniprovider_NativeCrypto.h"
 #include "NativeCrypto_md.h"
 
-#define OPENSSL_VERSION_1_0 "OpenSSL 1.0."
-#define OPENSSL_VERSION_1_1 "OpenSSL 1.1."
+#define OPENSSL_VERSION_CODE(major, minor, fix, patch) \
+        ((((jlong)(major)) << 28) | ((minor) << 20) | ((fix) << 12) | (patch))
+
+#define OPENSSL_VERSION_1_0_0 OPENSSL_VERSION_CODE(1, 0, 0, 0)
+#define OPENSSL_VERSION_1_1_0 OPENSSL_VERSION_CODE(1, 1, 0, 0)
+#define OPENSSL_VERSION_2_0_0 OPENSSL_VERSION_CODE(2, 0, 0, 0)
 /* Per new OpenSSL naming convention starting from OpenSSL 3, all major versions are ABI and API compatible. */
-#define OPENSSL_VERSION_3_X "OpenSSL 3."
+#define OPENSSL_VERSION_3_0_0 OPENSSL_VERSION_CODE(3, 0, 0, 0)
+#define OPENSSL_VERSION_4_0_0 OPENSSL_VERSION_CODE(4, 0, 0, 0)
 
 /* needed for OpenSSL 1.0.2 Thread handling routines */
 #define CRYPTO_LOCK 1
@@ -261,13 +267,39 @@ static void printErrors(void)
     fflush(stderr);
 }
 
+/*
+ * We use a 8 digit map (ABBCCDDD) to represent the version of openssl.
+ * A is the major version,
+ * BB is the minor version,
+ * CC is the fix,
+ * DDD is the patch that could be present in any version.
+ * For example, if an openssl version is in this scheme 1.2.3.d
+ * where major is 1, minor is 2, fix is 3 and patch is d -> 4.
+ * So the result would be 0x10203004, where A is 1, BB is 02, CC is 03, DDD is 004.
+ */
+static jlong extractVersionToJlong(const char *astring)
+{
+    long major = 0;
+    long minor = 0;
+    long fix = 0;
+    long patch = 0;
+    char patch_char = 0;
+    if (sscanf(astring, "OpenSSL %ld.%ld.%ld%c", &major, &minor, &fix, &patch_char) < 3) {
+        return -1;
+    }
+    if (isalpha(patch_char)) {
+        patch = tolower(patch_char) - 'a' + 1;
+    }
+    return (jlong)OPENSSL_VERSION_CODE(major, minor, fix, patch);
+}
+
 static void *crypto_library = NULL;
 /*
  * Class:     jdk_crypto_jniprovider_NativeCrypto
  * Method:    loadCrypto
  * Signature: (Z)I
  */
-JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
+JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
   (JNIEnv *env, jclass thisObj, jboolean trace)
 {
     char *error;
@@ -276,7 +308,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     /* Determine the version of OpenSSL. */
     OSSL_version_t* OSSL_version;
     const char * openssl_version;
-    int ossl_ver;
+    jlong ossl_ver = 0;
 
     /* Load OpenSSL Crypto library */
     crypto_library = load_crypto_library(trace);
@@ -310,7 +342,8 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
         } else {
             openssl_version = (*OSSL_version)(0); /* get OPENSSL_VERSION */
             /* Ensure the OpenSSL version is "OpenSSL 1.0.x" */
-            if (0 != strncmp(openssl_version, OPENSSL_VERSION_1_0, strlen(OPENSSL_VERSION_1_0))) {
+            ossl_ver = extractVersionToJlong(openssl_version);
+            if (!((OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0))) {
                 if (trace) {
                     fprintf(stderr, "Error loading OpenSSL: Incompatible OpenSSL version found: %s\n", openssl_version);
                     fflush(stderr);
@@ -319,13 +352,13 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
                 crypto_library = NULL;
                 return -1;
             }
-            ossl_ver = 0;
         }
     } else {
         openssl_version = (*OSSL_version)(0); /* get OPENSSL_VERSION */
         /* Ensure the OpenSSL version is "OpenSSL 1.1.x" or "OpenSSL 3.x.x". */
-        if ((0 != strncmp(openssl_version, OPENSSL_VERSION_1_1, strlen(OPENSSL_VERSION_1_1)))
-        && (0 != strncmp(openssl_version, OPENSSL_VERSION_3_X, strlen(OPENSSL_VERSION_3_X)))
+        ossl_ver = extractVersionToJlong(openssl_version);
+        if (!((OPENSSL_VERSION_1_1_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_2_0_0))
+        &&  !((OPENSSL_VERSION_3_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_4_0_0))
         ) {
             if (trace) {
                 fprintf(stderr, "Error loading OpenSSL: Incompatible OpenSSL version found: %s\n", openssl_version);
@@ -335,7 +368,6 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
             crypto_library = NULL;
             return -1;
         }
-        ossl_ver = 1;
     }
 
     /* Load the function symbols for OpenSSL errors. */
@@ -344,7 +376,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     OSSL_get_error = (OSSL_get_error_t*)find_crypto_symbol(crypto_library, "ERR_get_error");
 
     /* Load Threading routines for OpenSSL 1.0.2 */
-    if (0 == ossl_ver) {
+    if (ossl_ver < OPENSSL_VERSION_1_1_0) {
         OSSL_CRYPTO_num_locks = (OSSL_CRYPTO_num_locks_t*)find_crypto_symbol(crypto_library, "CRYPTO_num_locks");
         OSSL_CRYPTO_THREADID_set_numeric = (OSSL_CRYPTO_THREADID_set_numeric_t*)find_crypto_symbol(crypto_library, "CRYPTO_THREADID_set_numeric");
         OSSL_OPENSSL_malloc = (OSSL_OPENSSL_malloc_t*)find_crypto_symbol(crypto_library, "CRYPTO_malloc");
@@ -360,7 +392,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     OSSL_sha384 = (OSSL_sha_t*)find_crypto_symbol(crypto_library, "EVP_sha384");
     OSSL_sha512 = (OSSL_sha_t*)find_crypto_symbol(crypto_library, "EVP_sha512");
 
-    if (1 == ossl_ver) {
+    if (ossl_ver >= OPENSSL_VERSION_1_1_0) {
         OSSL_MD_CTX_new = (OSSL_MD_CTX_new_t*)find_crypto_symbol(crypto_library, "EVP_MD_CTX_new");
         OSSL_MD_CTX_reset = (OSSL_MD_CTX_reset_t*)find_crypto_symbol(crypto_library, "EVP_MD_CTX_reset");
         OSSL_MD_CTX_free = (OSSL_MD_CTX_free_t*)find_crypto_symbol(crypto_library, "EVP_MD_CTX_free");
@@ -394,7 +426,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     OSSL_DecryptFinal = (OSSL_DecryptFinal_t*)find_crypto_symbol(crypto_library, "EVP_DecryptFinal");
 
     /* Load the functions symbols for OpenSSL ChaCha20 algorithms. (Need OpenSSL 1.1.x or above) */
-    if (1 == ossl_ver) {
+    if (ossl_ver >= OPENSSL_VERSION_1_1_0) {
         OSSL_chacha20 = (OSSL_cipher_t*)find_crypto_symbol(crypto_library, "EVP_chacha20");
         OSSL_chacha20_poly1305 = (OSSL_cipher_t*)find_crypto_symbol(crypto_library, "EVP_chacha20_poly1305");
     } else {
@@ -405,7 +437,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     /* Load the functions symbols for OpenSSL RSA algorithm. */
     OSSL_RSA_new = (OSSL_RSA_new_t*)find_crypto_symbol(crypto_library, "RSA_new");
 
-    if (1 == ossl_ver) {
+    if (ossl_ver >= OPENSSL_VERSION_1_1_0) {
         OSSL_RSA_set0_key = (OSSL_RSA_set0_key_t*)find_crypto_symbol(crypto_library, "RSA_set0_key");
         OSSL_RSA_set0_factors = (OSSL_RSA_set0_factors_t*)find_crypto_symbol(crypto_library, "RSA_set0_factors");
         OSSL_RSA_set0_crt_params = (OSSL_RSA_set0_key_t*)find_crypto_symbol(crypto_library, "RSA_set0_crt_params");
@@ -524,14 +556,14 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
         (NULL == OSSL_EC_KEY_check_key) ||
         (NULL == OSSL_PKCS12_key_gen) ||
         /* Check symbols that are only available in OpenSSL 1.1.x and above */
-        ((1 == ossl_ver) && ((NULL == OSSL_chacha20) || (NULL == OSSL_chacha20_poly1305))) ||
+        ((ossl_ver >= OPENSSL_VERSION_1_1_0) && ((NULL == OSSL_chacha20) || (NULL == OSSL_chacha20_poly1305))) ||
         /* Check symbols that are only available in OpenSSL 1.0.x and above */
-        ((NULL == OSSL_CRYPTO_num_locks) && (0 == ossl_ver)) ||
-        ((NULL == OSSL_CRYPTO_THREADID_set_numeric) && (0 == ossl_ver)) ||
-        ((NULL == OSSL_OPENSSL_malloc) && (0 == ossl_ver)) ||
-        ((NULL == OSSL_OPENSSL_free) && (0 == ossl_ver)) ||
-        ((NULL == OSSL_CRYPTO_THREADID_set_callback) && (0 == ossl_ver)) ||
-        ((NULL == OSSL_CRYPTO_set_locking_callback) && (0 == ossl_ver))) {
+        ((NULL == OSSL_CRYPTO_num_locks) && (ossl_ver < OPENSSL_VERSION_1_1_0)) ||
+        ((NULL == OSSL_CRYPTO_THREADID_set_numeric) && (ossl_ver < OPENSSL_VERSION_1_1_0)) ||
+        ((NULL == OSSL_OPENSSL_malloc) && (ossl_ver < OPENSSL_VERSION_1_1_0)) ||
+        ((NULL == OSSL_OPENSSL_free) && (ossl_ver < OPENSSL_VERSION_1_1_0)) ||
+        ((NULL == OSSL_CRYPTO_THREADID_set_callback) && (ossl_ver < OPENSSL_VERSION_1_1_0)) ||
+        ((NULL == OSSL_CRYPTO_set_locking_callback) && (ossl_ver < OPENSSL_VERSION_1_1_0))) {
             if (trace) {
                 fprintf(stderr, "Error loading OpenSSL: One or more of the required symbols are missing in the crypto library: %s\n", openssl_version);
             }
@@ -549,7 +581,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
                 free(library_path);
             }
         }
-        if (0 == ossl_ver) {
+        if (ossl_ver < OPENSSL_VERSION_1_1_0) {
             if (0 != thread_setup()) {
                 unload_crypto_library(crypto_library);
                 crypto_library = NULL;
--- a/src/java.base/share/classes/com/sun/crypto/provider/SunJCE.java
+++ b/src/java.base/share/classes/com/sun/crypto/provider/SunJCE.java
@@ -336,7 +336,7 @@ public final class SunJCE extends Provider {
         attrs.clear();
         attrs.put("SupportedKeyFormats", "RAW");
 
-        if (useNativeChaCha20Cipher && (NativeCrypto.getVersionIfAvailable() >= 1)) {
+        if (useNativeChaCha20Cipher && (NativeCrypto.getVersionIfAvailable() >= NativeCrypto.OPENSSL_VERSION_1_1_0)) {
             ps("Cipher", "ChaCha20",
                     "com.sun.crypto.provider.NativeChaCha20Cipher$ChaCha20Only",
                     null, attrs);


closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +264,36 @@ static void printErrors(void)
...
+ * The ossl_ver should be 10101011 ( k is the 11th letter value for DD here... ).
+ *
+*/
+long ver_to_l(const char* astring) {
+    long major = 0;
+    long minor = 0;
+    long revision = 0;
+    long diff = 0;
+    sscanf(astring, "OpenSSL %ld.%ld.%ld", &major, &minor, &revision);
+    char last_char = astring[strlen(astring) - 1];
```
keithc-ca, line 282: The return value of `sscanf()` should not be ignored.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +264,36 @@ static void printErrors(void)
...
+    if (isalpha(last_char)) {
+        diff = abs(last_char - 'a') + 1;
```
keithc-ca, line 284: Using `tolower()` would be an improvement.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +264,36 @@ static void printErrors(void)
     fflush(stderr);
 }
 
+/*
+ * We use a 8 digit map (AABBCCDD) to represent the version of openssl.
```
keithc-ca, line 268: The structure described in `NativeCrypto.java` used three digits for `DDD`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +264,36 @@ static void printErrors(void)
...
+    long major = 0;
+    long minor = 0;
+    long revision = 0;
+    long diff = 0;
+    sscanf(astring, "OpenSSL %ld.%ld.%ld", &major, &minor, &revision);
+    char last_char = astring[strlen(astring) - 1];
+    if (isalpha(last_char)) {
+        diff = abs(last_char - 'a') + 1;
+    }
+    return major * 10000000 + minor * 100000 + revision * 1000 + diff;
```
keithc-ca, line 286: Many missing parentheses; care is needed converting from `long` to `jlong`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +264,36 @@ static void printErrors(void)
...
+}
+
+long ossl_ver;
```
keithc-ca, line 289: This should continue to be a local variable within `loadCrypto()`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -310,7 +335,8 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
         } else {
             openssl_version = (*OSSL_version)(0); /* get OPENSSL_VERSION */
             /* Ensure the OpenSSL version is "OpenSSL 1.0.x" */
-            if (0 != strncmp(openssl_version, OPENSSL_VERSION_1_0, strlen(OPENSSL_VERSION_1_0))) {
+            ossl_ver = ver_to_l(openssl_version);
+            if(!(ossl_ver >= OPENSSL_VERSION_1_0_0 && ossl_ver < OPENSSL_VERSION_1_1_0)) {
```
keithc-ca, line 339: Missing parentheses and space after `if`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -319,14 +345,13 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
...
         }
     } else {
         openssl_version = (*OSSL_version)(0); /* get OPENSSL_VERSION */
         /* Ensure the OpenSSL version is "OpenSSL 1.1.x" or "OpenSSL 3.x.x". */
-        if ((0 != strncmp(openssl_version, OPENSSL_VERSION_1_1, strlen(OPENSSL_VERSION_1_1)))
-        && (0 != strncmp(openssl_version, OPENSSL_VERSION_3_X, strlen(OPENSSL_VERSION_3_X)))
-        ) {
+        ossl_ver = ver_to_l(openssl_version);
+        if(!(ossl_ver >= OPENSSL_VERSION_1_1_0 && ossl_ver < OPENSSL_VERSION_2_0_0)
+        && !(ossl_ver >= OPENSSL_VERSION_3_0_0 && ossl_ver < OPENSSL_VERSION_4_0_0)) {
```
keithc-ca, line 354: Missing parentheses and space after `if`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -344,7 +368,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     OSSL_get_error = (OSSL_get_error_t*)find_crypto_symbol(crypto_library, "ERR_get_error");
 
     /* Load Threading routines for OpenSSL 1.0.2 */
-    if (0 == ossl_ver) {
+    if (ossl_ver >= OPENSSL_VERSION_1_0_0 && ossl_ver < OPENSSL_VERSION_1_1_0) {
```
keithc-ca, line 371: Missing parentheses (here and lines 387-388, 422-423, 554-555).
Suggest using "between" pattern:  `(OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0)`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -549,7 +578,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
                 free(library_path);
             }
         }
-        if (0 == ossl_ver) {
+        if((ossl_ver >= OPENSSL_VERSION_1_0_0) && (ossl_ver < OPENSSL_VERSION_1_1_0)) {
```
keithc-ca, line 581: Missing space after `if`.

keithc-ca (changes requested):

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -47,6 +47,7 @@ public class NativeCrypto {
     public static final int SHA2_256 = 2;
     public static final int SHA5_384 = 3;
     public static final int SHA5_512 = 4;
+    public static final int OPENSSL_VERSION_1_1_0 = 10100000;
```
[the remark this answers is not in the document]
WilburZjh: Edited

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -62,12 +63,13 @@ private static final class InstanceHolder {
 
     //ossl_vers:
     // -1 : library load failed
-    //  0 : openssl 1.0.x
-    //  1 : openssl 1.1.x or newer
-    private final int ossl_ver;
+    //  10000000 : openssl 1.0.x
+    //  10100000 : openssl 1.1.x
+    //  30000000 : openssl 3.x.x
```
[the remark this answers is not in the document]
WilburZjh: Define more constants for 1.1.x and 3.x.x.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -38,10 +39,12 @@
...
 
-#define OPENSSL_VERSION_1_0 "OpenSSL 1.0."
-#define OPENSSL_VERSION_1_1 "OpenSSL 1.1."
+#define OPENSSL_VERSION_1_0_0 10000000
+#define OPENSSL_VERSION_1_1_0 10100000
+#define OPENSSL_VERSION_2_0_0 20000000
 /* Per new OpenSSL naming convention starting from OpenSSL 3, all major versions are ABI and API compatible. */
-#define OPENSSL_VERSION_3_X "OpenSSL 3."
+#define OPENSSL_VERSION_3_0_0 30000000
+#define OPENSSL_VERSION_4_0_0 40000000
```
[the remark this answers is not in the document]
WilburZjh: Changed the value to `long`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +264,36 @@ static void printErrors(void)
...
 
+/*
+ * We use a 8 digit map (AABBCCDD) to represent the version of openssl.
+ * AA is the major version,
+ * BB is the minor version,
+ * CC is the revision, and DD for the letter that could be present in any version.
+ * For example, if an openssl version is in this scheme 1.1.1k.
+ * The ossl_ver should be 10101011 ( k is the 11th letter value for DD here... ).
+ *
+*/
```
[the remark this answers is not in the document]
WilburZjh: Removed

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +264,36 @@ static void printErrors(void)
...
+long ver_to_l(const char* astring) {
```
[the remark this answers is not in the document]
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Got the return value and added the check for it as well.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Added.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Hi @keithc-ca , may I know where the three digits are used for DDD? I didnt find it...

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Added more parentheses and converted from `long` to `jlong`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Put it within `loadCrypto()` function.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Changed all the patterns by using "between".

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
WilburZjh: I changed them to the hexadecimal format, but it seems that [Jenkins](https://hyc-runtimes-jenkins.swg-devops.com/job/Build_JDKnext_x86-64_linux_Personal/1096/) doesnt recognize the underscore.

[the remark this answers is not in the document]
keithc-ca: Java allows underscores in numeric literals, but C does not. Perhaps we want a helper macro:
```c
#define OPENSSL_VERSION(major, minor) (((major) << 24) | ((minor) << 20))
```
which might be sufficient on its own, or we can use it to define these:
```
#define OPENSSL_VERSION_1_0_0 OPENSSL_VERSION(1, 0)
#define OPENSSL_VERSION_1_1_0 OPENSSL_VERSION(1, 1)
#define OPENSSL_VERSION_2_0_0 OPENSSL_VERSION(2, 0)
#define OPENSSL_VERSION_3_0_0 OPENSSL_VERSION(3, 0)
#define OPENSSL_VERSION_4_0_0 OPENSSL_VERSION(4, 0)
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
keithc-ca: The name has not changed, nor has the formatting I suggested for the parameter type (`*` binds more tightly to the parameter than the target type).

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: The return value should be 3 (given we hope to match three values).

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: See this in `NativeCrypto.java`:
```java
    public static final long OPENSSL_VERSION_1_0_0 = 0x1_00_00_000L;
```

keithc-ca: ... and the multiplication by 1000 on line 286 (now 289) below.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: Why did you add the redundant `ossl_ver < 0` part of the test?
```c
            if ((ossl_ver < 0) || !((OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0))) {
```
can be
```c
            if (!((OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0))) {
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: Redundant test `(ossl_ver < 0)` here too. Please put `) {` on a separate line.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Since we didnt use `tolower()` function before, `abs()` was used to handle the UpperCase and LowerCase.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Because we check the return value of `sscanf()`, if it doesnt match 3 values, it will return -1.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,22 +264,47 @@ static void printErrors(void)
...
+ * For example, if an openssl version is in this scheme 1.1.1k.
+ * The ossl_ver should be 10101011 ( k is the 11th letter value for DD here... ).
+ */
+static jlong ver_to_l(const char* astring)
+{
+    long major = 0;
+    long minor = 0;
+    long revision = 0;
+    long diff = 0;
+    int num = sscanf(astring, "OpenSSL %ld.%ld.%ld", &major, &minor, &revision);
```
keithc-ca, line 281: I think `last_char` should be acquired here:
```c
    char last_char = 0;
    if (sscanf(astring, "OpenSSL %ld.%ld.%ld%c", &major, &minor, &revision, &last_char) < 3) {
        return -1;
    }
```
The string returned by openssl doesn't have that character at the end, e.g.
```sh
$ openssl version
OpenSSL 1.1.1f  31 Mar 2020
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,22 +264,47 @@ static void printErrors(void)
...
  * Method:    loadCrypto
  * Signature: (Z)I
  */
-JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
+JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
   (JNIEnv *env, jclass thisObj, jboolean trace)
 {
     char *error;
     typedef const char* OSSL_version_t(int);
+    long ossl_ver;
```
keithc-ca, line 303: Please leave this declaration in its original location, and initialize to zero.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -344,7 +371,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     OSSL_get_error = (OSSL_get_error_t*)find_crypto_symbol(crypto_library, "ERR_get_error");
 
     /* Load Threading routines for OpenSSL 1.0.2 */
-    if (0 == ossl_ver) {
+    if ((OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0)) {
```
keithc-ca, line 374: If `ossl_ver < OPENSSL_VERSION_1_1_0` we've already verified that `OPENSSL_VERSION_1_0_0 <= ossl_ver` so only the first test is needed.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -360,7 +387,8 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     OSSL_sha384 = (OSSL_sha_t*)find_crypto_symbol(crypto_library, "EVP_sha384");
     OSSL_sha512 = (OSSL_sha_t*)find_crypto_symbol(crypto_library, "EVP_sha512");
 
-    if (1 == ossl_ver) {
+    if (((OPENSSL_VERSION_1_1_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_2_0_0))
+    || ((OPENSSL_VERSION_3_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_4_0_0))) {
```
keithc-ca, line 391: I think this condition can be simplified to `ossl_ver >= OPENSSL_VERSION_1_1_0`; likewise other places where the compound condition is used.

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
WilburZjh: If we define the `OPENSSL_VERSION` like these, should we change the way we calculate ossl_ver in `ver_to_l()` function at line 289?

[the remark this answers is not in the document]
keithc-ca: Actually, the name `OPENSSL_VERSION` is used already, suggest `OPENSSL_VERSION_CODE` instead:
```c
#define OPENSSL_VERSION_CODE(major, minor) (((major) << 24) | ((minor) << 20))

#define OPENSSL_VERSION_1_0_0 OPENSSL_VERSION_CODE(1, 0)
#define OPENSSL_VERSION_1_1_0 OPENSSL_VERSION_CODE(1, 1)
#define OPENSSL_VERSION_2_0_0 OPENSSL_VERSION_CODE(2, 0)
#define OPENSSL_VERSION_3_0_0 OPENSSL_VERSION_CODE(3, 0)
#define OPENSSL_VERSION_4_0_0 OPENSSL_VERSION_CODE(4, 0)
```

keithc-ca:
Please update the commit message to correct the spelling of "granularity" and split lines so they're no more than 72 characters long; see https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Added.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Backed and initialized.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: No. The return value indicates how may patterns were matched; see https://cplusplus.com/reference/cstdio/sscanf/.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited and removed `(ossl_ver < 0)`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Removed `(ossl_ver < 0)`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -38,10 +39,13 @@
 #include "jdk_crypto_jniprovider_NativeCrypto.h"
 #include "NativeCrypto_md.h"
 
-#define OPENSSL_VERSION_1_0 "OpenSSL 1.0."
-#define OPENSSL_VERSION_1_1 "OpenSSL 1.1."
+#define OPENSSL_VERSION_CODE(major, minor) (((major) << 24) | ((minor) << 20))
+#define OPENSSL_VERSION_1_0_0 OPENSSL_VERSION_CODE(1, 0)
```
keithc-ca, line 43: Please leave a blank line before the definition of `OPENSSL_VERSION_1_0_0`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
WilburZjh: I was wondering if we define different variables like these, should we change the way we calculate `ossl_ver` in `extractVersionToJlong()` function at line 290?

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +265,38 @@ static void printErrors(void)
     fflush(stderr);
 }
 
+/*
+ * We use a 8 digit map (AABBCCDD) to represent the version of openssl.
+ * AA is the major version,
+ * BB is the minor version,
+ * CC is the revision, and DD for the letter that could be present in any version.
+ * For example, if an openssl version is in this scheme 1.1.1k.
+ * The ossl_ver should be 10101011 ( k is the 11th letter value for DD here... ).
```
keithc-ca, line 274: This comment needs to be updated: versions are expressed as up to 9 hex digits (three for DDD). For example, `1.1.1k` would map to `0x1010100b`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -276,7 +305,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     /* Determine the version of OpenSSL. */
     OSSL_version_t* OSSL_version;
     const char * openssl_version;
-    int ossl_ver;
+    long ossl_ver = 0;
```
keithc-ca, line 308: The type should be `jlong` to match `extractVersionToJlong()`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +265,38 @@ static void printErrors(void)
...
+    long revision = 0;
+    long diff = 0;
+    char last_char = 0;
+    if (sscanf(astring, "OpenSSL %ld.%ld.%ld%c", &major, &minor, &revision, &last_char) < 3) {
+        return -1;
+    }
+    last_char = tolower(astring[strlen(astring) - 1]);
+    if (isalpha(last_char)) {
+        diff = (last_char - 'a') + 1;
+    }
```
keithc-ca, line 289: This still gets the wrong character. We want (the lowercase version of) the one seen by `sscanf()`.
Also, letters are not contiguous in EBCDIC, so I think we want the following which works in both ASCII and EBCDIC:
```c
    last_char = tolower(last_char);
    if (('a' <= last_char) && (last_char <= 'i')) {
        diff = last_char - 'a' + 1;
    } else if (('j' <= last_char) && (last_char <= 'r')) {
        diff = last_char - 'j' + 10;
    } else if (('s' <= last_char) && (last_char <= 'z')) {
        diff = last_char - 's' + 19;
    }
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -524,14 +553,14 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
...
-        ((NULL == OSSL_OPENSSL_malloc) && (0 == ossl_ver)) ||
-        ((NULL == OSSL_OPENSSL_free) && (0 == ossl_ver)) ||
-        ((NULL == OSSL_CRYPTO_THREADID_set_callback) && (0 == ossl_ver)) ||
-        ((NULL == OSSL_CRYPTO_set_locking_callback) && (0 == ossl_ver))) {
+        ((NULL == OSSL_CRYPTO_num_locks) && ((OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0))) ||
+        ((NULL == OSSL_CRYPTO_THREADID_set_numeric) && ((OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0))) ||
+        ((NULL == OSSL_OPENSSL_malloc) && ((OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0))) ||
+        ((NULL == OSSL_OPENSSL_free) && ((OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0))) ||
+        ((NULL == OSSL_CRYPTO_THREADID_set_callback) && ((OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0))) ||
+        ((NULL == OSSL_CRYPTO_set_locking_callback) && ((OPENSSL_VERSION_1_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_1_1_0)))) {
```
keithc-ca, line 563: The second part of each of these conditions (and the one on line 581) can each be `(ossl_ver < OPENSSL_VERSION_1_1_0)`.

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: We use e.g. 10100000 for version 1.1.0 where AA is 10, BB is 10, CC is 00 and DD is 00. After changing to hexadecimal format (0x1_01_00_000L), A will be 1, BB will be 01, CC will be 00 and DDD will be 000, right?

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: We used e.g. 10101011 for version 1.1.1k where AA is 10, BB is 10, CC is 10 and DD is 11. After changing to hexadecimal format (0x1_01_01_011L), A will be 1, BB will be 01, CC will be 01 and DDD will be 011, right?

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
keithc-ca: Sorry, I should have said (28, not 24):
```c
#define OPENSSL_VERSION_CODE(major, minor) (((major) << 28) | ((minor) << 20))
```
and, yes, `extractVersionToJlong()` will have to change accordingly:
```c
    return (jlong)((major << 28) + (minor << 20) + (revision << 12) + diff);
```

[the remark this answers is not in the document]
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -38,10 +39,14 @@
 #include "jdk_crypto_jniprovider_NativeCrypto.h"
 #include "NativeCrypto_md.h"
 
-#define OPENSSL_VERSION_1_0 "OpenSSL 1.0."
-#define OPENSSL_VERSION_1_1 "OpenSSL 1.1."
+#define OPENSSL_VERSION_CODE(major, minor) (((major) << 28) | ((minor) << 20))
```
keithc-ca, line 42: I suggest this macro should accept four arguments so it can be used to avoid code duplication (the parameter names are taken from [opensslv.h](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1t/include/openssl/opensslv.h#L19)):
```c
#define OPENSSL_VERSION_CODE(major, minor, fix, patch) \
	(((major) << 28) | ((minor) << 20) | ((fix) << 12) | (patch))
```
The updated macro can then be used in `extractVersionToJlong()` with the locals renamed to match:
```c
    return (jlong)OPENSSL_VERSION_CODE(major, minor, fix, patch);
```

WilburZjh: Edited.

WilburZjh: If we changed to this format, it seems that the version which only contains major and minor will return odd values. Check the following results.
```
1.0 formatted: FFFFFFFF
1.0.0 formatted: 10000000
1.1.1a formatted: 10101001
1.1.1k formatted: 1010100B
1.1.1.z formatted: 10101000
2.0 formatted: FFFFFFFF
5.0 formatted: FFFFFFFF
10.0 formatted: FFFFFFFF
10.1 formatted: FFFFFFFF
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: If we changed to this format, it seems that the version which only contains major and minor will return odd values. Check the following results.
```
1.0 formatted: FFFFFFFF
1.0.0 formatted: 10000000
1.1.1a formatted: 10101001
1.1.1k formatted: 1010100B
1.1.1.z formatted: 10101000
2.0 formatted: FFFFFFFF
5.0 formatted: FFFFFFFF
10.0 formatted: FFFFFFFF
10.1 formatted: FFFFFFFF
```
Hi @keithc-ca , do we need to consider these cases which only contains major and minor?

WilburZjh: One more thing that raise our attention is `major << 28` for openssl version 16.0:
```
5.0 formatted: 50000000
10.0 formatted: A0000000
10.1 formatted: A0100000
10.0.1a formatted: A0001001
16.0.0 formatted: 0
```
`major << 28` shifts beyond its capacity at 16 for major and becomes 0.

WilburZjh:
Hi @keithc-ca , If we add this following check:
```
    if (sscanf(astring, "OpenSSL %ld.%ld.%ld%c", &major, &minor, &revision, &last_char) < 3) {
        return -1;
    }
```
It seems that the version which only contains major and minor will return -1 which is `FFFFFFFF` in hex format. Check the following results.
```
1.0 formatted: FFFFFFFF
1.0.0 formatted: 10000000
1.1.1a formatted: 10101001
1.1.1k formatted: 1010100B
1.1.1.z formatted: 10101000
2.0 formatted: FFFFFFFF
5.0 formatted: FFFFFFFF
10.0 formatted: FFFFFFFF
10.1 formatted: FFFFFFFF
```
Do we need to consider these cases which only contains major and minor?

WilburZjh:
One more thing that raise our attention is `major << 28` for openssl version 16.0:
```
5.0 formatted: 50000000
10.0 formatted: A0000000
10.1 formatted: A0100000
10.0.1a formatted: A0001001
16.0.0 formatted: 0
```
`major << 28` shifts beyond its capacity at 16 for major and becomes 0.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -38,10 +39,14 @@
 #include "jdk_crypto_jniprovider_NativeCrypto.h"
 #include "NativeCrypto_md.h"
 
-#define OPENSSL_VERSION_1_0 "OpenSSL 1.0."
-#define OPENSSL_VERSION_1_1 "OpenSSL 1.1."
+#define OPENSSL_VERSION_CODE(major, minor, fix, patch) (((major) << 28) | ((minor) << 20) | ((fix) << 12) | (patch))
```
keithc-ca, line 42: Please split this as in my earlier suggestion.
Of course, all uses of `OPENSSL_VERSION_CODE` now need four arguments.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +266,44 @@ static void printErrors(void)
...
+ * where major is 1, minor is 2, reversion is 3 and last char is d -> 4.
+ * So the result would be 0x10203004.
+ */
+static jlong extractVersionToJlong(const char *astring)
+{
+    long major = 0;
+    long minor = 0;
+    long revision = 0;
+    long diff = 0;
+    char last_char = 0;
```
keithc-ca, line 285: Please rename these for consistency with the new definition of `OPENSSL_VERSION_CODE` (and update the documentation above accordingly).

keithc-ca (changes requested):

keithc-ca:
> `major << 28` shifts beyond its capacity at 16 for major and becomes 0.

This shouldn't be a problem with careful use of `jlong` (assuming `long` is not 64 bits wide).

keithc-ca:
> Do we need to consider these cases which only contains major and minor?

Do you anticipate openssl expressing the version with fewer than 3 components? If not, we don't have a problem.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Updated.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Updated.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +267,44 @@ static void printErrors(void)
...
+ * where major is 1, minor is 2, reversion is 3 and last char is d -> 4.
+ * So the result would be 0x10203004.
+ */
+static jlong extractVersionToJlong(const char *astring)
+{
+    long major = 0;
+    long minor = 0;
+    long fix = 0;
+    long patch = 0;
+    char last_char = 0;
```
keithc-ca, line 286: Suggest `patch_char` is a better name for this.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +267,44 @@ static void printErrors(void)
...
+        return -1;
+    }
+    last_char = tolower(last_char);
+    if (('a' <= last_char) && (last_char <= 'i')) {
+        patch = last_char - 'a' + 1;
+    } else if (('j' <= last_char) && (last_char <= 'r')) {
+        patch = last_char - 'j' + 10;
+    } else if (('s' <= last_char) && (last_char <= 'z')) {
+        patch = last_char - 's' + 19;
+    }
```
keithc-ca, line 297: I thought about this some more; on z/OS this code will be compiled with `convlit(ISO8859-1)` so the literals will be ASCII (and `sscanf()` will do the appropriate conversions), so we can use the simpler version we had earlier:
```c
    if (isalpha(patch_char)) {
        patch = tolower(patch_char) - 'a' + 1;
    }
````

File: closed/src/java.base/share/native/libjncrypto/NativeCrypto.c [hunk 29]
```diff
@@ -38,10 +39,15 @@
 #include "jdk_crypto_jniprovider_NativeCrypto.h"
 #include "NativeCrypto_md.h"
 
-#define OPENSSL_VERSION_1_0 "OpenSSL 1.0."
-#define OPENSSL_VERSION_1_1 "OpenSSL 1.1."
+#define OPENSSL_VERSION_CODE(major, minor, fix, patch) \
+        (((major) << 28) | ((minor) << 20) | ((fix) << 12) | (patch))
```
keithc-ca, line 43: Suggest adding a cast to `jlong` here to avoid loss of information when `major` is 16 or more:
```c
#define OPENSSL_VERSION_CODE(major, minor, fix, patch) \
        ((((jlong)(major)) << 28) | ((minor) << 20) | ((fix) << 12) | (patch))
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +267,44 @@ static void printErrors(void)
...
 }
 
+/*
+ * We use a 8 digit map (ABBCCDDD) to represent the version of openssl.
+ * A is the major version,
+ * BB is the minor version,
+ * CC is the revision,
+ * DDD for the letter that could be present in any version.
+ * For example, if an openssl version is in this scheme 1.2.3.d
+ * where major is 1, minor is 2, reversion is 3 and last char is d -> 4.
```
keithc-ca, line 277: Please use "fix" and "patch" instead of "revision" (or "reversion") and "last char" for consistency with the names in `OPENSSL_VERSION_CODE` and below.

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Updated.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Updated.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Edited.

WilburZjh:
> 

No, I havent seen any openssl versions with only major and minor. But do we need to add a check for this situation that only contains major and minor in case it will happen in the future?

WilburZjh:
> > Do we need to consider these cases which only contains major and minor?
> 
> Do you anticipate openssl expressing the version with fewer than 3 components? If not, we don't have a problem.

Instead, we noticed that there are some versions which contains multiple patch_chars, for example, OpenSSL 3.1.0-beta1 from this [openssl-tags](https://github.com/openssl/openssl/tags) webpage. Currently, the code only can handle single patch_char, do we need to consider this in our cases?

keithc-ca:
> No, I havent seen any openssl versions with only major and minor. But do we need to add a check for this situation that only contains major and minor in case it will happen in the future?
> 
> Instead, we noticed that there are some versions which contains multiple patch_chars, for example, `OpenSSL 3.1.0-beta1` from this [openssl-tags](https://github.com/openssl/openssl/tags) webpage. Currently, the code only can handle single patch_char, do we need to consider this in our cases?

I'm not sure there's much value in supporting versions without 3 components if we haven't seen any.

I think we should treat "3.1.0-beta1" the same as we would "3.1.0" (i.e. ignore the "-" character).

WilburZjh:
> > No, I havent seen any openssl versions with only major and minor. But do we need to add a check for this situation that only contains major and minor in case it will happen in the future?
> > Instead, we noticed that there are some versions which contains multiple patch_chars, for example, `OpenSSL 3.1.0-beta1` from this [openssl-tags](https://github.com/openssl/openssl/tags) webpage. Currently, the code only can handle single patch_char, do we need to consider this in our cases?
> 
> I'm not sure there's much value in supporting versions without 3 components if we haven't seen any.
> 
> I think we should treat "3.1.0-beta1" the same as we would "3.1.0" (i.e. ignore the "-" character).

Im okay with that.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -261,13 +267,39 @@ static void printErrors(void)
...
+    long fix = 0;
+    long patch = 0;
+    char patch_char = 0;
+    if (sscanf(astring, "OpenSSL %ld.%ld.%ld%c", &major, &minor, &fix, &patch_char) < 3) {
+        return -1;
+    }
+    if (isalpha(patch_char)) {
+        patch = tolower(patch_char) - 'a' + 1;
+    }
+    return (jlong)OPENSSL_VERSION_CODE(major, minor, fix, diff);
```
keithc-ca, line 293: `diff` must b `patch` - it won't compile otherwise.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -319,14 +352,14 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
...
     } else {
         openssl_version = (*OSSL_version)(0); /* get OPENSSL_VERSION */
         /* Ensure the OpenSSL version is "OpenSSL 1.1.x" or "OpenSSL 3.x.x". */
-        if ((0 != strncmp(openssl_version, OPENSSL_VERSION_1_1, strlen(OPENSSL_VERSION_1_1)))
-        && (0 != strncmp(openssl_version, OPENSSL_VERSION_3_X, strlen(OPENSSL_VERSION_3_X)))
-        ) {
+        ossl_ver = extractVersionToJlong(openssl_version);
+        if ((!((OPENSSL_VERSION_1_1_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_2_0_0))
+        && !((OPENSSL_VERSION_3_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_4_0_0))))
+        {
```
keithc-ca, line 362: There's an extra, unnecessary pair of parentheses at the top level of this condition: please remove that pair.
Also, please move the closing parenthesis to the line with the opening brace:
```c
        if (!((OPENSSL_VERSION_1_1_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_2_0_0))
        &&  !((OPENSSL_VERSION_3_0_0 <= ossl_ver) && (ossl_ver < OPENSSL_VERSION_4_0_0))
        ) {
```

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Updated.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
WilburZjh: Updated.

keithc-ca:
Jenkins test sanity xlinux64 jdknext

keithc-ca:

keithc-ca (approved):

keithc-ca:
The hanging test seems to be a known issue: https://github.com/eclipse-openj9/openj9/issues/16779.

PR #539, opened 2023-01-10
Lazily initialize native crypto libraries

JasonFengJ9:
Delay native crypto library loading to avoid early initialization which might cause CRIU cross architectures restore error; 
Skip native library loading before taking a checkpoint;
Reload the providers from scratch;
Create a singleton NativeCrypto instance.

~depends on https://github.com/eclipse-openj9/openj9/pull/16430~

Signed-off-by: Jason Feng <fengj@ca.ibm.com>

PR #564, opened 2023-03-10
Add support for EC key generation using native OpenSSL library and restructure ECDH key agreement.

KostasTsiounis:
As part of this update, two things happen:

1. A new class is created and equivalent functionality is added that allows the use the native OpenSSL library to perform EC key generation. As part of that, NativeECKeyPairGenerator.java is created and several other classes are amended to support it, namely SunEC to load the appropriate class, and the ECKeyPairGenerator offered by Sun as a fallback option.
2. The code for the ECDH Key Agreement is updated. In that part, helper functions are created in the NativeCrypto.c file to avoid unnecessary passing of EC_KEY pointers. A NativeECUtil.java class is, also, created, so as to bundle commonly used functionality. Lastly, the handling of freeing arrays, keys and BIGNUM structs that were part of that functionality is updated.

Signed-off by: Kostas Tsiounis kostas.tsiounis@ibm.com

### DIFF
--- a/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -48,6 +48,10 @@ public class NativeCrypto {
     public static final int SHA5_384 = 3;
     public static final int SHA5_512 = 4;
 
+    /* Define constants for the EC field types. */
+    public static final int ECField_Fp = 0;
+    public static final int ECField_F2m = 1;
+
     public static final long OPENSSL_VERSION_1_0_0 = 0x1_00_00_000L;
     public static final long OPENSSL_VERSION_1_1_0 = 0x1_01_00_000L;
     public static final long OPENSSL_VERSION_3_0_0 = 0x3_00_00_000L;
@@ -339,6 +343,15 @@ public class NativeCrypto {
                                        int tagLen);
 
     /* Native EC interfaces */
+    public final native int ECGenerateKeyPair(long key,
+                                              byte[] x,
+                                              int xLen,
+                                              byte[] y,
+                                              int yLen,
+                                              byte[] s,
+                                              int sLen,
+                                              int fieldType);
+
     public final native int ECCreatePublicKey(long key,
                                               byte[] x,
                                               int xLen,
@@ -350,35 +363,21 @@ public class NativeCrypto {
                                                byte[] s,
                                                int sLen);
 
-    public final native long ECEncodeGFp(byte[] a,
-                                         int aLen,
-                                         byte[] b,
-                                         int bLen,
-                                         byte[] p,
-                                         int pLen,
-                                         byte[] x,
-                                         int xLen,
-                                         byte[] y,
-                                         int yLen,
-                                         byte[] n,
-                                         int nLen,
-                                         byte[] h,
-                                         int hLen);
-
-    public final native long ECEncodeGF2m(byte[] a,
-                                          int aLen,
-                                          byte[] b,
-                                          int bLen,
-                                          byte[] p,
-                                          int pLen,
-                                          byte[] x,
-                                          int xLen,
-                                          byte[] y,
-                                          int yLen,
-                                          byte[] n,
-                                          int nLen,
-                                          byte[] h,
-                                          int hLen);
+    public final native long ECEncodeGF(int fieldType,
+                                        byte[] a,
+                                        int aLen,
+                                        byte[] b,
+                                        int bLen,
+                                        byte[] p,
+                                        int pLen,
+                                        byte[] x,
+                                        int xLen,
+                                        byte[] y,
+                                        int yLen,
+                                        byte[] n,
+                                        int nLen,
+                                        byte[] h,
+                                        int hLen);
 
     public final native int ECDestroyKey(long key);
 
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -115,10 +115,15 @@ typedef int OSSL_RSA_set0_factors_t(RSA *, BIGNUM *, BIGNUM *);
 typedef void OSSL_RSA_free_t (RSA *);
 typedef int OSSL_RSA_public_decrypt_t(int, const unsigned char *, unsigned char *, RSA *, int);
 typedef int OSSL_RSA_private_encrypt_t (int, const unsigned char *, unsigned char *, RSA *, int);
-typedef BIGNUM* OSSL_BN_bin2bn_t (const unsigned char *, int, BIGNUM *);
+
+typedef BIGNUM *OSSL_BN_new_t();
+typedef BIGNUM *OSSL_BN_bin2bn_t (const unsigned char *, int, BIGNUM *);
 typedef void OSSL_BN_set_negative_t (BIGNUM *, int);
 typedef void OSSL_BN_free_t (BIGNUM *);
+typedef int OSSL_BN_bn2bin_t(const BIGNUM *, unsigned char *);
+typedef int OSSL_BN_num_bits_t(const BIGNUM *);
 
+typedef int OSSL_EC_KEY_generate_key_t(EC_KEY *);
 typedef void OSSL_EC_KEY_free_t(EC_KEY *);
 typedef int OSSL_ECDH_compute_key_t(void *, size_t, const EC_POINT *, EC_KEY *, void *(*KDF)(const void *, size_t, void *, size_t *));
 typedef const EC_POINT* OSSL_EC_KEY_get0_public_key_t(const EC_KEY *);
@@ -132,6 +137,8 @@ typedef int OSSL_EC_KEY_set_group_t(EC_KEY *, const EC_GROUP *);
 typedef EC_POINT* OSSL_EC_POINT_new_t(const EC_GROUP *);
 typedef int OSSL_EC_POINT_set_affine_coordinates_GFp_t(const EC_GROUP *, EC_POINT *, const BIGNUM *, const BIGNUM *, BN_CTX *);
 typedef int OSSL_EC_POINT_set_affine_coordinates_GF2m_t(const EC_GROUP *, EC_POINT *, const BIGNUM *, const BIGNUM *, BN_CTX *);
+typedef int OSSL_EC_POINT_get_affine_coordinates_GFp_t(const EC_GROUP *, const EC_POINT *, BIGNUM *, BIGNUM *, BN_CTX *);
+typedef int OSSL_EC_POINT_get_affine_coordinates_GF2m_t(const EC_GROUP *, const EC_POINT *, BIGNUM *, BIGNUM *, BN_CTX *);
 typedef int OSSL_EC_GROUP_set_generator_t(EC_GROUP *, const EC_POINT *, const BIGNUM *, const BIGNUM *);
 typedef const EC_GROUP* OSSL_EC_KEY_get0_group_t(const EC_KEY *);
 typedef void OSSL_EC_POINT_free_t(EC_POINT *);
@@ -140,6 +147,7 @@ typedef void OSSL_BN_CTX_free_t(BN_CTX *);
 typedef int OSSL_EC_KEY_set_public_key_t(EC_KEY *, const EC_POINT *);
 typedef int OSSL_EC_KEY_check_key_t(const EC_KEY *);
 typedef int EC_set_public_key_t(EC_KEY *, BIGNUM *, BIGNUM *, int);
+typedef const BIGNUM *OSSL_EC_KEY_get0_private_key_t(const EC_KEY *);
 
 typedef int OSSL_PKCS12_key_gen_t(const char *, int, unsigned char *, int, int, int, int, unsigned char *, const EVP_MD *);
 
@@ -213,15 +221,21 @@ OSSL_RSA_set0_key_t* OSSL_RSA_set0_crt_params;
 OSSL_RSA_free_t* OSSL_RSA_free;
 OSSL_RSA_public_decrypt_t* OSSL_RSA_public_decrypt;
 OSSL_RSA_private_encrypt_t* OSSL_RSA_private_encrypt;
+
+/* Define pointers for OpenSSL BIGNUM structs. */
+OSSL_BN_new_t *OSSL_BN_new;
 OSSL_BN_bin2bn_t* OSSL_BN_bin2bn;
 OSSL_BN_set_negative_t* OSSL_BN_set_negative;
 OSSL_BN_free_t* OSSL_BN_free;
+OSSL_BN_bn2bin_t *OSSL_BN_bn2bin;
+OSSL_BN_num_bits_t *OSSL_BN_num_bits;
 
 /* Define pointers for OpenSSL functions to handle ChaCha20 algorithm. */
 OSSL_cipher_t* OSSL_chacha20;
 OSSL_cipher_t* OSSL_chacha20_poly1305;
 
 /* Define pointers for OpenSSL functions to handle EC algorithm. */
+OSSL_EC_KEY_generate_key_t *OSSL_EC_KEY_generate_key;
 OSSL_EC_KEY_free_t* OSSL_EC_KEY_free;
 OSSL_ECDH_compute_key_t* OSSL_ECDH_compute_key;
 OSSL_EC_KEY_get0_public_key_t* OSSL_EC_KEY_get0_public_key;
@@ -235,6 +249,8 @@ OSSL_EC_KEY_set_group_t* OSSL_EC_KEY_set_group;
 OSSL_EC_POINT_new_t* OSSL_EC_POINT_new;
 OSSL_EC_POINT_set_affine_coordinates_GFp_t* OSSL_EC_POINT_set_affine_coordinates_GFp;
 OSSL_EC_POINT_set_affine_coordinates_GF2m_t* OSSL_EC_POINT_set_affine_coordinates_GF2m;
+OSSL_EC_POINT_get_affine_coordinates_GFp_t *OSSL_EC_POINT_get_affine_coordinates_GFp;
+OSSL_EC_POINT_get_affine_coordinates_GF2m_t *OSSL_EC_POINT_get_affine_coordinates_GF2m;
 OSSL_EC_GROUP_set_generator_t* OSSL_EC_GROUP_set_generator;
 OSSL_EC_KEY_get0_group_t* OSSL_EC_KEY_get0_group;
 OSSL_EC_POINT_free_t* OSSL_EC_POINT_free;
@@ -243,6 +259,7 @@ OSSL_BN_CTX_free_t* OSSL_BN_CTX_free;
 OSSL_EC_KEY_set_public_key_t* OSSL_EC_KEY_set_public_key;
 OSSL_EC_KEY_check_key_t* OSSL_EC_KEY_check_key;
 EC_set_public_key_t* EC_set_public_key;
+OSSL_EC_KEY_get0_private_key_t *OSSL_EC_KEY_get0_private_key;
 
 /* Define pointers for OpenSSL functions to handle PBE algorithm. */
 OSSL_PKCS12_key_gen_t* OSSL_PKCS12_key_gen;
@@ -449,14 +466,21 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     OSSL_RSA_free = (OSSL_RSA_free_t *)find_crypto_symbol(crypto_library, "RSA_free");
     OSSL_RSA_public_decrypt = (OSSL_RSA_public_decrypt_t *)find_crypto_symbol(crypto_library, "RSA_public_decrypt");
     OSSL_RSA_private_encrypt = (OSSL_RSA_private_encrypt_t *)find_crypto_symbol(crypto_library, "RSA_private_decrypt");
+
+    /* Load the function symbols for BIGNUM manipulation. */
+    OSSL_BN_new = (OSSL_BN_new_t *)find_crypto_symbol(crypto_library, "BN_new");
     OSSL_BN_bin2bn = (OSSL_BN_bin2bn_t *)find_crypto_symbol(crypto_library, "BN_bin2bn");
     OSSL_BN_set_negative = (OSSL_BN_set_negative_t *)find_crypto_symbol(crypto_library, "BN_set_negative");
     OSSL_BN_free = (OSSL_BN_free_t *)find_crypto_symbol(crypto_library, "BN_free");
+    OSSL_BN_bn2bin = (OSSL_BN_bn2bin_t *)find_crypto_symbol(crypto_library, "BN_bn2bin");
+    OSSL_BN_num_bits = (OSSL_BN_num_bits_t *)find_crypto_symbol(crypto_library, "BN_num_bits");
 
     /* Load the functions symbols for OpenSSL EC algorithm. */
+    OSSL_EC_KEY_generate_key = (OSSL_EC_KEY_generate_key_t *)find_crypto_symbol(crypto_library, "EC_KEY_generate_key");
     OSSL_EC_KEY_free = (OSSL_EC_KEY_free_t*)find_crypto_symbol(crypto_library, "EC_KEY_free");
     OSSL_ECDH_compute_key = (OSSL_ECDH_compute_key_t*)find_crypto_symbol(crypto_library, "ECDH_compute_key");
     OSSL_EC_KEY_get0_public_key = (OSSL_EC_KEY_get0_public_key_t*)find_crypto_symbol(crypto_library, "EC_KEY_get0_public_key");
+    OSSL_EC_KEY_get0_private_key = (OSSL_EC_KEY_get0_private_key_t *)find_crypto_symbol(crypto_library, "EC_KEY_get0_private_key");
     OSSL_EC_KEY_new = (OSSL_EC_KEY_new_t*)find_crypto_symbol(crypto_library, "EC_KEY_new");
     OSSL_EC_KEY_set_public_key_affine_coordinates = (OSSL_EC_KEY_set_public_key_affine_coordinates_t*)find_crypto_symbol(crypto_library, "EC_KEY_set_public_key_affine_coordinates");
     OSSL_EC_KEY_set_private_key = (OSSL_EC_KEY_set_private_key_t*)find_crypto_symbol(crypto_library, "EC_KEY_set_private_key");
@@ -473,6 +497,7 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     OSSL_EC_KEY_set_public_key = (OSSL_EC_KEY_set_public_key_t*)find_crypto_symbol(crypto_library, "EC_KEY_set_public_key");
     OSSL_EC_KEY_check_key = (OSSL_EC_KEY_check_key_t*)find_crypto_symbol(crypto_library, "EC_KEY_check_key");
     OSSL_EC_POINT_set_affine_coordinates_GFp = (OSSL_EC_POINT_set_affine_coordinates_GFp_t*)find_crypto_symbol(crypto_library, "EC_POINT_set_affine_coordinates");
+    OSSL_EC_POINT_get_affine_coordinates_GFp = (OSSL_EC_POINT_get_affine_coordinates_GFp_t *)find_crypto_symbol(crypto_library, "EC_POINT_get_affine_coordinates");
     if (NULL == OSSL_EC_KEY_set_public_key_affine_coordinates) {
         /* method missing in OpenSSL version 1.0.0 */
         EC_set_public_key = &setECPublicKey;
@@ -485,6 +510,13 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
         OSSL_EC_POINT_set_affine_coordinates_GF2m = (OSSL_EC_POINT_set_affine_coordinates_GF2m_t*)find_crypto_symbol(crypto_library, "EC_POINT_set_affine_coordinates_GF2m");
     } else {
         OSSL_EC_POINT_set_affine_coordinates_GF2m = (OSSL_EC_POINT_set_affine_coordinates_GF2m_t*)find_crypto_symbol(crypto_library, "EC_POINT_set_affine_coordinates");
+    }
+    if (NULL == OSSL_EC_POINT_get_affine_coordinates_GFp) {
+        /* deprecated in OpenSSL version 1.1.1 */
+        OSSL_EC_POINT_get_affine_coordinates_GFp = (OSSL_EC_POINT_get_affine_coordinates_GFp_t *)find_crypto_symbol(crypto_library, "EC_POINT_get_affine_coordinates_GFp");
+        OSSL_EC_POINT_get_affine_coordinates_GF2m = (OSSL_EC_POINT_get_affine_coordinates_GF2m_t *)find_crypto_symbol(crypto_library, "EC_POINT_get_affine_coordinates_GF2m");
+    } else {
+        OSSL_EC_POINT_get_affine_coordinates_GF2m = (OSSL_EC_POINT_get_affine_coordinates_GF2m_t *)OSSL_EC_POINT_get_affine_coordinates_GFp;
     }
     if ((NULL == OSSL_EC_GROUP_new_curve_GF2m) || (NULL == OSSL_EC_POINT_set_affine_coordinates_GF2m)) {
         /* the OPENSSL_NO_EC2M flag is set and the EC2m methods are unavailable */
@@ -534,12 +566,17 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
         (NULL == OSSL_RSA_free) ||
         (NULL == OSSL_RSA_public_decrypt) ||
         (NULL == OSSL_RSA_private_encrypt) ||
+        (NULL == OSSL_BN_new) ||
         (NULL == OSSL_BN_bin2bn) ||
         (NULL == OSSL_BN_set_negative) ||
         (NULL == OSSL_BN_free) ||
+        (NULL == OSSL_BN_bn2bin) ||
+        (NULL == OSSL_BN_num_bits) ||
+        (NULL == OSSL_EC_KEY_generate_key) ||
         (NULL == OSSL_EC_KEY_free) ||
         (NULL == OSSL_ECDH_compute_key) ||
         (NULL == OSSL_EC_KEY_get0_public_key) ||
+        (NULL == OSSL_EC_KEY_get0_private_key) ||
         (NULL == OSSL_EC_KEY_new) ||
         (NULL == OSSL_EC_KEY_set_private_key) ||
         (NULL == OSSL_BN_CTX_new) ||
@@ -547,6 +584,8 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
         (NULL == OSSL_EC_KEY_set_group) ||
         (NULL == OSSL_EC_POINT_new) ||
         (NULL == OSSL_EC_POINT_set_affine_coordinates_GFp) ||
+        (NULL == OSSL_EC_POINT_get_affine_coordinates_GFp) ||
+        (NULL == OSSL_EC_POINT_get_affine_coordinates_GF2m) ||
         (NULL == OSSL_EC_GROUP_set_generator) ||
         (NULL == OSSL_EC_KEY_get0_group) ||
         (NULL == OSSL_EC_POINT_free) ||
@@ -2306,6 +2345,138 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
     return OSSL_ECGF2M;
 }
 
+static int
+getArrayFromBN(const BIGNUM *bn, unsigned char *out, int len)
+{
+    int ret = -1;
+    int bn_len_bits = (*OSSL_BN_num_bits)(bn);
+    int bn_len = (bn_len_bits + 7) / 8;
+
+    if (bn_len <= len) {
+        int size_diff = len - bn_len;
+        int retLen = (*OSSL_BN_bn2bin)(bn, out + size_diff);
+        if (retLen > 0) {
+            if (size_diff > 0) {
+                memset(out, 0x00, size_diff);
+            }
+            ret = 1;
+        }
+    }
+
+    return ret;
+}
+
+/* Generate an EC Key Pair
+ *
+ * Class:     jdk_crypto_jniprovider_NativeCrypto
+ * Method:    ECGenerateKeyPair
+ * Signature: (J[BI[BI[BII)I
+ */
+JNIEXPORT jint JNICALL
+Java_jdk_crypto_jniprovider_NativeCrypto_ECGenerateKeyPair
+  (JNIEnv *env, jclass obj, jlong key, jbyteArray x, jint xLen, jbyteArray y, jint yLen, jbyteArray s, jint sLen, jint fieldType)
+{
+    jint ret = -1;
+
+    unsigned char *nativeX = NULL;
+    unsigned char *nativeY = NULL;
+    unsigned char *nativeS = NULL;
+    BN_CTX *ctx = NULL;
+    const EC_POINT *publicKey = NULL;
+    const EC_GROUP *publicGroup = NULL;
+    BIGNUM *xBN = (*OSSL_BN_new)();
+    BIGNUM *yBN = (*OSSL_BN_new)();
+    const BIGNUM *sBN = NULL;
+    EC_KEY *nativeKey = (EC_KEY *)(intptr_t) key;
+
+    if (NULL == nativeKey) {
+        goto cleanup;
+    }
+
+    nativeX = (unsigned char *)((*env)->GetPrimitiveArrayCritical(env, x, 0));
+    if (NULL == nativeX) {
+        goto cleanup;
+    }
+
+    nativeY = (unsigned char *)((*env)->GetPrimitiveArrayCritical(env, y, 0));
+    if (NULL == nativeY) {
+        goto cleanup;
+    }
+
+    nativeS = (unsigned char *)((*env)->GetPrimitiveArrayCritical(env, s, 0));
+    if (NULL == nativeS) {
+        goto cleanup;
+    }
+
+    if (0 == (*OSSL_EC_KEY_generate_key)(nativeKey)) {
+        goto cleanup;
+    }
+
+    // to translate the public key to java format, we need to extract the public key coordinates: xBN, yBN
+    ctx = (*OSSL_BN_CTX_new)();
+    if (NULL == ctx) {
+        goto cleanup;
+    }
+
+    publicKey = (*OSSL_EC_KEY_get0_public_key)(nativeKey);
+    publicGroup = (*OSSL_EC_KEY_get0_group)(nativeKey);
+
+    if (jdk_crypto_jniprovider_NativeCrypto_ECField_Fp == fieldType) {
+        if (0 == (*OSSL_EC_POINT_get_affine_coordinates_GFp)(publicGroup, publicKey, xBN, yBN, ctx)) {
+            goto cleanup;
+        }
+    } else {
+        if (0 == (*OSSL_EC_POINT_get_affine_coordinates_GF2m)(publicGroup, publicKey, xBN, yBN, ctx)) {
+            goto cleanup;
+        }
+    }
+
+    ret = getArrayFromBN(xBN, nativeX, xLen);
+    if (ret == -1) {
+        goto cleanup;
+    }
+
+    ret = getArrayFromBN(yBN, nativeY, yLen);
+    if (ret == -1) {
+        goto cleanup;
+    }
+
+    // to translate the private key to java format, we need the private key BIGNUM
+    sBN = (*OSSL_EC_KEY_get0_private_key)(nativeKey);
+
+    ret = getArrayFromBN(sBN, nativeS, sLen);
+    if (ret == -1) {
+        goto cleanup;
+    }
+
+    ret = 1;
+
+cleanup:
+    if (NULL != nativeX) {
+        (*env)->ReleasePrimitiveArrayCritical(env, x, nativeX, JNI_ABORT);
+    }
+    if (NULL != nativeY) {
+        (*env)->ReleasePrimitiveArrayCritical(env, y, nativeY, JNI_ABORT);
+    }
+    if (NULL != nativeS) {
+        (*env)->ReleasePrimitiveArrayCritical(env, s, nativeS, JNI_ABORT);
+    }
+    if (NULL != ctx) {
+        (*OSSL_BN_CTX_free)(ctx);
+    }
+    if (NULL != nativeKey) {
+        (*OSSL_EC_KEY_free)(nativeKey);
+    }
+    if (NULL != xBN) {
+        (*OSSL_BN_free)(xBN);
+    }
+    if (NULL != yBN) {
+        (*OSSL_BN_free)(yBN);
+    }
+
+    return ret;
+}
+
 /* Create an EC Public Key
  *
  * Class:     jdk_crypto_jniprovider_NativeCrypto
@@ -2316,37 +2487,44 @@ JNIEXPORT jint JNICALL
 Java_jdk_crypto_jniprovider_NativeCrypto_ECCreatePublicKey
   (JNIEnv *env, jclass obj, jlong key, jbyteArray x, jint xLen, jbyteArray y, jint yLen, jint field)
 {
+    jint ret = -1;
+
     unsigned char *nativeX = NULL;
     unsigned char *nativeY = NULL;
     EC_KEY *publicKey = (EC_KEY*)(intptr_t) key;
     BIGNUM *xBN = NULL;
     BIGNUM *yBN = NULL;
-    int ret = 0;
 
     nativeX = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, x, 0));
     if (NULL == nativeX) {
-        return -1;
+        goto cleanup;
     }
 
     nativeY = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, y, 0));
     if (NULL == nativeY) {
-        (*env)->ReleasePrimitiveArrayCritical(env, x, nativeX, JNI_ABORT);
-        return -1;
+        goto cleanup;
     }
 
     xBN = convertJavaBItoBN(nativeX, xLen);
     yBN = convertJavaBItoBN(nativeY, yLen);
 
     if ((NULL == xBN) || (NULL == yBN)) {
-        (*env)->ReleasePrimitiveArrayCritical(env, x, nativeX, JNI_ABORT);
-        (*env)->ReleasePrimitiveArrayCritical(env, y, nativeY, JNI_ABORT);
-        return -1;
+        goto cleanup;
     }
 
-    ret = (*EC_set_public_key)(publicKey, xBN, yBN, field);
+    if (0 == (*EC_set_public_key)(publicKey, xBN, yBN, field)) {
+        goto cleanup;
+    }
+    ret = 1;
 
-    (*env)->ReleasePrimitiveArrayCritical(env, x, nativeX, JNI_ABORT);
-    (*env)->ReleasePrimitiveArrayCritical(env, y, nativeY, JNI_ABORT);
+cleanup:
+    if (NULL != nativeX) {
+        (*env)->ReleasePrimitiveArrayCritical(env, x, nativeX, JNI_ABORT);
+    }
+
+    if (NULL != nativeY) {
+        (*env)->ReleasePrimitiveArrayCritical(env, y, nativeY, JNI_ABORT);
+    }
 
     if (NULL != xBN) {
         (*OSSL_BN_free)(xBN);
@@ -2356,11 +2534,7 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECCreatePublicKey
         (*OSSL_BN_free)(yBN);
     }
 
-    if (0 == ret) {
-        return -1;
-    }
-
-    return 0;
+    return ret;
 }
 
 /* Create an EC Private Key
@@ -2373,36 +2547,39 @@ JNIEXPORT jint JNICALL
 Java_jdk_crypto_jniprovider_NativeCrypto_ECCreatePrivateKey
   (JNIEnv *env, jclass obj, jlong key, jbyteArray s, jint sLen)
 {
+    jint ret = -1;
+
     unsigned char *nativeS = NULL;
     EC_KEY *privateKey = (EC_KEY*)(intptr_t) key;
     BIGNUM *sBN = NULL;
-    int ret = 0;
 
     nativeS = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, s, 0));
     if (NULL == nativeS) {
-        return -1;
+        goto cleanup;
     }
 
     sBN = convertJavaBItoBN(nativeS, sLen);
 
     if (NULL == sBN) {
-        (*env)->ReleasePrimitiveArrayCritical(env, s, nativeS, JNI_ABORT);
-        return -1;
+        goto cleanup;
     }
 
-    ret = (*OSSL_EC_KEY_set_private_key)(privateKey, sBN);
+    if (0 == (*OSSL_EC_KEY_set_private_key)(privateKey, sBN)) {
+        goto cleanup;
+    }
 
-    (*env)->ReleasePrimitiveArrayCritical(env, s, nativeS, JNI_ABORT);
+    ret = 1;
+
+cleanup:
+    if (NULL != nativeS) {
+        (*env)->ReleasePrimitiveArrayCritical(env, s, nativeS, JNI_ABORT);
+    }
 
     if (NULL != sBN) {
         (*OSSL_BN_free)(sBN);
     }
 
-    if (0 == ret) {
-        return -1;
-    }
-
-    return 0;
+    return ret;
 }
 
 /* Encode an EC Elliptic Curve over a Prime Field */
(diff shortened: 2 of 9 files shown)


pshipton:
@jasonkatonica pls do the first review of this.

jasonkatonica:
Hi @pshipton , I had worked previously with Kostas in the past few weeks and already reviewed this code and the tests that were run against this code. I was OK with opening a pull request talking to Kostas last week.

This code also intersects and changes some of the same code that is being done for the memory leak at https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/552. The comments and code changes placed into that that PR have been incorporated into this PR although additional updates were required for this feature to work so they are done slightly differently.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Returns the EC curve type.
+     * @param params the parameters of the EC curve
+     * @return the name or OID of the EC curve
+     */
+    static String getCurveName(ECParameterSpec params) {
+        String curveName;
+        if (params instanceof NamedCurve) {
+            curveName = ((NamedCurve) params).getNameAndAliases()[0];
```
keithc-ca, line 107: Patterns:
```java
        if (params instanceof NamedCurve named) {
            curveName = named.getNameAndAliases()[0];
```

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
```diff
@@ -0,0 +1,152 @@
...
+        EllipticCurve curve = params.getCurve();
+        ECField field = curve.getField();
+        byte[] a = curve.getA().toByteArray();
+        byte[] b = curve.getB().toByteArray();
+        byte[] gx = generator.getAffineX().toByteArray();
+        byte[] gy = generator.getAffineY().toByteArray();
+        byte[] n = params.getOrder().toByteArray();
+        byte[] h = BigInteger.valueOf(params.getCofactor()).toByteArray();
+        long nativePointer;
+        int fieldType = 0;
```
keithc-ca, line 137: Please remove redundant initialization.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -60,10 +60,23 @@ public final class SunEC extends Provider {
...
+     * ECKeyGeneration implementation.
+     * OpenSSL 1.1.0 or above is required for EC key generation support.
+     */
+    private static final boolean useNativeECKeyGen = NativeCrypto.isAlgorithmEnabled("jdk.nativeECKeyGen", "SunEC");
+
+    /* The property 'jdk.nativeECDH' is used to control enablement of the native
+     * ECDH implementation.
      */
-    private static final boolean useNativeEC = NativeCrypto.isAlgorithmEnabled("jdk.nativeEC", "SunEC");
+    private static final boolean useNativeECDH = NativeCrypto.isAlgorithmEnabled("jdk.nativeECDH", "SunEC");
```
keithc-ca, line 79: The system property `jdk.nativeEC` must continue to have the same effect.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -161,7 +174,24 @@ public Object newInstance(Object ctrParamObj)
...
+                           (NativeCrypto.getVersionIfAvailable() >= 0x10100000) &&
+                           (!System.getProperty("os.name").contains("AIX"))) {
+                            if (nativeCryptTrace) {
+                                System.err.println("EC KeyPair Generation - Using OpenSSL integration.");
+                            }
+                            return new NativeECKeyPairGenerator();
+                        } else {
+                            if (nativeCryptTrace) {
+                                if (System.getProperty("os.name").contains("AIX")) {
+                                    System.err.println("EC KeyPair Generation - Not using OpenSSL integration in AIX.");
```
keithc-ca, line 188: What's special about AIX?

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -161,7 +174,24 @@ public Object newInstance(Object ctrParamObj)
...
+                                } else if (NativeCrypto.getVersionIfAvailable() < 0x10100000) {
+                                    System.err.println("EC KeyPair Generation - Not using OpenSSL integration due to older version of OpenSSL (<1.1.0).");
```
keithc-ca, line 190: Please use the named constant `NativeCrypto.OPENSSL_VERSION_1_1_0`.

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -449,14 +466,21 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
...
+    /* Load the function symbols for BIGNUM manipulation */
+    OSSL_BN_new = (OSSL_BN_new_t *)find_crypto_symbol(crypto_library, "BN_new");
     OSSL_BN_bin2bn = (OSSL_BN_bin2bn_t *)find_crypto_symbol(crypto_library, "BN_bin2bn");
     OSSL_BN_set_negative = (OSSL_BN_set_negative_t *)find_crypto_symbol(crypto_library, "BN_set_negative");
     OSSL_BN_free = (OSSL_BN_free_t *)find_crypto_symbol(crypto_library, "BN_free");
+    OSSL_BN_bn2bin = (OSSL_BN_bn2bin_t*)find_crypto_symbol(crypto_library, "BN_bn2bin");
+    OSSL_BN_num_bits = (OSSL_BN_num_bits_t*)find_crypto_symbol(crypto_library, "BN_num_bits");
 
     /* Load the functions symbols for OpenSSL EC algorithm. */
+    OSSL_EC_KEY_generate_key = (OSSL_EC_KEY_generate_key_t*)find_crypto_symbol(crypto_library, "EC_KEY_generate_key");
```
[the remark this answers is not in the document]
KostasTsiounis: I followed the format of the rest of the function pointers for EC. Should I, also, change the ones in lines 480-500?

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2344,147 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
...
+    int retLen;
+    int size_diff;
+    unsigned char* intermediate;
+
+    int bn_len_bits = (*OSSL_BN_num_bits)(bn);
+    int bn_len = (bn_len_bits + 7) / 8;
+
+    if (bn_len < len) {
+        intermediate = malloc(sizeof(unsigned char) * bn_len);
+        if (intermediate == NULL) {
```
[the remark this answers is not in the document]
KostasTsiounis: The language does guarantee a return value of 1 for `sizeof(unsigned char)`. I did it for readability reasons, but I can change it if you feel it is more confusing than helpful.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
KostasTsiounis: There was a regression for that functionality in AIX, even with higher OpenSSL versions, so we decided to go with the Java implementation to avoid that.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
```diff
@@ -0,0 +1,152 @@
...
+import sun.security.util.NamedCurve;
+
+import jdk.crypto.jniprovider.NativeCrypto;
+
+/**
+ * Utility methods for the native EC implementation.
+ */
+public class NativeECUtil {
+
+    private static final NativeCrypto nativeCrypto = NativeCrypto.getNativeCrypto();
```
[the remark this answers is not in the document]
KostasTsiounis: Do you mean I should to something like:
`if (nativeCrypto == null) {
   nativeCrypto = NativeCrypto.getNativeCrypto();
}`

Same way as it is in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/539 ?
And if yes, should I do the same for NativeECKeyPairGenerator and NativeECDHKeyAgreement too?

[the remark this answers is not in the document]
keithc-ca: Yes, that object should be lazily accessed like in #539, and in those other places too.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
keithc-ca: Please include a comment explaining that.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2373,40 +2555,43 @@ JNIEXPORT jint JNICALL
...
-        return -1;
-    }
-
-    return 0;
+    return ret;
 }
 
 /* Encode an EC Elliptic Curve over a Prime Field */
-static EC_KEY *
+static EC_KEY*
```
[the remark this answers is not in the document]
keithc-ca: The space makes no difference to the compiler, but IMO the separation helps readers.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
keithc-ca: It's fine to add new properties, but we don't want to break applications that used `-Djdk.nativeEC=false` for whatever reason.

KostasTsiounis: Yeah, sure. I fully understand. I just wanted to have a property that is more descriptive to the user. 

So, should I revert to the previous one or use the "or" option that I mentioned?

keithc-ca: If we wanted to use both properties, they would be combined with `&&`, not `||`, but I don't think it's worthwhile to introduce the new property.

jasonkatonica: It does seem like we could migrate from the use of one flag named `jdk.nativeEC` to another flag named `jdk.nativeECDH` by using a ``||` between the two values like Kostas mentioned above. I would certainly like to have two flags to turn off each feature ( ECDH vs EC key generation ) and hopefully this would allow us to have a descriptive flag using a new naming convention.

keithc-ca: I wouldn't be opposed to a new property to allow disabling ECDH separately from EC in general, but `jdk.nativeEC` currently controls all the OpenSSL integrations and should continue to do so, which means `&&` as I said above.

However, I'm not convinced this added flexibility is worth the complication.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -449,14 +466,21 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     OSSL_RSA_free = (OSSL_RSA_free_t *)find_crypto_symbol(crypto_library, "RSA_free");
     OSSL_RSA_public_decrypt = (OSSL_RSA_public_decrypt_t *)find_crypto_symbol(crypto_library, "RSA_public_decrypt");
     OSSL_RSA_private_encrypt = (OSSL_RSA_private_encrypt_t *)find_crypto_symbol(crypto_library, "RSA_private_decrypt");
+
+    /* Load the function symbols for BIGNUM manipulation */
```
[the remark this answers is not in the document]
KostasTsiounis: Fixed that.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2344,147 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
     return OSSL_ECGF2M;
 }
 
+static int
+getArrayFromBN(const BIGNUM *bn, unsigned char *out, int len)
+{
+    int ret = -1;
+    int retLen;
+    int size_diff;
+    unsigned char* intermediate;
```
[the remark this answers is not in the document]
KostasTsiounis: Initialized local variables and added space.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
KostasTsiounis: Removed `sizeof(unsigned char)` and moved constant on the left side.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2344,147 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
...
+    if (bn_len < len) {
+        intermediate = malloc(sizeof(unsigned char) * bn_len);
+        if (intermediate == NULL) {
+            return 0;
+        }
+        retLen = (*OSSL_BN_bn2bin)(bn, intermediate);
+        size_diff = len - bn_len;
+        for (int i = 0; i < size_diff; i++) {
+            out[i] = 0x00;
+        }
```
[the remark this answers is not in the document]
KostasTsiounis: Changed the code to use `memset()`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2344,147 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
...
+    BN_CTX *ctx = NULL;
+    const EC_POINT *publicKey = NULL;
+    const EC_GROUP *publicGroup = NULL;
+    BIGNUM *xBN = (*OSSL_BN_new)();
+    BIGNUM *yBN = (*OSSL_BN_new)();
+    const BIGNUM *sBN = NULL;
+    EC_KEY *nativeKey = (EC_KEY*)(intptr_t) key;
+
+    if (NULL == nativeKey) {
+        return ret;
```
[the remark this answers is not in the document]
KostasTsiounis: That was actually an error on my part. Thanks for catching it. Made the change.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
KostasTsiounis: Added the space where applicable.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECDHKeyAgreement.java
```diff
@@ -109,36 +109,15 @@ protected void engineInit(Key key, SecureRandom random)
             ECUtil.checkPrivateKey(this.privateKey);
 
             ECParameterSpec params = this.privateKey.getParams();
-            if (params instanceof NamedCurve nc) {
-                this.curve = nc.getNameAndAliases()[0];
+            this.curve = NativeECUtil.getCurveName(params);
+            if ((this.curve != null) && (NativeECUtil.isCurveSupported(this.curve, params))) {
```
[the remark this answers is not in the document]
KostasTsiounis: Removed parentheses.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECDHKeyAgreement.java
```diff
@@ -224,12 +201,10 @@ protected int engineGenerateSecret(byte[] sharedSecret, int offset)
         long nativePublicKey = this.publicKey.getNativePtr();
         long nativePrivateKey = this.privateKey.getNativePtr();
         if ((nativePublicKey == -1) || (nativePrivateKey == -1)) {
-            if (curveSupported.putIfAbsent(this.curve, Boolean.FALSE) != null) {
+            if (!NativeECUtil.putCurveIfAbsent(this.curve, Boolean.FALSE, this.curve +
+                    " is not supported by OpenSSL, using Java crypto implementation for preparing agreement.")) {
```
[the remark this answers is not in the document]
KostasTsiounis: I refactored the whole method to accommodate for this request. If statement and concatenation of Strings was removed and is now happening after getting the result.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
```diff
@@ -0,0 +1,259 @@
...
+import java.security.PublicKey;
+import java.security.SecureRandom;
+import java.security.spec.InvalidParameterSpecException;
+import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.ECGenParameterSpec;
+import java.security.spec.ECParameterSpec;
+import java.security.spec.ECPoint;
+import java.security.spec.ECField;
+import java.security.spec.ECFieldFp;
+import java.security.spec.ECFieldF2m;
```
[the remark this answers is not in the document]
KostasTsiounis: Sorted imports.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
```diff
@@ -0,0 +1,259 @@
...
+    private static final boolean nativeCryptTrace = NativeCrypto.isTraceEnabled();
+
+    /* used to seed the keypair generator */
+    private SecureRandom random;
+
+    /* size of the key to generate, KEY_SIZE_MIN <= keySize <= KEY_SIZE_MAX */
+    private int keySize;
+
+    /* parameters specified via init, if any */
+    private ECParameterSpec params = null;
```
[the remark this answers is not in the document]
KostasTsiounis: Removed redundant initialization.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
```diff
@@ -0,0 +1,259 @@
...
+        BigInteger xBI = new BigInteger(1, x);
+        BigInteger yBI = new BigInteger(1, y);
+        BigInteger sBI = new BigInteger(1, s);
+        ECPoint w = new ECPoint(xBI, yBI);
+        PublicKey publicKey = null;
+        PrivateKey privateKey = null;
+        try {
+            publicKey = new ECPublicKeyImpl(w, this.params);
+        } catch (Exception ex) {
+            ex.printStackTrace();
```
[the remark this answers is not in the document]
KostasTsiounis: Removed printing of stack traces.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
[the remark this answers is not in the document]
KostasTsiounis: Changed all applicable classes to lazily load the native library.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
KostasTsiounis: Changed to use pattern matching.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
KostasTsiounis: Removed redundant initializations.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
KostasTsiounis: I changed it so that the property `idk.nativeEC` maintains the same functionality.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
KostasTsiounis: Changes version check to use constant.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2344,145 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
...
+            return 0;
+        }
+        retLen = (*OSSL_BN_bn2bin)(bn, intermediate);
+        size_diff = len - bn_len;
+        memset(out, 0x00, size_diff);
+        memcpy(out + size_diff, intermediate, (size_t) bn_len);
+        free(intermediate);
+    } else {
+        retLen = (*OSSL_BN_bn2bin)(bn, out);
+    }
```
keithc-ca, line 2370: This should simply fail if `bn_len > len` (rather than write beyond the end of the array.
You should perhaps consider using `BN_bn2binpad()` which applies more directly here.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECDHKeyAgreement.java
```diff
@@ -224,11 +210,14 @@ protected int engineGenerateSecret(byte[] sharedSecret, int offset)
...
+            /* only print the first time a curve is used */
+            if (absent && nativeCryptTrace) {
+                System.err.println(this.curve +
+                        " is not supported by OpenSSL, using Java crypto implementation for preparing agreement.");
             }
-            if (nativeCryptTrace) {
-                System.err.println(this.curve + " is not supported by OpenSSL, using Java crypto implementation.");
+            if (!absent) {
+                throw new ProviderException("Could not convert keys to native format");
             }
```
keithc-ca, line 221: Suggest reordering this (and similar sequences below) so `absent` need only be tested once:
```java
            absent = NativeECUtil.putCurveIfAbsent(this.curve, Boolean.FALSE);
            if (!absent) {
                throw new ProviderException("Could not convert keys to native format");
            }
            /* only print the first time a curve is used */
            if (nativeCryptTrace) {
                System.err.println(this.curve +
                        " is not supported by OpenSSL, using Java crypto implementation for preparing agreement.");
            }
```

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
```diff
@@ -0,0 +1,271 @@
...
+        this.keySize = keySize;
+        this.params = ECUtil.getECParameterSpec(null, keySize);
+        if (this.params == null) {
+            throw new InvalidParameterException(
+                "No EC parameters available for key size " + keySize + " bits");
+        }
+        this.random = random;
+
+        this.curve = NativeECUtil.getCurveName(this.params);
+        if ((this.curve != null) && (NativeECUtil.isCurveSupported(this.curve, this.params))) {
```
keithc-ca, line 117: Please remove redundant parentheses around right argument to `&&`:
```java
        if ((this.curve != null) && NativeECUtil.isCurveSupported(this.curve, this.params)) {
```

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
```diff
@@ -0,0 +1,271 @@
...
+        ECParameterSpec ecSpec = null;
+
+        if (params instanceof ECParameterSpec ecParams) {
+            ecSpec = ECUtil.getECParameterSpec(null, ecParams);
+            if (ecSpec == null) {
+                throw new InvalidAlgorithmParameterException(
+                    "Unsupported curve: " + params);
+            }
+        } else if (params instanceof ECGenParameterSpec) {
+            String name = ((ECGenParameterSpec) params).getName();
```
keithc-ca, line 137: Use pattern-matching here too:
```java
        } else if (params instanceof ECGenParameterSpec ecGenParams) {
            String name = ecGenParams.getName();
```

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
```diff
@@ -0,0 +1,271 @@
...
+            }
+            return this.javaImplementation.generateKeyPair();
+        }
+
+        int coordinatesSize = (params.getCurve().getField().getFieldSize() + 7) >> 3;
+        byte[] x = new byte[coordinatesSize];
+        byte[] y = new byte[coordinatesSize];
+        byte[] s = new byte[coordinatesSize];
+
+        int fieldType;
```
keithc-ca, line 194: Suggest this belongs after line 196.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
```diff
@@ -0,0 +1,156 @@
...
+import java.security.spec.ECGenParameterSpec;
+import java.security.AlgorithmParameters;
+import sun.security.util.NamedCurve;
+
+import jdk.crypto.jniprovider.NativeCrypto;
+
+/**
+ * Utility methods for the native EC implementation.
+ */
+public class NativeECUtil {
```
keithc-ca, line 55: This class should be `final` and have an explicit (private) constructor.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Records whether the specified EC curve is supported by OpenSSL or not,
+     * if the curve is not already associated with a value.
+     * @param curve the EC curve type
+     * @param supported true if the curve is supported by OpenSSL, false otherwise
+     * @param trace the message to print on success, if the native crypto trace is enabled
+     * @return true on success (i.e. the curve was not associated with a value), false otherwise
+     */
+    static boolean putCurveIfAbsent(String curve, Boolean supported) {
+        return (curveSupported.putIfAbsent(curve, supported) == null);
```
keithc-ca, line 97: Please remove outermost parentheses.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
```diff
@@ -0,0 +1,156 @@
...
+        byte[] n = params.getOrder().toByteArray();
+        byte[] h = BigInteger.valueOf(params.getCofactor()).toByteArray();
+        long nativePointer;
+        int fieldType;
+        byte[] p;
+        if (field instanceof ECFieldFp) {
+            p = ((ECFieldFp)field).getP().toByteArray();
+            fieldType = 0;
+        } else if (field instanceof ECFieldF2m) {
+            p = ((ECFieldF2m)field).getReductionPolynomial().toByteArray();
```
keithc-ca, line 144: Pattern matching:
```java
        if (field instanceof ECFieldFp fieldFp) {
            p = fieldFp.getP().toByteArray();
            fieldType = 0;
        } else if (field instanceof ECFieldF2m fieldF2m) {
            p = fieldF2m.getReductionPolynomial().toByteArray();
```

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
```diff
@@ -0,0 +1,156 @@
...
+            fieldType = 1;
+        } else {
+            return -1;
+        }
+        if (nativeCrypto == null) {
+            nativeCrypto = NativeCrypto.getNativeCrypto();
+        }
+        nativePointer = nativeCrypto.ECEncodeGF(fieldType, a, a.length, b, b.length, p, p.length, gx, gx.length, gy, gy.length, n, n.length, h, h.length);
+
+        return nativePointer;
```
keithc-ca, line 154: Suggest removing useless local variable:
```java
        return nativeCrypto.ECEncodeGF(fieldType, a, a.length, b, b.length, p, p.length, gx, gx.length, gy, gy.length, n, n.length, h, h.length);
```

src/jdk.crypto.ec/share/classes/sun/security/ec/ECKeyPairGenerator.java
```diff
@@ -116,7 +122,7 @@ public void initialize(AlgorithmParameterSpec params, SecureRandom random)
         this.random = random;
     }
 
-    private static void ensureCurveIsSupported(ECParameterSpec ecSpec)
+    public static void ensureCurveIsSupported(ECParameterSpec ecSpec)
```
keithc-ca, line 125: This should not be public (package-private will suffice).

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -60,10 +60,23 @@ public final class SunEC extends Provider {
...
+     * OpenSSL 1.1.0 or above is required for EC key generation support.
+     */
+    private static final boolean useNativeECKeyGen = NativeCrypto.isAlgorithmEnabled("jdk.nativeECKeyGen", "SunEC");
+
     /* The property 'jdk.nativeEC' is used to control enablement of the native
-     * EC implementation.
+     * ECDH implementation.
      */
-    private static final boolean useNativeEC = NativeCrypto.isAlgorithmEnabled("jdk.nativeEC", "SunEC");
+    private static final boolean useNativeECDH = NativeCrypto.isAlgorithmEnabled("jdk.nativeEC", "SunEC");
```
keithc-ca, line 79: I think `jdk.nativeEC=false` should disable all native EC paths, so this needs to be reordered (keeping comments):
```java
    private static final boolean useNativeEC = NativeCrypto.isAlgorithmEnabled("jdk.nativeEC", "SunEC");
    private static final boolean useNativeECKeyGen = useNativeEC && NativeCrypto.isAlgorithmEnabled("jdk.nativeECKeyGen", "SunEC");
```
Notice the continued use of `useNativeEC` instead of `useNativeECDH`, which is more appropriate given the property names.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -60,10 +60,23 @@ public final class SunEC extends Provider {
 
     private static final long serialVersionUID = -2279741672933606418L;
 
+    // flag indicating whether the full EC implementation is present
+    // (when native library is absent then fewer EC algorithms are available)
+    private static boolean useFullImplementation = true;
+
+    private static final boolean nativeCryptTrace = NativeCrypto.isTraceEnabled();
+
+
```
keithc-ca, line 69: Just one blank line between declarations.

We'll also want this:
```java
    private static final boolean isAIX = "AIX".equals(System.getProperty("os.name"));
```
to avoid repeatedly fetching that property.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -161,7 +174,25 @@ public Object newInstance(Object ctrParamObj)
...
-                        return new ECKeyPairGenerator();
+                        /* Disabling OpenSSL usage in AIX due to perfomance regression observed. */
+                        if (useNativeECKeyGen &&
+                           (NativeCrypto.getVersionIfAvailable() >= NativeCrypto.OPENSSL_VERSION_1_1_0) &&
+                           (!System.getProperty("os.name").contains("AIX"))) {
+                            if (nativeCryptTrace) {
+                                System.err.println("EC KeyPair Generation - Using OpenSSL integration.");
+                            }
+                            return new NativeECKeyPairGenerator();
+                        } else {
```
keithc-ca, line 185: This yields misleading trace output if `useNativeECKeyGen == false`; suggest:
```java
                        if (useNativeECKeyGen) {
                            if (NativeCrypto.getVersionIfAvailable() < NativeCrypto.OPENSSL_VERSION_1_1_0) {
                                if (nativeCryptTrace) {
                                    System.err.println("EC KeyPair Generation - Not using OpenSSL integration due to older version of OpenSSL (<1.1.0).");
                                }
                            } else if (isAIX) {
                                /* Disabling OpenSSL usage on AIX due to perfomance regression observed. */
                                if (nativeCryptTrace) {
                                    System.err.println("EC KeyPair Generation - Not using OpenSSL integration on AIX.");
                                }
                            } else {
                                if (nativeCryptTrace) {
                                    System.err.println("EC KeyPair Generation - Using OpenSSL integration.");
                                }
                                return new NativeECKeyPairGenerator();
                            }
                        }
                        return new ECKeyPairGenerator();
```

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: `bn_len` is never supposed to be greater than `len`. The only reason it might be smaller is because the struct will strip any zeroes at the beginning of the number. So the else statement is for `bn_len == len`. I could add another segment in case something goes horribly wrong, even though it's not likely since the length parameters are coming from the curve definition.

As far as the `BN_bn2binpad()` method, it was the one I initially used, but it is not available in earlier versions, so I had to write my own padding implementation.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
KostasTsiounis: I was talking to @jasonkatonica about this and he was telling me that it would be better if we had more granularity, regarding the ability to disable features. 

I could have the `useNativeEC` disable everything and then have two separate properties for key generation and key agreement. Would that make more sense?

jasonkatonica: It is a good idea to have separate flags to turn on and off individual algorithms for native code paths instead of one blanket flag to turn off all EC related code paths. This is because `EC` and `ECDH` are different cases in field issue scenarios that we would want to turn off in customer environments. For example, we might want to turn off EC key generation but keep ECDH key exchange enabled.

The code originally wanted to rename the blanket EC algorithm flag ( `jdk.nativeEC` ) to something more specific/ appropriate, but we can certainly live with the `jdk.nativeECKeyGen` referring to EC key generation and then `jdk.nativeEC` referring to `ECDH` key exchange.

It is not worth having one flag to disable all EC algorithms (both ECDH and EC key generation) and also individual flags to turn off features on top of that. This seems a bit harder to manage.

For any users who are currently running with `jdk.nativeEC` set to `false`, we would not necessarily want to turn off this new feature in those environments as they would not benefit from the tested performance improvements for EC key generation.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -486,6 +511,13 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     } else {
         OSSL_EC_POINT_set_affine_coordinates_GF2m = (OSSL_EC_POINT_set_affine_coordinates_GF2m_t*)find_crypto_symbol(crypto_library, "EC_POINT_set_affine_coordinates");
     }
+    if (NULL == OSSL_EC_POINT_get_affine_coordinates_GFp) {
+        /* deprecated in OpenSSL version 1.1.1 */
+        OSSL_EC_POINT_get_affine_coordinates_GFp = (OSSL_EC_POINT_get_affine_coordinates_GFp_t*)find_crypto_symbol(crypto_library, "EC_POINT_get_affine_coordinates_GFp");
+        OSSL_EC_POINT_get_affine_coordinates_GF2m = (OSSL_EC_POINT_get_affine_coordinates_GF2m_t*)find_crypto_symbol(crypto_library, "EC_POINT_get_affine_coordinates_GF2m");
+    } else {
+        OSSL_EC_POINT_get_affine_coordinates_GF2m = (OSSL_EC_POINT_get_affine_coordinates_GF2m_t*)find_crypto_symbol(crypto_library, "EC_POINT_get_affine_coordinates");
```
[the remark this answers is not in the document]
KostasTsiounis: Changed that.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2344,145 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
     return OSSL_ECGF2M;
 }
 
+static int
+getArrayFromBN(const BIGNUM *bn, unsigned char *out, int len)
+{
+    int ret = -1;
+    int retLen = 0;
+    int size_diff = 0;
+    unsigned char *intermediate;
```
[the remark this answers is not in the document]
KostasTsiounis: Moved it to where it is given a value.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2344,145 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
...
+        goto cleanup;
+    }
+
+    // to translate the public key to java format, we need to extract the public key coordinates: xBN, yBN
+    ctx = (*OSSL_BN_CTX_new)();
+    publicKey = (*OSSL_EC_KEY_get0_public_key)(nativeKey);
+    publicGroup = (*OSSL_EC_KEY_get0_group)(nativeKey);
+
+    if (fieldType == 0) {
+        if ((*OSSL_EC_POINT_get_affine_coordinates_GFp)(publicGroup, publicKey, xBN, yBN, ctx) == 0) {
```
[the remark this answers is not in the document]
KostasTsiounis: Put constant on the left.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2844,26 +2903,34 @@ JNIEXPORT jint JNICALL
...
-
-    (*env)->ReleasePrimitiveArrayCritical(env, secret, nativeSecret, 0);
+    publicKeyPoint = (*OSSL_EC_KEY_get0_public_key)(nativePublicKey);
+    if (NULL == publicKeyPoint) {
+        goto cleanup;
+    }
 
-    if (0 == ret) {
-        return -1;
+    if ((*OSSL_ECDH_compute_key)((nativeSecret + secretOffset), secretLen, publicKeyPoint, nativePrivateKey, NULL) == 0) {
```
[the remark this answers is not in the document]
KostasTsiounis: Moved constant and removed parentheses.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
```diff
@@ -0,0 +1,271 @@
...
+            }
+            return this.javaImplementation.generateKeyPair();
+        }
+
+        BigInteger xBI = new BigInteger(1, x);
+        BigInteger yBI = new BigInteger(1, y);
+        BigInteger sBI = new BigInteger(1, s);
+        ECPoint w = new ECPoint(xBI, yBI);
+        PublicKey publicKey = null;
+        PrivateKey privateKey = null;
```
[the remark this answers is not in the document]
KostasTsiounis: Removed.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECDHKeyAgreement.java
KostasTsiounis: Changed it everywhere.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
KostasTsiounis: Actually moved it further down, right before they are used.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
KostasTsiounis: Removed.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
KostasTsiounis: Changed.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
KostasTsiounis: Removed variable.

src/jdk.crypto.ec/share/classes/sun/security/ec/ECKeyPairGenerator.java
KostasTsiounis: Changed to package level.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
KostasTsiounis: Did as agreed and put `jdk.nativeEC` property first.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
KostasTsiounis: Fixed.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
KostasTsiounis: Changed to suggested.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -335,13 +367,21 @@ void putEntries() {
         /*
          *  Key Pair Generator engine
          */
-        putService(new ProviderServiceA(this, "KeyPairGenerator",
-            "EC", "sun.security.ec.ECKeyPairGenerator", ATTRS));
+        /* Disabling OpenSSL usage in AIX due to perfomance regression observed */
+        if (useNativeECKeyGen &&
+            (NativeCrypto.getVersionIfAvailable() >= NativeCrypto.OPENSSL_VERSION_1_1_0)&&
+            (!isAIX)) {
```
keithc-ca, line 373: Put `&&` at the beginning of each continued line, `) {` on a line by itself and remove extra parentheses:
```java
        if (useNativeECKeyGen
            && (NativeCrypto.getVersionIfAvailable() >= NativeCrypto.OPENSSL_VERSION_1_1_0)
            && !isAIX
        ) {
```

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -60,10 +60,25 @@ public final class SunEC extends Provider {
 
     private static final long serialVersionUID = -2279741672933606418L;
 
+    // flag indicating whether the full EC implementation is present
+    // (when native library is absent then fewer EC algorithms are available)
+    private static boolean useFullImplementation = true;
```
keithc-ca, line 65: Please remove this unused field.

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2345,146 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
...
+        free(intermediate);
+    } else if (bn_len == len) {
+        retLen = (*OSSL_BN_bn2bin)(bn, out);
+    } else {
+        /* Do nothing. Since retLen == 0, ret will be returned unchanged. */
+    }
+
+    if (retLen > 0) {
+        ret = 1;
+    }
```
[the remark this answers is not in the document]
KostasTsiounis: Refactored according to proposed way.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2345,146 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
...
+    if (NULL == nativeS) {
+        goto cleanup;
+    }
+
+    if (0 == (*OSSL_EC_KEY_generate_key)(nativeKey)) {
+        goto cleanup;
+    }
+
+    // to translate the public key to java format, we need to extract the public key coordinates: xBN, yBN
+    ctx = (*OSSL_BN_CTX_new)();
```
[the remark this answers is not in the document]
KostasTsiounis: Good catch. Added the check.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2345,146 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
...
+        goto cleanup;
+    }
+
+    ret = getArrayFromBN(yBN, nativeY, yLen);
+    if (ret == -1) {
+        goto cleanup;
+    }
+
+    // to translate the private key to java format, we need the private key BIGNUM
+    sBN = (*OSSL_EC_KEY_get0_private_key)(nativeKey);
```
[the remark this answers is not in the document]
KostasTsiounis: This cannot be `NULL`. The key is already created and contains `sBN`. Method just returns a reference. Creation of key would have failed earlier if there was any issue.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2316,37 +2495,44 @@ JNIEXPORT jint JNICALL
...
 
     if ((NULL == xBN) || (NULL == yBN)) {
-        (*env)->ReleasePrimitiveArrayCritical(env, x, nativeX, JNI_ABORT);
-        (*env)->ReleasePrimitiveArrayCritical(env, y, nativeY, JNI_ABORT);
-        return -1;
+        goto cleanup;
     }
 
-    ret = (*EC_set_public_key)(publicKey, xBN, yBN, field);
+    if ((*EC_set_public_key)(publicKey, xBN, yBN, field) == 0) {
```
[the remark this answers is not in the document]

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2373,36 +2555,39 @@ JNIEXPORT jint JNICALL
...
 
     sBN = convertJavaBItoBN(nativeS, sLen);
 
     if (NULL == sBN) {
-        (*env)->ReleasePrimitiveArrayCritical(env, s, nativeS, JNI_ABORT);
-        return -1;
+        goto cleanup;
+    }
+
+    if ((*OSSL_EC_KEY_set_private_key)(privateKey, sBN) == 0) {
```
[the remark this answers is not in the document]

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2650,141 +2833,20 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECEncodeGFp
...
-        (*env)->ReleasePrimitiveArrayCritical(env, y, nativeY, JNI_ABORT);
-    }
-
-    if (NULL != nativeN) {
-        (*env)->ReleasePrimitiveArrayCritical(env, n, nativeN, JNI_ABORT);
-    }
-
-    if (NULL != nativeH) {
-        (*env)->ReleasePrimitiveArrayCritical(env, h, nativeH, JNI_ABORT);
+    if (fieldType == 0) {
```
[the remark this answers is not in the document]
KostasTsiounis: Done.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECDHKeyAgreement.java
```diff
@@ -109,35 +106,19 @@ protected void engineInit(Key key, SecureRandom random)
...
         } else {
-            if ((curveSupported.putIfAbsent("ECKeyImpl", Boolean.FALSE) == null) && nativeCryptTrace) {
-                System.err.println("Only ECPrivateKeyImpl and ECPublicKeyImpl are supported by the native implementation,"
-                        + " using Java crypto implementation.");
+            boolean absent = NativeECUtil.putCurveIfAbsent("ECKeyImpl", Boolean.FALSE);
+            /* only print the first time a curve is used */
+            if (absent && nativeCryptTrace) {
+                System.err.println("Only ECPrivateKeyImpl and ECPublicKeyImpl" +
+                                   " are supported by the native implementation, " +
+                                   "using Java crypto implementation for key agreement.");
```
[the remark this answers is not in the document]
KostasTsiounis: Done.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
```diff
@@ -0,0 +1,163 @@
...
+        ECPoint generator = params.getGenerator();
+        EllipticCurve curve = params.getCurve();
+        ECField field = curve.getField();
+        byte[] a = curve.getA().toByteArray();
+        byte[] b = curve.getB().toByteArray();
+        byte[] gx = generator.getAffineX().toByteArray();
+        byte[] gy = generator.getAffineY().toByteArray();
+        byte[] n = params.getOrder().toByteArray();
+        byte[] h = BigInteger.valueOf(params.getCofactor()).toByteArray();
+        long nativePointer;
```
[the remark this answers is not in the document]
KostasTsiounis: Done.

src/jdk.crypto.ec/share/classes/sun/security/ec/ECPrivateKeyImpl.java
```diff
@@ -222,29 +214,11 @@ long getNativePtr() {
...
-                        nativePointer = nativeCrypto.ECEncodeGF2m(a, a.length, b, b.length, p, p.length, gx, gx.length, gy, gy.length, n, n.length, h, h.length);
-                    } else {
-                        nativePointer = -1;
-                    }
+                    long nativePointer = NativeECUtil.encodeGroup(this.params);
                     if (nativePointer != -1) {
+                        if (nativeCrypto == null) {
+                            nativeCrypto = NativeCrypto.getNativeCrypto();
+                        }
                         nativeCrypto.createECKeyCleaner(this, nativePointer);
```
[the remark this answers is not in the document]
KostasTsiounis: `ECKeyCleaner` will remove it anyways, but I changed it to what you proposed to avoid keeping unused native handlers for longer than required.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -339,6 +339,15 @@ public final native int ChaCha20FinalDecrypt(long context,
...
 
     /* Native EC interfaces */
+    public final native int ECGenerateKeyPair(long key,
+                                              byte[] x,
+                                              int xLen,
+                                              byte[] y,
+                                              int yLen,
+                                              byte[] s,
+                                              int sLen,
+                                              int fieldType);
```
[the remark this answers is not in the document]
KostasTsiounis: Added constants and updated the references wherever required.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
KostasTsiounis: Done.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
KostasTsiounis: Done.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -115,10 +115,15 @@ typedef int OSSL_RSA_set0_factors_t(RSA *, BIGNUM *, BIGNUM *);
 typedef void OSSL_RSA_free_t (RSA *);
 typedef int OSSL_RSA_public_decrypt_t(int, const unsigned char *, unsigned char *, RSA *, int);
 typedef int OSSL_RSA_private_encrypt_t (int, const unsigned char *, unsigned char *, RSA *, int);
+
+typedef BIGNUM* OSSL_BN_new_t();
```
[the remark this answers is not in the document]
keithc-ca: Not all added or modified lines have been corrected.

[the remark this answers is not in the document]
KostasTsiounis: Not sure what you mean. I went to all the lines I changed and put the space before the `*` instead of after.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2345,141 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
     return OSSL_ECGF2M;
 }
 
+static int
+getArrayFromBN(const BIGNUM *bn, unsigned char *out, int len)
+{
+    int ret = -1;
+    int retLen = 0;
+    int size_diff = 0;
```
keithc-ca, line 2353: Please remove these unused local variables.

src/jdk.crypto.ec/share/classes/sun/security/ec/ECPublicKeyImpl.java
```diff
@@ -157,36 +157,26 @@ long getNativePtr() {
...
-                        if (nativeCrypto.ECCreatePublicKey(nativePointer, x, x.length, y, y.length, fieldType) == -1) {
-                            nativePointer = -1;
+                    long nativePointer = NativeECUtil.encodeGroup(this.params);
+                    try {
+                        if (nativePointer != -1) {
+                            byte[] x = this.w.getAffineX().toByteArray();
+                            byte[] y = this.w.getAffineY().toByteArray();
+                            int fieldType = NativeCrypto.ECField_Fp; // ECFieldFp
+                            if (this.params.getCurve().getField() instanceof ECFieldF2m) {
+                                fieldType = NativeCrypto.ECField_F2m; // ECFieldF2m
```
keithc-ca, line 170: Please remove the comments on lines 168 and 170; they're at best redundant.

keithc-ca (changes requested):
Before merging, I'll want you squash these commits; at that time, I'll ask you to correct the commit message: "dor" should be "for" and the subject line should use the imperative mood
```
Add support for EC key generation and update EC key agreement
```
See more at https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
```diff
@@ -0,0 +1,161 @@
...
+    /* stores whether a curve is supported by OpenSSL (true) or not (false) */
+    private static final Map<String, Boolean> curveSupported = new ConcurrentHashMap<>();
+
+    private NativeECUtil() {}
+
+    /**
+     * Checks whether the given EC curve is not supported by OpenSSL.
+     * @param curve the EC curve type
+     * @param params the parameters of the EC curve
+     * @return false if the curve is not supported, true otherwise.
```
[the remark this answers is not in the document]
KostasTsiounis: Done.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: Done.

src/jdk.crypto.ec/share/classes/sun/security/ec/ECPublicKeyImpl.java
KostasTsiounis: Done.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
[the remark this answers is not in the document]
keithc-ca: The word "not" is still in the first line.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -449,14 +466,21 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
...
     OSSL_BN_free = (OSSL_BN_free_t *)find_crypto_symbol(crypto_library, "BN_free");
+    OSSL_BN_bn2bin = (OSSL_BN_bn2bin_t *)find_crypto_symbol(crypto_library, "BN_bn2bin");
+    OSSL_BN_num_bits = (OSSL_BN_num_bits_t *)find_crypto_symbol(crypto_library, "BN_num_bits");
 
     /* Load the functions symbols for OpenSSL EC algorithm. */
+    OSSL_EC_KEY_generate_key = (OSSL_EC_KEY_generate_key_t *)find_crypto_symbol(crypto_library, "EC_KEY_generate_key");
     OSSL_EC_KEY_free = (OSSL_EC_KEY_free_t*)find_crypto_symbol(crypto_library, "EC_KEY_free");
     OSSL_ECDH_compute_key = (OSSL_ECDH_compute_key_t*)find_crypto_symbol(crypto_library, "ECDH_compute_key");
     OSSL_EC_KEY_get0_public_key = (OSSL_EC_KEY_get0_public_key_t*)find_crypto_symbol(crypto_library, "EC_KEY_get0_public_key");
+    OSSL_EC_KEY_get0_private_key = (OSSL_EC_KEY_get0_private_key_t*)find_crypto_symbol(crypto_library, "EC_KEY_get0_private_key");
```
keithc-ca, line 483: There are still a number of spaces missing before `*`; this is just the first (in new or modified code).

KostasTsiounis: Sorry, I didn't realize I had to change the loading of the functions too. It's done now.

closed/src/jdk.crypto.ec/share/classes/sun/security/ec/NativeECUtil.java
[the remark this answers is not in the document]
KostasTsiounis: Fixed.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2345,138 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
...
+    unsigned char *nativeX = NULL;
+    unsigned char *nativeY = NULL;
+    unsigned char *nativeS = NULL;
+    BN_CTX *ctx = NULL;
+    const EC_POINT *publicKey = NULL;
+    const EC_GROUP *publicGroup = NULL;
+    BIGNUM *xBN = (*OSSL_BN_new)();
+    BIGNUM *yBN = (*OSSL_BN_new)();
+    const BIGNUM *sBN = NULL;
+    EC_KEY *nativeKey = (EC_KEY*)(intptr_t) key;
```
keithc-ca, line 2390: Still more missing spaces, here, lines 2396, 2401, 2406 and line 2904.

KostasTsiounis: Went through all of the pointer declarations and casts in the code I added or modified and changed the space to be before the `*`.

keithc-ca:
Jenkins test sanity alinux64 jdknext

KostasTsiounis:
> Jenkins test sanity alinux64 jdknext

keithc-ca (approved):
None of the test failures appear to be new; most seem to be related to Thread and ThreadGroup changes in Java 21.

pshipton:
Seems this introduced a zlinux build problem https://github.com/eclipse-openj9/openj9/issues/17242

pshipton:
This was reverted by https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/589 due to https://github.com/eclipse-openj9/openj9/issues/17242

PR #567, opened 2023-03-15
Update harfbuzz to not use builtin, which doesn't exist, with AIX xlc

pshipton:
Update hb-algs.hh to avoid using __builtin_mul_overflow


pshipton:
jenkins compile aix jdknext

pshipton:
@backwaterred is looking at the build problem.

pshipton:
Tyler mentioned: The issue with builtins was resolved in version 7.1, and there is an issue open for the second one. I'll see if I can solve this by bumping the harfbuzz version.
For reference: https://github.com/harfbuzz/harfbuzz/issues/4138

keithc-ca:
Closing: This has been fixed upstream.

PR #569, opened 2023-03-16
AIX: don't try to use non-existent __builtin_mul_overflow()

keithc-ca:
Fixes: https://github.com/eclipse-openj9/openj9/issues/16936.

### DIFF
--- a/src/java.desktop/share/native/libharfbuzz/hb-algs.hh
+++ b/src/java.desktop/share/native/libharfbuzz/hb-algs.hh
@@ -26,6 +26,10 @@
  * Facebook Author(s): Behdad Esfahbod
  */
 
+/*
+ * Portions Copyright IBM Corporation, 2023. All Rights Reserved.
+ */
+
 #ifndef HB_ALGS_HH
 #define HB_ALGS_HH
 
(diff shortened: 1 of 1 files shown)


keithc-ca:
Jenkins compile aix,alinux64 jdknext

keithc-ca:
Jenkins compile aix jdknext

pshipton:
@keithc-ca fyi https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/567

keithc-ca:
This was suggested to me as perhaps a more graceful way to deal with missing support for `__builtin_mul_overflow` that would adjust automatically with a compiler upgrade:
```c
#ifndef __has_builtin
#define __has_builtin(X) 0
#endif

#if __has_builtin(__builtin_mul_overflow)
...
```

pshipton:
My point is 
a) we expect to get the buildin fix from OpenJDK via Harfbuzz 7.1.0
b) there are more problems exposed after fixing this that need to be resolved

https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/567#issuecomment-1470954378

src/java.desktop/share/native/libharfbuzz/hb-algs.hh
```diff
@@ -875,7 +879,9 @@ hb_in_ranges (T u, T lo1, T hi1, Ts... ds)
 static inline bool
 hb_unsigned_mul_overflows (unsigned int count, unsigned int size, unsigned *result = nullptr)
 {
-#if (defined(__GNUC__) && (__GNUC__ >= 4)) || (defined(__clang__) && (__clang_major__ >= 8))
+#if defined(AIX)
```
hubert-reinterpretcast, line 882: The built-in is potentially supported by various compilers on AIX such as GCC or the Open XL C/C++ compilers.

keithc-ca:
Closing in favour of #588.

PR #570, opened 2023-03-20
Fix SunTlsRsaPremasterSecret

stephenDiosDev:
- Add code to change mechanism if key spec version is 3.0

This is identical to the SunTlsRsaPremasterSecret fix placed in the zOS repo.

Signed-off-by: Stephen Dios <stephen.dios@ibm.com>

### DIFF
--- a/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/P11TlsRsaPremasterSecretGenerator.java
+++ b/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/P11TlsRsaPremasterSecretGenerator.java
@@ -60,7 +60,7 @@ final class P11TlsRsaPremasterSecretGenerator extends KeyGeneratorSpi {
     private final String algorithm;
 
     // mechanism id
-    private final long mechanism;
+    private long mechanism;
 
     @SuppressWarnings("deprecation")
     private TlsRsaPremasterSecretParameterSpec spec;
--- a/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/P11TlsRsaPremasterSecretGenerator.java
+++ b/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/P11TlsRsaPremasterSecretGenerator.java
@@ -22,6 +22,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
+ * ===========================================================================
+ */
 
 package sun.security.pkcs11;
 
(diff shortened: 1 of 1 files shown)


src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/P11TlsRsaPremasterSecretGenerator.java
```diff
@@ -96,6 +101,13 @@ protected void engineInit(AlgorithmParameterSpec params,
                      " TLS 1.0, TLS 1.1 and TLS 1.2 are supported (" +
                      tlsVersion + ")");
         }
+        // Change mechanism to CKM_TLS_PRE_MASTER_KEY_GEN if version is 3.01, 3.02, 3.03.
+        // Only version 3.00 needs mechanism of CKM_SSL3_PRE_MASTER_KEY_GEN.
+        if (tlsVersion == 0x0300) {
+            mechanism = CKM_SSL3_PRE_MASTER_KEY_GEN;
+        } else {
+            mechanism = CKM_TLS_PRE_MASTER_KEY_GEN;
+        }
```
keithc-ca, line 110: This doesn't compile:
```
error: cannot assign a value to final variable mechanism
```

keithc-ca (changes requested):

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/P11TlsRsaPremasterSecretGenerator.java
keithc-ca: Upon further review, I'm not sure this change is appropriate.

If `tlsVersion == 0x0300` and `mechanism != CKM_SSL3_PRE_MASTER_KEY_GEN`, an exception would have been thrown on line 99, so line 107 doesn't do anything meaningful.

If `tlsVersion != 0x0300` and `mechanism != CKM_TLS_PRE_MASTER_KEY_GEN` perhaps this should throw `InvalidAlgorithmParameterException` as well?

stephenDiosDev: In testing we did see cases where `tlsVersion == 0x0300` and `mechanism != CKM_SSL3_PRE_MASTER_KEY_GEN` and the exception was not thrown (so tlsVersion and mechanism were out of sync), which then caused an error to be thrown by the `C_GenerateKey` call around line 133 because of the mismatched version and mechanism.

keithc-ca: Please explain how the exception doesn't happen when `tlsVersion == 0x0300` and `mechanism != CKM_SSL3_PRE_MASTER_KEY_GEN`. The field `mechanism` was final prior to this change so `!supportSSLv3` is the same condition as `mechanism != CKM_SSL3_PRE_MASTER_KEY_GEN`; put that together with `tlsVersion == 0x0300` and you have a situation that satisfies the test on line 92 in which case the exception is thrown.

stephenDiosDev: My apologies, you're right for this repo since mechanism is marked as `final` the exception would be thrown and this change would not work. Since that's the case, I would argue that this fix doesn't apply to this repo.

keithc-ca:
> this fix doesn't apply to this repo

PR #571, opened 2023-03-21
PKCS#12 file-based keystore

WilburZjh:
This PR aims to provide the PKCS#12 file-based keystore support.

The algorithm that used for keystore operations is called PBE(Password-Based Encryption), this PBE algorithm need to be fips certified. There are two sources of FIPS certified croptography that accepted to Java right now, one is NSS and the other one is openssl. We are providing the support by using a FIPS openssl version, it has the PBE algorithms named PBES2 which is FIPS compliance.

While loading a keystore file and use a fips certified PBE algorithm and now you will have a key in memory. This key will be loaded to do various operations for example TLS handshake. There is one problem that the way NSS works in FIPS mode is they keep their own keys in their own memory. Those keys are not accessable, you cannot read or write those keys from or to NSS. Therefore, we need to import/export the keys to/from NSS.

### DIFF
--- a/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -400,4 +400,15 @@ public class NativeCrypto {
                                       int id,
                                       int hashAlgorithm);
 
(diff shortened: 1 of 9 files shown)


WilburZjh:
This is a draft PR since the extended.jck test is not finished. Once it finished. I will open it.

PR #573, opened 2023-03-22
PKCS#12 file-based Keystore support in FIPS mode

WilburZjh:
This PR aims to provide the PKCS#12 file-based keystore support in FIPS mode.

According to FIPS regulations, the storage of keys and the algorithms used to protect them are only relevant until the keys are moved outside of the cryptographic module. In our case, the keys are moved to a computer, which requires an additional process or code using FIPS key wrapping techniques to remain compliant.

To comply with FIPS regulations while maintaining maximum flexibility and functionality, we plan to utilize non FIPS algorithms for the key storage operations that PKCS12 keystores rely upon. This approach will enable us to achieve compliance without compromising our desired level of flexibility ,and still support various key storage mechanisms like we did in the past in IBM Java 8.

While loading a pkcs12 keystore file, we need to derive a key. This derived key will be stored in memory. This key will be loaded to do various operations for example TLS handshake. There is one problem that the way NSS works in FIPS mode is that NSS keeps its own keys in its own memory. Those keys are not accessable, we cannot read or write those keys from or to NSS. Therefore, we need to import/export the keys to/from NSS.

Currently, we support import/export RSA and EC keys.

### DIFF
--- a/src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
+++ b/src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
@@ -22,6 +22,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * ===========================================================================
+ */
 
 package com.sun.crypto.provider;
 
@@ -42,7 +47,10 @@ import javax.crypto.spec.PBEKeySpec;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
+import jdk.crypto.jniprovider.NativeCrypto;
 import jdk.internal.ref.CleanerFactory;
+
+import openj9.internal.security.RestrictedSecurity;
 
 /**
  * This class represents a PBE key derived using PBKDF2 defined
@@ -117,7 +125,7 @@ final class PBKDF2KeyImpl implements javax.crypto.interfaces.PBEKey {
             } else if (keyLength < 0) {
                 throw new InvalidKeySpecException("Key length is negative");
             }
-            this.prf = Mac.getInstance(prfAlgo, SunJCE.getInstance());
+            this.prf = Mac.getInstance(prfAlgo);
             this.key = deriveKey(prf, passwdBytes, salt, iterCount, keyLength);
         } catch (NoSuchAlgorithmException nsae) {
             // not gonna happen; re-throw just in case
--- a/src/java.base/share/conf/security/java.security
+++ b/src/java.base/share/conf/security/java.security
@@ -137,11 +137,54 @@ RestrictedSecurity1.jce.provider.2 = SUN [{CertificateFactory, X.509, Implemente
     {CertStore, com.sun.security.IndexedCollection, ImplementedIn=Software}, \
     {Policy, JavaPolicy, *}, {Configuration, JavaLoginConfig, *}, \
     {CertPathBuilder, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}, \
-    {CertPathValidator, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}]
+    {CertPathValidator, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}, \
+    {KeyStore, PKCS12, *}]
 RestrictedSecurity1.jce.provider.3 = SunEC [{KeyFactory, EC, ImplementedIn=Software: \
     SupportedKeyClasses=java.security.interfaces.ECPublicKey|java.security.interfaces.ECPrivateKey: \
     KeySize=256}, {AlgorithmParameters, EC, *}]
 RestrictedSecurity1.jce.provider.4 = SunJSSE
+RestrictedSecurity1.jce.provider.5 = SunJCE [{AlgorithmParameters, PBES2, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA1AndAES_128, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA1AndAES_256, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA224AndAES_128, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA224AndAES_256, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA256AndAES_128, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA256AndAES_256, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA384AndAES_128, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA384AndAES_256, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA512AndAES_128, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA512AndAES_256, *}, \
+    {Cipher, PBEWithHmacSHA1AndAES_128, *}, \
+    {Cipher, PBEWithHmacSHA1AndAES_256, *}, \
+    {Cipher, PBEWithHmacSHA224AndAES_128, *}, \
+    {Cipher, PBEWithHmacSHA224AndAES_256, *}, \
+    {Cipher, PBEWithHmacSHA256AndAES_128, *}, \
+    {Cipher, PBEWithHmacSHA256AndAES_256, *}, \
+    {Cipher, PBEWithHmacSHA384AndAES_128, *}, \
+    {Cipher, PBEWithHmacSHA384AndAES_256, *}, \
+    {Cipher, PBEWithHmacSHA512AndAES_128, *}, \
+    {Cipher, PBEWithHmacSHA512AndAES_256, *}, \
+    {Mac, HmacPBESHA1, *}, \
+    {Mac, HmacPBESHA224, *}, \
+    {Mac, HmacPBESHA256, *}, \
+    {Mac, HmacPBESHA384, *}, \
+    {Mac, HmacPBESHA512, *}, \
+    {Mac, PBEWithHmacSHA1, *}, \
+    {Mac, PBEWithHmacSHA224, *}, \
+    {Mac, PBEWithHmacSHA256, *}, \
+    {Mac, PBEWithHmacSHA384, *}, \
+    {Mac, PBEWithHmacSHA512, *}, \
+    {SecretKeyFactory, PBEWithHmacSHA1AndAES_128, *}, \
+    {SecretKeyFactory, PBEWithHmacSHA1AndAES_256, *}, \
+    {SecretKeyFactory, PBEWithHmacSHA224AndAES_128, *}, \
+    {SecretKeyFactory, PBEWithHmacSHA224AndAES_256, *}, \
+    {SecretKeyFactory, PBEWithHmacSHA256AndAES_128, *}, \
+    {SecretKeyFactory, PBEWithHmacSHA256AndAES_256, *}, \
+    {SecretKeyFactory, PBEWithHmacSHA384AndAES_128, *}, \
+    {SecretKeyFactory, PBEWithHmacSHA384AndAES_256, *}, \
+    {SecretKeyFactory, PBEWithHmacSHA512AndAES_128, *}, \
+    {SecretKeyFactory, PBEWithHmacSHA512AndAES_256, *}, \
+    {SecretKeyFactory, PBEWithMD5AndDES, *}]
 
 RestrictedSecurity1.keystore.type = PKCS11
 RestrictedSecurity1.javax.net.ssl.keyStore = NONE
--- a/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
+++ b/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
@@ -33,9 +33,11 @@ package sun.security.pkcs11;
 
 import java.io.*;
 import java.util.*;
+import java.math.BigInteger;
 
 import java.security.*;
 import java.security.interfaces.*;
+import java.security.spec.InvalidKeySpecException;
 import java.util.function.Consumer;
 
 import javax.crypto.BadPaddingException;
@@ -56,7 +58,9 @@ import com.sun.crypto.provider.ChaCha20Poly1305Parameters;
 
 import jdk.internal.misc.InnocuousThread;
 import openj9.internal.security.RestrictedSecurity;
+import sun.security.rsa.RSAUtil.KeyType;
 import sun.security.util.Debug;
+import sun.security.util.ECUtil;
 import sun.security.util.ResourcesMgr;
 import static sun.security.util.SecurityConstants.PROVIDER_VER;
 import static sun.security.util.SecurityProviderConstants.getAliases;
@@ -64,6 +68,7 @@ import static sun.security.util.SecurityProviderConstants.getAliases;
 import sun.security.pkcs11.Secmod.*;
 import sun.security.pkcs11.TemplateManager;
 import sun.security.pkcs11.wrapper.*;
+import sun.security.rsa.RSAPrivateCrtKeyImpl;
 import static sun.security.pkcs11.wrapper.PKCS11Constants.*;
 import static sun.security.pkcs11.wrapper.PKCS11Exception.RV.*;
 
@@ -527,10 +532,17 @@ public final class SunPKCS11 extends AuthProvider {
         }
     }
 
+    private static BigInteger getBigIntegerOrZero(Map<Long, CK_ATTRIBUTE> ckAttrsMap, long attributeType) {
+        CK_ATTRIBUTE attribute = ckAttrsMap.get(attributeType);
+        return (attribute != null) ? attribute.getBigInteger() : BigInteger.ZERO;
+    }
+
     public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exception {
         long keyClass = 0;
         long keyType = 0;
         byte[] keyBytes = null;
+        Map<Long, CK_ATTRIBUTE> ckAttrsMap = new HashMap<>();
+
         // Extract key information.
         for (CK_ATTRIBUTE attr : attributes) {
             if (attr.type == CKA_CLASS) {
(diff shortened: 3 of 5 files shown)


WilburZjh:
@keithc-ca @pshipton @jasonkatonica This PR is a draft for now since extended.jck test has not been finished yet. Once its finished, I will make this PR open.

WilburZjh:
Grinder tests are finished. But we met one problem here. 

OpenSSL has the nomac option while creating a pkcs12 keystore, but while using keytool to create a p12 keystore, it comes up with an exception which is
`Caused by: java.security.NoSuchAlgorithmException: Algorithm HmacPBESHA256 not available`

This is due to keytool has the mac integrity check, but openssl skip it. It seems that all users of the PKCS12KeyStore.engineLoad method will fail to read PKCS12 files in fips mode. Including keytool and any other users of this method. And this is not a keytool problem, the major problem with loading PKCS12 files that have a mac.

We are worried about that other tools will also not have the option to drop the mac like openssl. And this may have conflictions with the liberty existing pkcs12 keystore files.

WilburZjh:
@keithc-ca @pshipton @jasonkatonica PR is ready for review.

keithc-ca:
Please copy the requirements here (as opposed to using an internal URL).

src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
```diff
@@ -22,6 +22,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
```
keithc-ca, line 27: The first year should be 2023.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -33,9 +33,13 @@
 
 import java.io.*;
 import java.util.*;
+import java.math.BigInteger;
 
 import java.security.*;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
```
keithc-ca, line 40: Please remove these unused imports.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -531,6 +538,9 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
         long keyClass = 0;
         long keyType = 0;
         byte[] keyBytes = null;
+        BigInteger tmp = null;
```
keithc-ca, line 541: Please remove this unused local variable.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -539,12 +549,51 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
             if (attr.type == CKA_KEY_TYPE) {
                 keyType = attr.getLong();
             }
-            if (attr.type == CKA_VALUE) {
-                keyBytes = attr.getByteArray();
```
keithc-ca, line 543: Why is this no longer appropriate?

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -539,12 +549,51 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
...
+            ckAttrsMap.put(attr.type, attr);
+        }
+
+        if(keyClass == CKO_PRIVATE_KEY) {
+            if(keyType == CKK_RSA) {
```
keithc-ca, line 556: Formatting: space before `(`.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -539,12 +549,51 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
...
+                        KeyType.RSA,
+                        null,
+                        ckAttrsMap.getOrDefault(CKA_MODULUS, new CK_ATTRIBUTE(CKA_MODULUS, BigInteger.ZERO)).getBigInteger(),
+                        ckAttrsMap.getOrDefault(CKA_PUBLIC_EXPONENT, new CK_ATTRIBUTE(CKA_PUBLIC_EXPONENT, BigInteger.ZERO)).getBigInteger(),
+                        ckAttrsMap.getOrDefault(CKA_PRIVATE_EXPONENT, new CK_ATTRIBUTE(CKA_PRIVATE_EXPONENT, BigInteger.ZERO)).getBigInteger(),
+                        ckAttrsMap.getOrDefault(CKA_PRIME_1, new CK_ATTRIBUTE(CKA_PRIME_1, BigInteger.ZERO)).getBigInteger(),
+                        ckAttrsMap.getOrDefault(CKA_PRIME_2, new CK_ATTRIBUTE(CKA_PRIME_2, BigInteger.ZERO)).getBigInteger(),
+                        ckAttrsMap.getOrDefault(CKA_EXPONENT_1, new CK_ATTRIBUTE(CKA_EXPONENT_1, BigInteger.ZERO)).getBigInteger(),
+                        ckAttrsMap.getOrDefault(CKA_EXPONENT_2, new CK_ATTRIBUTE(CKA_EXPONENT_2, BigInteger.ZERO)).getBigInteger(),
+                        ckAttrsMap.getOrDefault(CKA_COEFFICIENT, new CK_ATTRIBUTE(CKA_COEFFICIENT, BigInteger.ZERO)).getBigInteger()
```
keithc-ca, line 568: Suggest creating a helper method to avoid creating `CK_ATTRIBUTE` objects needlessly (and reducing the code volume).

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -539,12 +549,51 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
...
+                } catch (InvalidKeyException e) {
+                    throw new PKCS11Exception(0x00000005L, null);
+                }
+            } else if (keyType == CKK_EC) {
+                try {
+                    keyBytes = ECUtil.generateECPrivateKey(
+                        ckAttrsMap.getOrDefault(CKA_VALUE, new CK_ATTRIBUTE(CKA_VALUE, BigInteger.ZERO)).getBigInteger(),
+                        ECUtil.getECParameterSpec(
+                            Security.getProvider("SunEC"),
+                            ckAttrsMap.get(CKA_EC_PARAMS).getByteArray())
```
keithc-ca, line 579: What guarantees that `CKA_EC_PARAMS` exists?

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -539,12 +549,51 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
...
+                    // attribute to unwrap it. Otherwise, C_UnwrapKey will produce an Exception.
+                    if (token.config.getNssNetscapeDbWorkaround() && !ckAttrsMap.containsKey(CKA_NETSCAPE_DB)) {
+                        ckAttrsMap.put(CKA_NETSCAPE_DB, new CK_ATTRIBUTE(CKA_NETSCAPE_DB, BigInteger.ZERO));
+                    }
+                } catch (IOException | InvalidKeySpecException e) {
+                    throw new PKCS11Exception(0x00000005L, null);
+                }
             }
+        } else if (keyClass == CKO_SECRET_KEY) {
+            keyBytes = ckAttrsMap.get(CKA_VALUE).getByteArray();
```
keithc-ca, line 591: What guarantees that `CKA_VALUE` exists?

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -539,12 +549,51 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
...
         }
 
-        if ((keyClass == CKO_SECRET_KEY) && (keyBytes != null) && (keyBytes.length > 0)) {
+        if(((keyClass == CKO_SECRET_KEY)
+            || ((keyClass == CKO_PRIVATE_KEY) && ((keyType == CKK_EC) || (keyType == CKK_RSA))))
+            && (keyBytes != null) && (keyBytes.length > 0)) {
```
keithc-ca, line 596: Formatting: space after `if`, closing `) {` on a line by itself.
Suggest testing `keyBytes` first,

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -568,7 +617,9 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
                 byte[] wrappedBytes = wrapCipher.doFinal(keyBytes);
 
                 // Unwrap the secret key.
-                CK_ATTRIBUTE[] unwrapAttributes = token.getAttributes(TemplateManager.O_IMPORT, keyClass, keyType, attributes);
+                // Need additional attributes for EC private key.
+                attributes = new CK_ATTRIBUTE[ckAttrsMap.size()];
+                CK_ATTRIBUTE[] unwrapAttributes = token.getAttributes(TemplateManager.O_IMPORT, keyClass, keyType, ckAttrsMap.values().toArray(attributes));
```
keithc-ca, line 622: Please don't modify parameters; and it's not necessary here:
```java
                CK_ATTRIBUTE[] unwrapAttributes = token.getAttributes(
                        TemplateManager.O_IMPORT,
                        keyClass,
                        keyType,
                        ckAttrsMap.values().toArray(new CK_ATTRIBUTE[ckAttrsMap.size()]));
```

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/wrapper/PKCS11.java
```diff
@@ -55,6 +55,8 @@
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
```
keithc-ca, line 59: Please remove these unused imports.

keithc-ca (changes requested):

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
WilburZjh: That one is used for secret key, we need something else for rsa & ec keys.

src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
WilburZjh: edited

src/java.base/share/conf/security/java.security
```diff
@@ -137,11 +137,54 @@ RestrictedSecurity1.jce.provider.2 = SUN [{CertificateFactory, X.509, Implemente
...
+RestrictedSecurity1.jce.provider.5 = SunJCE [{AlgorithmParameters, PBES2, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA1AndAES_128, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA224AndAES_128, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA256AndAES_128, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA384AndAES_128, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA512AndAES_128, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA1AndAES_256, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA224AndAES_256, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA256AndAES_256, *}, \
+    {AlgorithmParameters, PBEWithHmacSHA384AndAES_256, *}, \
```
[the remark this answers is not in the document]
WilburZjh: sorted

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
WilburZjh: edited

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
WilburZjh: As far as I know, an Elliptic Curve Private Key Object Attributes contains both `CKA_EC_PARAMS` and `CKA_VALUE`. The CKA_EC_PARAMS attribute of a private key which must be provided when the private key object is created.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
WilburZjh: As far as I know, an Generic Secret Key Object Attributes contains both CKA_VALUE and CKA_VALUE_LEN. The CKA_VALUE attribute of a generic secret key must be provided when the secret key object is created.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
WilburZjh: updated

WilburZjh:
Hi @keithc-ca , the codes are updated based on the reviews, and the questions are answered. Please help to review and advise.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -539,12 +552,52 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
...
             }
+        } else if (keyClass == CKO_SECRET_KEY) {
+            keyBytes = ckAttrsMap.get(CKA_VALUE).getByteArray();
         }
 
-        if ((keyClass == CKO_SECRET_KEY) && (keyBytes != null) && (keyBytes.length > 0)) {
+        if ((keyBytes != null) && (keyBytes.length > 0)
+            && ((keyClass == CKO_SECRET_KEY)
+            || ((keyClass == CKO_PRIVATE_KEY) && ((keyType == CKK_EC) || (keyType == CKK_RSA))))
+            ) {
```
keithc-ca, line 600: `||` should be indented one level more, `) {` one level less.

keithc-ca (changes requested):

src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
```diff
@@ -117,7 +125,11 @@ private static byte[] getPasswordBytes(char[] passwd) {
             } else if (keyLength < 0) {
                 throw new InvalidKeySpecException("Key length is negative");
             }
-            this.prf = Mac.getInstance(prfAlgo, SunJCE.getInstance());
+            if (RestrictedSecurity.isFIPSEnabled()) {
+                this.prf = Mac.getInstance(prfAlgo);
+            } else {
+                this.prf = Mac.getInstance(prfAlgo, SunJCE.getInstance());
+            }
```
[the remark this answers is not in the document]
keithc-ca: I think it was an oversight that my proposed change was not made as part of the "restricted security" work.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
keithc-ca: Yes, given this is predicated on `keyClass == CKO_PRIVATE_KEY`, I should have referred to `generatePrivate()` which does provide `CKA_EC_PARAMS`.

Another path that leads here via `C_CreateObject()` is from `P11KeyStore.storePkey()`, but that seems to provide the necessary attributes as well.

This seems rather "delicate" to me: perhaps an explicit check is warranted here that `CKA_EC_PARAMS` is available (throwing a `PKCS11Exception` if not).

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
keithc-ca: I think this should continue to avoid making the assumption that `CKA_VALUE` exists in the list of attributes.
```java
            CK_ATTRIBUTE value = ckAttrsMap.get(CKA_VALUE);
            if (value != null) {
                keyBytes = value.getByteArray();
            }
```

keithc-ca (changes requested):

src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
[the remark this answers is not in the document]
WilburZjh: Ok, so should I remove the provider argument and the RestrictedSecurity check?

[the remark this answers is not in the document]
keithc-ca: Yes, unless you have a good reason not to do so.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -539,12 +551,58 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
...
+                        getBigIntegerOrZero(ckAttrsMap, CKA_EXPONENT_2),
+                        getBigIntegerOrZero(ckAttrsMap, CKA_COEFFICIENT)
+                    ).getEncoded();
+                } catch (InvalidKeyException e) {
+                    throw new PKCS11Exception(0x00000005L, null);
+                }
+            } else if (keyType == CKK_EC) {
+                if (!ckAttrsMap.containsKey(CKA_EC_PARAMS)) {
+                    throw new PKCS11Exception(0x00000005L, "CKA_EC_PARAMS attribute is missing.");
+                }
```
keithc-ca, line 578: Suggest this should just `get()` the attribute earlier and use a `null` check.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -539,12 +551,58 @@ public long importKey(long hSession, CK_ATTRIBUTE[] attributes) throws PKCS11Exc
...
+                        getBigIntegerOrZero(ckAttrsMap, CKA_EXPONENT_1),
+                        getBigIntegerOrZero(ckAttrsMap, CKA_EXPONENT_2),
+                        getBigIntegerOrZero(ckAttrsMap, CKA_COEFFICIENT)
+                    ).getEncoded();
+                } catch (InvalidKeyException e) {
+                    throw new PKCS11Exception(0x00000005L, null);
+                }
+            } else if (keyType == CKK_EC) {
+                if (!ckAttrsMap.containsKey(CKA_EC_PARAMS)) {
+                    throw new PKCS11Exception(0x00000005L, "CKA_EC_PARAMS attribute is missing.");
```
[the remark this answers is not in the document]
WilburZjh: Updated.

src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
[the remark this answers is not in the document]
WilburZjh: Updated.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
keithc-ca: The point of just using `get()` was to avoid two accesses to `ckAttrsMap`; the result of the (first and only) call to `get()` should be captured in a local variable for reuse on line 584.

WilburZjh: Updated.

keithc-ca:
Jenkins test sanity,extended zlinux jdknext

keithc-ca (approved):
None of the test failures seem related to this change.

PR #581, opened 2023-04-06

JasonFengJ9:
Resolved conflicts:
```
```

There is a build issue since https://openj9-jenkins.osuosl.org/job/Build_JDKnext_aarch64_mac_OpenJDK/78/ (Apr 3rd)
* https://github.com/eclipse-openj9/openj9/issues/17121

### DIFF
--- a/src/java.base/share/classes/java/lang/Thread.java
+++ b/src/java.base/share/classes/java/lang/Thread.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1994, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1994, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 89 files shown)


keithc-ca:
This should target the openj9-staging branch. Please adjust accordingly.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1867,7 +1874,7 @@ public final boolean isAlive() {
      * This method is non-final so it can be overridden.
      */
     boolean alive() {
-        return isAlive0();
+        return eetop != 0;
```
keithc-ca, line 1877: For consistency, this should use `NO_REF` instead of `0`.
Please remove the method `isAlive0()` that is no longer used.

keithc-ca (changes requested):

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: Moved `isAlive0()` implementation into `alive()`.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1867,10 +1874,6 @@ public final boolean isAlive() {
      * This method is non-final so it can be overridden.
      */
     boolean alive() {
-        return isAlive0();
-    }
-
-    private boolean isAlive0() {
         synchronized (interruptLock) {
             return eetop != NO_REF;
```
keithc-ca, line 1878: There's no longer any value in using `synrchronized` here now that `eetop` is `volatile`.

JasonFengJ9: Removed synchronization on `interruptLock` for `alive()`.

keithc-ca:
Jenkins compile  zlinux jdknext depends https://github.com/eclipse-openj9/openj9/pull/17126

keithc-ca (approved):
I will wait until https://github.com/eclipse-openj9/openj9/pull/17126 is merged before merging this.

pshipton:
Merging this introduces [crashes in OpenJDK testing](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/581#issuecomment-1505475824). This doesn't need to block merging, but we'll have to figure it out at some point and it might be easier to do it now than a few months later.

keithc-ca:
I reproduced the crashes referenced above; see https://github.com/eclipse-openj9/openj9/issues/17189.

PR #591, opened 2023-04-24
Add support for a suspendOnRestore option in JDWP

mikezhang1234567890:
CRIU supports Java debugger during checkpoint and restore

Added support for a `VM_RESTORE` `JVMTI` event;
Added command line option `suspendOnRestore=[y|n]` to determine if the VM restored from a checkpoint is to be suspended;
`-agentlib:jdwp=transport=dt_socket,server=y,suspend=[y|n]` assumes `suspendOnRestore=y`;
Always respect `suspendOnRestore=n` regardless the order of `suspend` and `suspendOnRestore`, both
`-agentlib:jdwp=transport=dt_socket,server=y,suspendOnRestore=n,suspend=y` and `-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,suspendOnRestore=n` don't suspend the JVM post-restore.

~depends on https://github.com/eclipse-openj9/openj9/pull/19405~ merged

Signed-off-by: Mike Zhang <mike.h.zhang@ibm.com>
Signed-off-by: Jason Feng <fengj@ca.ibm.com>

PR #592, opened 2023-04-25
Recognize OpenJ9 flags in openjdk jcl natives

JasonFengJ9:
Recognize `OpenJ9` flags in `openjdk` `jcl` natives

Copied `j9cfg.h` & `omrcfg.h` into `$(SUPPORT_OUTPUTDIR)/modules_include/java.base`.

related https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/591

### DIFF
--- a/closed/custom/modules/java.base/Copy.gmk
+++ b/closed/custom/modules/java.base/Copy.gmk
@@ -30,6 +30,17 @@ $(call openj9_copy_files,, \
 	$(OPENJ9_VM_BUILD_DIR)/include/jvmti.h \
 	$(INCLUDE_TARGET_DIR)/jvmti.h)
 
+ifeq (true,$(OPENJ9_ENABLE_CMAKE))
+  CONFIG_HEADERS := j9cfg.h omr/omrcfg.h
+else
+  CONFIG_HEADERS := include/j9cfg.h omr/include_core/omrcfg.h
+endif
+
+$(foreach file, $(CONFIG_HEADERS), \
+	$(call openj9_copy_files,, \
+		$(OPENJ9_VM_BUILD_DIR)/$(file) \
+		$(SUPPORT_OUTPUTDIR)/modules_include/java.base/$(notdir $(file))))
+
 ifeq (zos,$(OPENJDK_TARGET_OS))
 $(call openj9_copy_files,, \
 	$(OPENJ9_TOPDIR)/runtime/include/jni_convert.h \


keithc-ca:
Rather than this, when we need to change native code in support of CRIU, we could start with
```c
#include "j9cfg.h"
```
(or something comparable) which would make `J9VM_OPT_CRIU_SUPPORT` (and many other macros) available.

JasonFengJ9:
Added include paths for `#include "j9cfg.h"`.

make/autoconf/flags-cflags.m4
```diff
@@ -441,7 +445,9 @@ AC_DEFUN([FLAGS_SETUP_CFLAGS_HELPER],
       -I${TOPDIR}/src/java.base/share/native/libjava \
       -I${TOPDIR}/src/java.base/$OPENJDK_TARGET_OS_TYPE/native/libjava \
       -I${TOPDIR}/src/hotspot/share/include \
-      -I${TOPDIR}/src/hotspot/os/${HOTSPOT_TARGET_OS_TYPE}/include"
+      -I${TOPDIR}/src/hotspot/os/${HOTSPOT_TARGET_OS_TYPE}/include \
+      -I${OUTPUTDIR}/vm/runtime \
+      -I${OUTPUTDIR}/vm/runtime/omr"
```
keithc-ca, line 450: These paths are incorrect for UMA builds, and there isn't a common prefix that works for both.
Suggest changing `custom-spec.gmk.in` to define `OPENJ9_JDK_INCLUDES` with `OPENJ9_VM_BUILD_DIR`:
```
ifeq (true,$(OPENJ9_ENABLE_CMAKE))
  OPENJ9_VM_BUILD_DIR = $(OUTPUTDIR)/vm/runtime
  OPENJ9_JDK_INCLUDES = \
	-I$(OPENJ9_VM_BUILD_DIR) \
	-I$(OPENJ9_VM_BUILD_DIR)/omr
else
  OPENJ9_VM_BUILD_DIR = $(OUTPUTDIR)/vm
  OPENJ9_JDK_INCLUDES = \
	-I$(OPENJ9_VM_BUILD_DIR)/include \
	-I$(OPENJ9_VM_BUILD_DIR)/omr/include_core
endif
```
Then this can use that macro:
```
    \$(OPENJ9_JDK_INCLUDES)
```

keithc-ca: Actually, another arguably better approach, would be to copy `j9cfg.h` and `omrcfg.h` to `$(SUPPORT_OUTPUTDIR)/modules_include/java.base` which is already on the include path.

JasonFengJ9: Copied `j9cfg.h` & `omrcfg.h` into `$(SUPPORT_OUTPUTDIR)/modules_include/java.base`.

closed/custom/modules/java.base/Copy.gmk
```diff
@@ -30,6 +30,14 @@ $(call openj9_copy_files,, \
 	$(OPENJ9_VM_BUILD_DIR)/include/jvmti.h \
 	$(INCLUDE_TARGET_DIR)/jvmti.h)
 
+$(call openj9_copy_files,, \
+	$(OPENJ9_VM_BUILD_DIR)/j9cfg.h \
+	$(SUPPORT_OUTPUTDIR)/modules_include/java.base/j9cfg.h)
+
+$(call openj9_copy_files,, \
+	$(OPENJ9_VM_BUILD_DIR)/omr/omrcfg.h \
+	$(SUPPORT_OUTPUTDIR)/modules_include/java.base/omrcfg.h)
```
keithc-ca, line 39: These paths are incorrect for UMA builds; see https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/592#discussion_r1178328424.

Have you verified that `java.base-copy` is a prerequisite of the makefile targets for building JDK native code?

keithc-ca (changes requested):

closed/custom/modules/java.base/Copy.gmk
JasonFengJ9: Added a case for UMA builds, and verified it works.

Also checked that modification of `openj9/runtime/include/j9cfg.h.[in|ftl]` is updated in generated `j9cfg.h` and JDK native build, though the target dependency is not clear from the makefile script perspective.

closed/custom/modules/java.base/Copy.gmk
```diff
@@ -30,6 +30,22 @@ $(call openj9_copy_files,, \
...
+else # OPENJ9_ENABLE_CMAKE
+$(call openj9_copy_files,, \
+	$(OPENJ9_VM_BUILD_DIR)/include/j9cfg.h \
+	$(SUPPORT_OUTPUTDIR)/modules_include/java.base/j9cfg.h)
+$(call openj9_copy_files,, \
+	$(OPENJ9_VM_BUILD_DIR)/omr/include_core/omrcfg.h \
+	$(SUPPORT_OUTPUTDIR)/modules_include/java.base/omrcfg.h)
+endif # OPENJ9_ENABLE_CMAKE
```
keithc-ca, line 47: Suggest extracting the differences so this is clearer:
```
ifeq (true,$(OPENJ9_ENABLE_CMAKE))
  CONFIG_HEADERS := j9cfg.h omr/omrcfg.h
else
  CONFIG_HEADERS := include/j9cfg.h omr/include_core/omrcfg.h
endif

$(foreach file, $(CONFIG_HEADERS), \
	$(call openj9_copy_files,, \
		$(OPENJ9_VM_BUILD_DIR)/$(file) \
		$(SUPPORT_OUTPUTDIR)/modules_include/java.base/$(notdir $(file))))
```

JasonFengJ9: Updated.

closed/custom/modules/java.base/Copy.gmk
keithc-ca: I suspect that the necessary dependency is in [`make/Main.gmk`](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/make/Main.gmk#L907):
```
  $(LIBS_TARGETS): $(JVM_MAIN_LIB_TARGETS)
```
This expresses the need to build the VM before building any openjdk native libraries, aided by [`closed/custom/Main.gmk`](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/closed/custom/Main.gmk#L29):
```
JVM_MAIN_LIB_TARGETS := j9vm-build
```

keithc-ca (approved):

keithc-ca:

closed/custom/modules/java.base/Copy.gmk
keithc-ca: The dependencies I mentioned above are not, by themselves, sufficient to guarantee a proper build order.
This also relies on the dependency in `closed/custom/Main.gmk`:
```
java.base-libs : java.base-copy
```
From my perspective, that makes this somewhat fragile, and, unfortunately, that doesn't benefit rules to build libraries for other modules. I think this should be reworked so the header files are copied by `closed/OpenJ9.gmk` and that work is complete by the time the `build-j9` target is finished.

JasonFengJ9: Created
* https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/596

PR #593, opened 2023-04-26
Add support for EC key generation using native OpenSSL library and restructure ECDH key agreement.

KostasTsiounis:
As part of this update, two things happen:

1. A new class is created and equivalent functionality is added that allows the use the native OpenSSL library to perform EC key generation. As part of that, NativeECKeyPairGenerator.java is created and several other classes are amended to support it, namely SunEC to load the appropriate class, and the ECKeyPairGenerator offered by Sun as a fallback option.
2. The code for the ECDH Key Agreement is updated. In that part, helper functions are created in the NativeCrypto.c file to avoid unnecessary passing of EC_KEY pointers. A NativeECUtil.java class is, also, created, so as to bundle commonly used functionality. Lastly, the handling of freeing arrays, keys and BIGNUM structs that were part of that functionality is updated.

### DIFF
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -115,10 +115,15 @@ typedef int OSSL_RSA_set0_factors_t(RSA *, BIGNUM *, BIGNUM *);
 typedef void OSSL_RSA_free_t (RSA *);
 typedef int OSSL_RSA_public_decrypt_t(int, const unsigned char *, unsigned char *, RSA *, int);
 typedef int OSSL_RSA_private_encrypt_t (int, const unsigned char *, unsigned char *, RSA *, int);
-typedef BIGNUM* OSSL_BN_bin2bn_t (const unsigned char *, int, BIGNUM *);
+
+typedef BIGNUM *OSSL_BN_new_t();
+typedef BIGNUM *OSSL_BN_bin2bn_t (const unsigned char *, int, BIGNUM *);
 typedef void OSSL_BN_set_negative_t (BIGNUM *, int);
 typedef void OSSL_BN_free_t (BIGNUM *);
+typedef int OSSL_BN_bn2bin_t(const BIGNUM *, unsigned char *);
+typedef int OSSL_BN_num_bits_t(const BIGNUM *);
 
+typedef int OSSL_EC_KEY_generate_key_t(EC_KEY *);
 typedef void OSSL_EC_KEY_free_t(EC_KEY *);
 typedef int OSSL_ECDH_compute_key_t(void *, size_t, const EC_POINT *, EC_KEY *, void *(*KDF)(const void *, size_t, void *, size_t *));
 typedef const EC_POINT* OSSL_EC_KEY_get0_public_key_t(const EC_KEY *);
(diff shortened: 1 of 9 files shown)


closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -486,7 +511,14 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
...
     }
-    if ((NULL == OSSL_EC_GROUP_new_curve_GF2m) || (NULL == OSSL_EC_POINT_set_affine_coordinates_GF2m)) {
+    if (NULL == OSSL_EC_POINT_get_affine_coordinates_GFp) {
+        /* deprecated in OpenSSL version 1.1.1 */
+        OSSL_EC_POINT_get_affine_coordinates_GFp = (OSSL_EC_POINT_get_affine_coordinates_GFp_t *)find_crypto_symbol(crypto_library, "EC_POINT_get_affine_coordinates_GFp");
+        OSSL_EC_POINT_get_affine_coordinates_GF2m = (OSSL_EC_POINT_get_affine_coordinates_GF2m_t *)find_crypto_symbol(crypto_library, "EC_POINT_get_affine_coordinates_GF2m");
+    } else {
+        OSSL_EC_POINT_get_affine_coordinates_GF2m = (OSSL_EC_POINT_get_affine_coordinates_GF2m_t *)OSSL_EC_POINT_get_affine_coordinates_GFp;
+    }
+    if ((NULL == OSSL_EC_GROUP_new_curve_GF2m) || (NULL == OSSL_EC_POINT_set_affine_coordinates_GF2m) || (NULL == OSSL_EC_POINT_get_affine_coordinates_GF2m)) {
```
keithc-ca, line 521: This line should be folded.
```c
    if ((NULL == OSSL_EC_GROUP_new_curve_GF2m)
    || (NULL == OSSL_EC_POINT_set_affine_coordinates_GF2m)
    || (NULL == OSSL_EC_POINT_get_affine_coordinates_GF2m)
    ) {
```

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -2306,6 +2344,138 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECNativeGF2m
...
+
+    publicKey = (*OSSL_EC_KEY_get0_public_key)(nativeKey);
+    publicGroup = (*OSSL_EC_KEY_get0_group)(nativeKey);
+
+    if (jdk_crypto_jniprovider_NativeCrypto_ECField_Fp == fieldType) {
+        if (0 == (*OSSL_EC_POINT_get_affine_coordinates_GFp)(publicGroup, publicKey, xBN, yBN, ctx)) {
+            goto cleanup;
+        }
+    } else {
+        if (0 == (*OSSL_EC_POINT_get_affine_coordinates_GF2m)(publicGroup, publicKey, xBN, yBN, ctx)) {
```
keithc-ca, line 2428: I think this function should test `OSSL_ECGF2M` (or `OSSL_EC_POINT_get_affine_coordinates_GF2m`) to avoid using the latter when it is `NULL` instead of relying on callers to behave.

KostasTsiounis: The java implementation checks that flag and won't go down to openssl if that is set. I don't think anyone other than the provider will use the method from NativeCrypto.

Exactly the same logic is followed for ECDH, which uses the `OSSL_EC_POINT_set_affine_coordinates_GF2m` that has the same kind of issue.

I could add the check, but I feel like it is obsolete. Let me know what you want me to do.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: Done.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: I actually added the check, just to be on the safe side here.

keithc-ca:
Some of the amac test failures are known: https://github.com/eclipse-openj9/openj9/issues/17280; the other seems unrelated to this change: https://github.com/eclipse-openj9/openj9/issues/17301.

keithc-ca (approved):

keithc-ca:
Similar test results for zlinux.

KostasTsiounis:
FYI @keithc-ca

pshipton:
@KostasTsiounis pls open an issue at https://github.com/eclipse-openj9/openj9-docs/issues/ to document jdk.nativeEC and jdk.nativeECKeyGen. We document the other properties such as [jdk.nativeChaCha20](https://eclipse-openj9.github.io/openj9-docs/djdknativechacha20/), etc.

PR #597, opened 2023-05-04
Get byte array from DerValue without using InputStream

KostasTsiounis:
This fix eliminates the problem of emptying the input stream when reading a DerValue to print for debug purposes.

### DIFF
--- a/src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
+++ b/src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
@@ -138,7 +138,7 @@ public class KDCRep {
                         " req type is " + req_type);
 
                 System.out.println(">>> KDCRep: Message in bytes is =>");
-                byte[] dataBytes = encoding.getDataBytes();
+                byte[] dataBytes = encoding.data().toByteArray();
                 StringBuilder sb = new StringBuilder();
                 for (int i = 0; i < dataBytes.length; i++) {
                     if ((i % 16) == 0) {


KostasTsiounis:
Should I remove the fix statement or do you want me to do the PRs for the rest of the versions now?

src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
```diff
@@ -138,7 +138,7 @@ protected void init(DerValue encoding, int req_type)
                         " req type is " + req_type);
 
                 System.out.println(">>> KDCRep: Message in bytes is =>");
-                byte[] dataBytes = encoding.getDataBytes();
+                byte[] dataBytes = encoding.getData().toByteArray();
```
keithc-ca, line 141: The javadoc for `getData()` says:
> Both this method and the data field should be avoided.
> Consider using data() instead.

Perhaps this should be:
```java
                byte[] dataBytes = encoding.data().toByteArray();
```

KostasTsiounis: The rest of this class is using the `getData()`, so I stayed with it for uniformity reasons. The `data()` is, also, not available in earlier versions of Java (like 11), and creates a new `DerInputStream` for no reason.

I can switch it to `data()`, if you feel we really have to, though.

keithc-ca: If there's no choice but to use `getData()` in earlier versions, that's something we'll just have to accept.

Using `getData()` is appropriate when the intent is to consume parts of the `DerValue` which is not the case here. Please use the recommended method.

KostasTsiounis: Done.

keithc-ca:

pshipton:
I hope jdk_security4 will be re-run after the change.

pshipton:
> Should I remove the fix statement or do you want me to do the PRs for the rest of the versions now?

The fix statement should be removed, from both the commit and the PR description.

keithc-ca:
Test passed in grinder https://openj9-jenkins.osuosl.org/job/Grinder/2312/.

keithc-ca (approved):

src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
```diff
@@ -138,7 +138,7 @@ protected void init(DerValue encoding, int req_type)
                         " req type is " + req_type);
 
                 System.out.println(">>> KDCRep: Message in bytes is =>");
-                byte[] dataBytes = encoding.getDataBytes();
+                byte[] dataBytes = encoding.data().toByteArray();
```
keithc-ca, line 141: As I was reviewing the back-ports of this, I noticed that `toByteArray()` might return `null` (if there are no bytes in encoding). I expect an exception to be thrown by
```java
        der = encoding.getData().getDerValue();
```
on line 159 below, but enabling debugging should replace that with a `NullPointerException` on line 143.

Please address that and update the back-ports similarly.

keithc-ca: I (partially) take that back.; `toByteArray()` never returns `null` in this version, but it might in earlier versions.

KostasTsiounis: The `toByteArray()` will only throw a `NullPointerException` if the buffer is null. But this cannot happen since the constructor of the original `DerValue` would throw an `IOException`.

keithc-ca: That potential problem only applies to Java 11.

PR #612, opened 2023-06-12
Add back the provider argument while getting MAC instance

WilburZjh:
This PR aims to fix the problem mentioned in https://github.com/eclipse-openj9/openj9/issues/17551. While using a third party provider, it seems that we still need to use the SunJCE provider to obtain the MAC instance in non restricted security mode.

### DIFF
--- a/src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
+++ b/src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java
@@ -49,6 +49,8 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 
 import jdk.internal.ref.CleanerFactory;
 
+import openj9.internal.security.RestrictedSecurity;
+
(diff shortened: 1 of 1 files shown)


keithc-ca:
Please update the commit message so the summary line matches the headline here.

keithc-ca:
Please also remove the unused import for `NativeCrypto`.

WilburZjh:
Updated

keithc-ca (approved):

keithc-ca:

pshipton:
There is a new repo https://github.com/ibmruntimes/openj9-openjdk-jdk21 which also needs this change.

WilburZjh:
> There is a new repo https://github.com/ibmruntimes/openj9-openjdk-jdk21 which also needs this change.

Hi @pshipton , PR in jdk21 is opened -> https://github.com/ibmruntimes/openj9-openjdk-jdk21/pull/1

PR #614, opened 2023-06-15
Improve speed of cipher operations for large inputs

KostasTsiounis:
In cases were bytes remain buffered and/or padding is enabled and encryption is performed, the entirety of the input provided in the `doFinal()` methods of the `CipherCore` class would be copied in a buffer with enough space to accommodate the extra bytes requires for buffered bytes and/or padding. The buffer would subsequently be zeroed out for security reasons.

To avoid said actions and improve the performance of the process in case of very large inputs, the updated code calls the `update()` method with as much data as possible before the final steps.

### DIFF
--- a/src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
+++ b/src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
@@ -748,12 +748,26 @@ final class CipherCore {
         throws IllegalBlockSizeException, BadPaddingException {
         try {
             byte[] output = new byte[getOutputSizeByOperation(inputLen, true)];
-            byte[] finalBuf = prepareInputBuffer(input, inputOffset,
-                    inputLen, output, 0);
+            int outputOffset = 0;
+            int outLen = 0;
+
+            if (inputLen > 0) {
+                /*
+                 * Call the update() method to get rid of as many bytes as
+                 * possible before potentially copying array.
+                 */
+                int updateOutLen = update(input, inputOffset, inputLen, output, outputOffset);
+                inputOffset = inputLen;
+                inputLen = 0;
+                outputOffset += updateOutLen;
+                outLen = updateOutLen;
+            }
+
+            byte[] finalBuf = prepareInputBuffer(input, inputOffset, inputLen, output, outputOffset);
             int finalOffset = (finalBuf == input) ? inputOffset : 0;
             int finalBufLen = (finalBuf == input) ? inputLen : finalBuf.length;
 
-            int outLen = fillOutputBuffer(finalBuf, finalOffset, output, 0,
+            outLen += fillOutputBuffer(finalBuf, finalOffset, output, outputOffset,
                     finalBufLen, input);
 
             endDoFinal();
@@ -817,13 +831,6 @@ final class CipherCore {
         int estOutSize = getOutputSizeByOperation(inputLen, true);
         int outputCapacity = checkOutputCapacity(output, outputOffset,
                 estOutSize);
-        int offset = outputOffset;
-        byte[] finalBuf = prepareInputBuffer(input, inputOffset,
-                inputLen, output, outputOffset);
-        byte[] internalOutput = null; // for decrypting only
-
-        int finalOffset = (finalBuf == input) ? inputOffset : 0;
-        int finalBufLen = (finalBuf == input) ? inputLen : finalBuf.length;
 
         if (decrypting) {
             // if the size of specified output buffer is less than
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
```diff
@@ -748,12 +748,26 @@ byte[] doFinal(byte[] input, int inputOffset, int inputLen)
         throws IllegalBlockSizeException, BadPaddingException {
         try {
             byte[] output = new byte[getOutputSizeByOperation(inputLen, true)];
-            byte[] finalBuf = prepareInputBuffer(input, inputOffset,
-                    inputLen, output, 0);
+            int outputOffset = 0;
+            int outLen = 0;
+
+            if (!decrypting && (padding != null) && (inputLen > 0)) {
```
keithc-ca, line 754: Why only when encrypting?

src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
```diff
@@ -748,12 +748,26 @@ byte[] doFinal(byte[] input, int inputOffset, int inputLen)
...
+                /*
+                * Call the update() method to get rid of as many bytes as
+                * possible before potentially copying array to pad.
+                */
+                int updateOutLen = update(input, inputOffset, inputLen, output, outputOffset);
+                inputOffset = inputLen;
+                inputLen = 0;
+                outputOffset += updateOutLen;
+                outLen = updateOutLen;
```
keithc-ca, line 763: `outputOffset` is known to be zero here, and the extra local `updateOutLen` is unnecessary
```java
                outLen = update(input, inputOffset, inputLen, output, 0);
                outputOffset += outLen;
                inputOffset = inputLen;
                inputLen = 0;
```

src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
```diff
@@ -817,9 +831,22 @@ int doFinal(byte[] input, int inputOffset, int inputLen, byte[] output,
...
+        if (!decrypting && (padding != null) && (inputLen > 0)) {
+            /*
+            * Call the update() method to get rid of as many bytes as
+            * possible before potentially copying array to pad.
+            */
+            int updateOutLen = update(input, inputOffset, inputLen, output, outputOffset);
+            inputOffset = inputLen;
+            inputLen = 0;
+            outputOffset += updateOutLen;
+            outLen = updateOutLen;
```
keithc-ca, line 845: Extra local variable `updateOutLen` is not needed:
```java
            outLen = update(input, inputOffset, inputLen, output, outputOffset);
            outputOffset += outLen;
            inputOffset = inputLen;
            inputLen = 0;
```

src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
KostasTsiounis: Because when decrypting is performed, or padding is not required, the input is not copied into an intermediate buffer. In that case, there is no point in calling `update()` and doing an extra trip to the native code with all the issues and delays that might be causing, since there is nothing to be gained.

src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
KostasTsiounis: You are right, but I wrote it that way to make the code more readable. It provides a clearer view into what happens in my opinion.

I could change it, but I feel it's easier to understand with the extra variable. Do you think that's OK, or should I go with your suggestion?

src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
KostasTsiounis: Same as above.

src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
keithc-ca: That's not the way I interpret the condition in `prepareInputBuffer()` following the comment:
```java
        /*
         * prepare the final input, assemble a new buffer if any
         * of the following is true:
         *  - 'input' and 'output' are the same buffer
         *  - there are internally buffered bytes
         *  - doing encryption and padding is needed
         */
```

keithc-ca (approved):
Identical to change in https://github.com/ibmruntimes/openj9-openjdk-jdk20/pull/60.

PR #619, opened 2023-06-21
Enable CRIU support builds for AArch64

0xdaryl:


tajila:
Failures unrelated to criu

keithc-ca:
This should be back-ported to https://github.com/ibmruntimes/openj9-openjdk-jdk21 as well (inserting xr64 in its proper position).

PR #623, opened 2023-06-27
[FFI/Jtreg_JDKnext] Fix the native segment access in test

ChengJin01:
The change is to restore the previous fix in OpenJDK
to deal with the native memory access given 
SymbolLookup.loaderLookup().find() returns a zero-sized
segment for ADDRESS as explained in OpenJDK, in which
case it should be reinterpreted as an valid segment
with the specified size to ensure it works correctly.

### DIFF
--- a/test/jdk/java/foreign/TestClassLoaderFindNative.java
+++ b/test/jdk/java/foreign/TestClassLoaderFindNative.java
@@ -23,7 +23,7 @@
 
 /*
  * ===========================================================================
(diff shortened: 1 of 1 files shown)


ChengJin01:
Reviewer: @keithc-ca
FYI: @tajila, @pshipton

keithc-ca (approved):
Change identical to https://github.com/ibmruntimes/openj9-openjdk-jdk21/pull/11.

PR #624, opened 2023-06-27
[FFI/Jtreg_JDKnext] Remove the promotion code for double on AIX

ChengJin01:
The change is to remove the previous fix of promoting
the float to double on AIX in TestVarArgs.java given
the related code in OpenJDK was updated in which case
the fix is no longer required on AIX.

### DIFF
--- a/test/jdk/java/foreign/TestVarArgs.java
+++ b/test/jdk/java/foreign/TestVarArgs.java
@@ -23,12 +23,6 @@
  */
 
 /*
- * ===========================================================================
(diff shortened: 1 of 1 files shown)


ChengJin01:
Reviewer: @keithc-ca
FYI: @tajila, @pshipton

keithc-ca (approved):

keithc-ca:
Merging: I don't think this needs to wait for https://github.com/eclipse-openj9/openj9/pull/17597.

PR #635, opened 2023-07-24
[FFI/Jtreg_JDKnext] Capture the invalid thread state in upcall

ChengJin01:
The tiny change is to capture the error message from `System.err`
rather than `System.out` to support the related code in OpenJ9
at #eclipse-openj9/openj9/pull/17847 which is different from the RI
implementation.

### DIFF
--- a/test/jdk/java/foreign/trivial/TestTrivialUpcall.java
+++ b/test/jdk/java/foreign/trivial/TestTrivialUpcall.java
@@ -22,6 +22,12 @@
  */
 
 /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * ===========================================================================
+ */
+
+/*
  * @test
  * @enablePreview
  * @library ../ /test/lib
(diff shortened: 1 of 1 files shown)


ChengJin01:
Reviewer: @keithc-ca
FYI: @tajila, @pshipton

test/jdk/java/foreign/trivial/TestTrivialUpcall.java
```diff
@@ -45,7 +51,7 @@ public class TestTrivialUpcall extends UpcallTestHelper {
     @Test
     public void testUpcallFailure() throws IOException, InterruptedException {
         // test to see if we catch a trivial downcall doing an upcall
-        runInNewProcess(Runner.class, true).assertStdOutContains("wrong thread state for upcall");
+        runInNewProcess(Runner.class, true).assertStdErrContains("wrong thread state for upcall");
```
keithc-ca, line 54: This doesn't appear to be specific to OpenJ9: Unless there was a change since Java 20, exception messages are printed to stderr by the RI as well.

ChengJin01: Not really. it crashed the VM in RI with an assertion failure in which case only the message only shows up in stdout`.

keithc-ca: Wow, that seems rather drastic behavior for the RI.

ChengJin01: Yes, that's why we don't think it is appropriate from JVM perspective given the problem comes from the incorrect setting in the user applications.

ChengJin01: See https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/635#discussion_r1272469586.

PR #641, opened 2023-08-10
vm.loadAgentLibrary() and -agentpath: might not load the same agent

JasonFengJ9:
`vm.loadAgentLibrary()` and `-agentpath:` might not load the same agent

`VirtualMachine.loadAgentLibrary("JvmtiAgent1")` and `-agentpath:Path.of(Utils.TEST_NATIVE_PATH, libname1).toAbsolutePath().toString()` might not load the same agent.

related 
* https://github.com/eclipse-openj9/openj9/issues/17500

### DIFF
--- a/test/jdk/com/sun/tools/attach/warnings/DynamicLoadWarningTest.java
+++ b/test/jdk/com/sun/tools/attach/warnings/DynamicLoadWarningTest.java
@@ -22,6 +22,12 @@
  */
 
 /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * ===========================================================================
+ */
+
+/*
  * @test
  * @bug 8307478
  * @summary Test that a warning is printed when an agent is dynamically loaded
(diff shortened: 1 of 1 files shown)


pshipton:
Pls provide more details about why the two things may be different and why this only affects OpenJ9 and not Hotspot.

Also, why not fix loadJvmtiAgent1() rather than effectively copying/modifying the code for only one of the cases.

JasonFengJ9:
[jvmtiStartup.c:findAgentLibrary](https://github.com/eclipse-openj9/openj9/blob/57e96f1f222caff9a05665cc43440cf25a1e063e/runtime/jvmti/jvmtiStartup.c#L1079)
```
	agentLibrary = pool_startDo(jvmtiData->agentLibraries, &state);
	while (agentLibrary) {
        nativeLibrary = &(agentLibrary->nativeLib);
 		nativeLibraryNameLength = strlen(nativeLibrary->name);
		if (nativeLibraryNameLength == libraryLength) {
			if (!strncmp(libraryAndOptions, nativeLibrary->name, libraryLength)) {
				Trc_JVMTI_findAgentLibrary_successExit(nativeLibrary->name);
				return agentLibrary;
			}
		}
		agentLibrary = pool_nextDo(&state);
```
`VirtualMachine.loadAgentLibrary("JvmtiAgent1")` loads the library from one of the library paths, but w/o the actual path in `nativeLibrary->name`.
`-agentpath:Path.of(Utils.TEST_NATIVE_PATH, libname1).toAbsolutePath().toString()` has the absolute path in `nativeLibrary->name`.
OpenJ9 thinks `JvmtiAgent1` is not loaded yet, will load it again.

Not sure about hotspot internals, I renamed a different library to `JvmtiAgent1` and put its absolute path within `-agentpath:`, hotspot thought the library is already loaded and doesn't print the warning message.
In this case, I think the warning message is still required since the `JvmtiAgent1` specified in `-agentpath:` is not the same as the one loading by `VirtualMachine.loadAgentLibrary("JvmtiAgent1")`.

pshipton:
It sounds like a fix to OpenJ9 is required so it can determine the library is already loaded, not a change to the test.

JasonFengJ9:
Not required w/ https://github.com/eclipse-openj9/openj9/pull/17990

PR #643, opened 2023-08-14
Cache openssl message digest contexts

jasonkatonica:
It has been identified that when using `OpenSSL` version `3.x` the message digest context initialization is much more expensive then using `OpenSSL` version `1.x`. This update optimizes the use of `OpenSSL` contexts by making use of a simple cache.

We will create a second initialized `OpenSSL` digest context when the message digest cipher is initialized. This second context, defined as `context->cachedInitalizedDigestContext`, will be copied over the working context, `context->ctx`, using the `EVP_MD_CTX_copy_ex` API whenever we wish to re-initalize this cipher. The restoration of a context occurs whenever a cipher is explicitly reset or whenever a final digest is computed.

Signed-off by: Jason Katonica katonica@us.ibm.com

### DIFF
--- a/closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
+++ b/closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
@@ -24,10 +24,9 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2019 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
  * ===========================================================================
  */
-
 
 package sun.security.provider;
 
@@ -136,7 +135,10 @@ abstract class NativeDigest extends MessageDigestSpi implements Cloneable {
             return;
         }
 
-        nativeCrypto.DigestReset(context);
+        int ret = nativeCrypto.DigestReset(context);
+        if (ret == -1) {
+            throw new ProviderException("Error in Native Digest Reset");
+        }
         bytesProcessed = 0;
     }
 
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -298,6 +298,7 @@ OSSL_PKCS12_key_gen_t* OSSL_PKCS12_key_gen;
 typedef struct OpenSSLMDContext {
     EVP_MD_CTX *ctx;
     const EVP_MD *digestAlg;
+    EVP_MD_CTX *cachedInitializedDigestContext;
 } OpenSSLMDContext;
 
 /* Handle errors from OpenSSL calls. */
@@ -901,22 +902,38 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestCreateCon
     context->ctx = ctx;
     context->digestAlg = digestAlg;
 
+    /*
+     * Create a second initialized openssl digest context. This is being done for performance reasons since
+     * creating and or re-initializing digest contexts later during processing is found to be expensive.
+     * This second context, context->cachedInitializedDigestContext, will be copied over the working context,
+     * context->ctx, using the EVP_MD_CTX_copy_ex API whenever we wish to re-initalize this cipher. This occurs
+     * during an explicit reset of the cipher or whenever a final digest is computed.
+     */
+    context->cachedInitializedDigestContext = (*OSSL_MD_CTX_new)();
+    if (NULL == context->cachedInitializedDigestContext) {
+        goto releaseContexts;
+    }
+
+    if (1 != (*OSSL_MD_CTX_copy_ex)(context->cachedInitializedDigestContext, context->ctx)) {
+        goto releaseContexts;
+    }
+
     if (0 != copyContext) {
         EVP_MD_CTX *contextToCopy = ((OpenSSLMDContext*)(intptr_t)copyContext)->ctx;
         if (NULL == contextToCopy) {
-            (*OSSL_MD_CTX_free)(ctx);
-            free(context);
-            return -1;
+            goto releaseContexts;
         }
         if (0 == (*OSSL_MD_CTX_copy_ex)(ctx, contextToCopy)) {
-            printErrors();
-            (*OSSL_MD_CTX_free)(ctx);
-            free(context);
-            return -1;
+            goto releaseContexts;
         }
     }
 
     return (jlong)(intptr_t)context;
+
+releaseContexts:
+    printErrors();
+    Java_jdk_crypto_jniprovider_NativeCrypto_DigestDestroyContext(env, thisObj, (jlong)(intptr_t)context);
+    return -1;
 }
 
 /*
@@ -928,11 +945,20 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestDestroyCon
   (JNIEnv *env, jclass thisObj, jlong c)
 {
     OpenSSLMDContext *context = (OpenSSLMDContext*)(intptr_t) c;
-    if ((NULL == context) || (NULL == context->ctx)) {
+    if (NULL == context) {
         return -1;
     }
 
-    (*OSSL_MD_CTX_free)(context->ctx);
+    if (NULL != context->ctx) {
+        (*OSSL_MD_CTX_free)(context->ctx);
+        context->ctx = NULL;
+    }
+
+    if (NULL != context->cachedInitializedDigestContext) {
+        (*OSSL_MD_CTX_free)(context->cachedInitializedDigestContext);
+        context->cachedInitializedDigestContext = NULL;
+    }
+
     free(context);
     return 0;
 }
@@ -989,7 +1015,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestComputeAnd
     unsigned char* messageNative = NULL;
     unsigned char* digestNative = NULL;
 
-    if ((NULL == context) || (NULL == context->ctx)) {
+    if ((NULL == context) || (NULL == context->ctx) || (NULL == context->cachedInitializedDigestContext)) {
         return -1;
     }
 
@@ -1021,10 +1047,23 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestComputeAnd
 
     (*env)->ReleasePrimitiveArrayCritical(env, digest, digestNative, 0);
 
-    (*OSSL_MD_CTX_reset)(context->ctx);
-
-    if (1 != (*OSSL_DigestInit_ex)(context->ctx, context->digestAlg, NULL)) {
+    /*
+     * Reset the message digest context to the original context. We are then ready to perform
+     * digest operations again using a copy of this cached context.
+     */
+    if (1 != (*OSSL_MD_CTX_copy_ex)(context->ctx, context->cachedInitializedDigestContext)) {
         printErrors();
+
+        if (NULL != context->ctx) {
+            (*OSSL_MD_CTX_free)(context->ctx);
+            context->ctx = NULL;
+        }
+
+        if (NULL != context->cachedInitializedDigestContext) {
+            (*OSSL_MD_CTX_free)(context->cachedInitializedDigestContext);
+            context->cachedInitializedDigestContext = NULL;
+        }
+
         return -1;
     }
 
@@ -1035,22 +1074,38 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestComputeAnd
  *
  * Class:     jdk_crypto_jniprovider_NativeCrypto
  * Method:    DigestReset
- * Signature: (J)V
+ * Signature: (J)I
  */
-JNIEXPORT void JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestReset
+JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestReset
   (JNIEnv *env, jclass thisObj, jlong c)
 {
     OpenSSLMDContext *context = (OpenSSLMDContext*)(intptr_t) c;
 
-    if ((NULL == context) || (NULL == context->ctx)) {
-        return;
+    if ((NULL == context) || (NULL == context->ctx) || (NULL == context->cachedInitializedDigestContext)) {
+        return -1;
     }
 
-    (*OSSL_MD_CTX_reset)(context->ctx);
-
-    if (1 != (*OSSL_DigestInit_ex)(context->ctx, context->digestAlg, NULL)) {
+    /*
+     * Reset the message digest context to the original context. We are then ready to perform
+     * digest operations again using a copy of this cached context.
+     */
+    if (1 != (*OSSL_MD_CTX_copy_ex)(context->ctx, context->cachedInitializedDigestContext)) {
         printErrors();
+
+        if (NULL != context->ctx) {
+            (*OSSL_MD_CTX_free)(context->ctx);
+            context->ctx = NULL;
+        }
+
+        if (NULL != context->cachedInitializedDigestContext) {
+            (*OSSL_MD_CTX_free)(context->cachedInitializedDigestContext);
+            context->cachedInitializedDigestContext = NULL;
+        }
+
+        return -1;
     }
+
+    return 0;
 }
 
 /*
--- a/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -218,7 +218,7 @@ public class NativeCrypto {
                                                   int digestOffset,
                                                   int digestLen);
 
-    public final native void DigestReset(long context);
+    public final native int DigestReset(long context);
 
     /* Native interfaces shared by CBC and ChaCha20 */
 


pshipton:
This is draft but fyi @keithc-ca for the future review.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -901,6 +902,23 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestCreateCon
...
+    if (NULL == (context->cachedInitalizedDigestContext = (*OSSL_MD_CTX_new)())) {
+        printErrors();
+        Java_jdk_crypto_jniprovider_NativeCrypto_DigestDestroyContext(env, thisObj, (jlong)(intptr_t)context);
+        return -1;
+    }
+
+    if (1 != (*OSSL_DigestInit_ex)(context->cachedInitalizedDigestContext, digestAlg, NULL)) {
+        printErrors();
+        Java_jdk_crypto_jniprovider_NativeCrypto_DigestDestroyContext(env, thisObj, (jlong)(intptr_t)context);
+        return -1;
```
keithc-ca, line 919: Suggest adding a label to the cleanup code (i.e. lines 911-913) to be used here and in the failure cases below (lines 925 and 930).

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1021,10 +1046,11 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestComputeAnd
...
     (*env)->ReleasePrimitiveArrayCritical(env, digest, digestNative, 0);
 
-    (*OSSL_MD_CTX_reset)(context->ctx);
-
-    if (1 != (*OSSL_DigestInit_ex)(context->ctx, context->digestAlg, NULL)) {
+    // Reset the message digest context to the original context. We are then ready to perform
+    // digest operations again using a copy of this cached context.
+    if (1 != (*OSSL_MD_CTX_copy_ex)(context->ctx, context->cachedInitalizedDigestContext)) {
         printErrors();
+        Java_jdk_crypto_jniprovider_NativeCrypto_DigestDestroyContext(env, thisObj, c);
```
keithc-ca, line 1053: Won't this call (and the one on line 1079) potentially lead to releasing the context and the SSL objects two (or more) times? How would the Java code know *not* to destroy the context?

keithc-ca (changes requested):
Please roughly quantify the (relative) costs of creating and initializing versus copying the digest state and provide expectations for how often the state gets reused (if there's no reuse, the changes here will only hinder performance).

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -901,6 +902,23 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestCreateCon
     context->ctx = ctx;
     context->digestAlg = digestAlg;
 
+    // Create a second initalized openssl digest context. This is being done for performance reasons since
+    // creating and or re-initializing digest contexts later during processing is found to be expensive.
+    // This second context, context->cachedInitalizedDigestContext, will be copied over the working context,
+    // context->ctx, using the EVP_MD_CTX_copy_ex API whenever we wish to re-initalize this cipher. This occurs
+    // during an explicit reset of the cipher or whenever a final digest is computed.
```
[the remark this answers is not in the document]
jasonkatonica: Updated comment styles to use /* and */ in this location and others where comments were added.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -901,6 +902,23 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestCreateCon
...
+    if (NULL == (context->cachedInitalizedDigestContext = (*OSSL_MD_CTX_new)())) {
```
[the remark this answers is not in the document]
jasonkatonica: Created separate assignment from the check for NULL on these lines.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
jasonkatonica: Added a label for cleanup called by these lines and the former lines 925-930.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
jasonkatonica: Changed the spelling on this line and in a few other places where I misspelled initialized. Including the variable name `context->cachedInitalizedDigestContext` which was renamed to `context->cachedInitializedDigestContext`

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
jasonkatonica: I added code to nullify `context->ctx` and `context->cachedInitializedDigestContext` right after they are freed to avoid potential double frees for these values.

I think you are right that the `context` itself could potentially be double freed. I was able to prevent the double free of `context->ctx` and `context->cachedInitializedDigestContext` but not the `context` itself. For this reason I ended up changing these lines ( and the former line 1079 ) to just free and nullify `context->ctx` and `context->cachedInitializedDigestContext`, I no longer call the DigestDestroyContext for the exact reason you point out that a double free on the context itself could occur.

The Java code knows when to free these contexts by making use of a [Cleaner](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/Cleaner.html) which is registered right after a context is created successfully [at this location in the NativeDigest.java class](https://github.ibm.com/runtimes/openj9-openjdk-jdk/blob/c38252fc774e577f2b94a50612ba4193ef6ea75f/closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java#L95).  From that point the long value ( which represents a pointer to the context ) is monitored by the Java code cleaner. Once that context becomes unreachable from a Java perspective then the  [DigestDestroyContext method](https://github.ibm.com/runtimes/openj9-openjdk-jdk/blob/c38252fc774e577f2b94a50612ba4193ef6ea75f/closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java#L77) is called by the cleaners run method to cleanup the context by calling the JNI method `Java_jdk_crypto_jniprovider_NativeCrypto_DigestDestroyContext`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
keithc-ca: Our convention is to use ` * ` at the beginning of internal lines:
```c
    /*
     * Create a second initialized openssl digest context. This is being done for performance reasons since
     * creating and or re-initializing digest contexts later during processing is found to be expensive.
     * This second context, context->cachedInitializedDigestContext, will be copied over the working context,
     * context->ctx, using the EVP_MD_CTX_copy_ex API whenever we wish to re-initalize this cipher. This occurs
     * during an explicit reset of the cipher or whenever a final digest is computed.
     */
```
(See also the function comment.)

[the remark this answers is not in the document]
jasonkatonica: Updated comments to include `*` at the start of each comment line.

jasonkatonica:
> Please roughly quantify the (relative) costs of creating and initializing versus copying the digest state and provide expectations for how often the state gets reused (if there's no reuse, the changes here will only hinder performance).

Our micro-benchmarks have shown this code to be performing approximately as follows showing a large increase in performance over the recent milestone 2 `.40` builds with `openssl 3.1` for code that is reusing a cipher within an application.

| Payload  | Threads | Java 11.0.20.0 Milestone 2 | Java 11 Prototype | 
| ------------- | ------------- | ------------- | ------------- | 
| 16  | 1  | 18,566,700 | 34,362,536 |
| 16  | 16  | 1,609,716 | 26,571,358 |
| 512  | 1  | 243,204,432 | 302,936,800 |
| 512  | 16  | 54,432,252 | 250,530,112 |

Applications should typically be reusing a cipher . As per above caching the context shows this is helpful on openssl 3.x. 

I do agree that there might be a case where applications create a Cipher which triggers creation of two contexts, computing a single digest, then throwing away the cipher. Given that the code as it stands today always performs a very expensive EVP_DigestInit_ex at the end of a digest computation.

One thing I could do is instead of initializing two contexts I can initialize one context and then copy it into the "cache" at `context->cachedInitalizedDigestContext`. I can make this update to avoid the double context creation.

jasonkatonica:
> One thing I could do is instead of initializing two contexts I can initialize one context and then copy it into the "cache" at `context->cachedInitalizedDigestContext`. I can make this update to avoid the double context creation.

Just made this update, will need to test this functionally to ensure this will work. If it works functionally I will run various memory leak tools to ensure no memory leaks occur.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -917,6 +935,11 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestCreateCon
     }
 
     return (jlong)(intptr_t)context;
+
+    releaseContexts:
+        printErrors();
+        Java_jdk_crypto_jniprovider_NativeCrypto_DigestDestroyContext(env, thisObj, (jlong)(intptr_t)context);
+        return -1;
```
keithc-ca, line 942: Please fix the indentation here (the label starts at the beginning of the line, other lines should be indented normally.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -901,6 +902,23 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestCreateCon
...
+    context->cachedInitializedDigestContext = (*OSSL_MD_CTX_new)();
+    if (NULL == context->cachedInitializedDigestContext) {
+        goto releaseContexts;
+    }
+
+    if (1 != (*OSSL_MD_CTX_copy_ex)(context->cachedInitializedDigestContext, context->ctx)) {
+        printErrors();
+        goto releaseContexts;
+    }
+
```
keithc-ca, line 921: The failure handling code beginning on lines 925 and 930 doesn't release the `cachedInitializedDigestContext`; suggest `goto releaseContexts;` instead.

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
jasonkatonica: Indentation was fixed.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -901,6 +902,23 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestCreateCon
...
+     * context->ctx, using the EVP_MD_CTX_copy_ex API whenever we wish to re-initalize this cipher. This occurs
+     * during an explicit reset of the cipher or whenever a final digest is computed.
+     */
+    context->cachedInitializedDigestContext = (*OSSL_MD_CTX_new)();
+    if (NULL == context->cachedInitializedDigestContext) {
+        goto releaseContexts;
+    }
+
+    if (1 != (*OSSL_MD_CTX_copy_ex)(context->cachedInitializedDigestContext, context->ctx)) {
+        printErrors();
```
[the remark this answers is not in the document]
jasonkatonica: I removed this duplicate `printErrors()` line.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
jasonkatonica: Agreed its better to call `goto releaseContexts;` so this update was made.

keithc-ca:
I just noticed that your performance comparison was between Java 11 and Java 22: Perhaps it would be worth comparing again given the changes to create the second context via copying, and to eliminate the version difference (comparing Java 22 with and without this change)?

jasonkatonica:
Which build would you suggest that i measure the 22 prototype against? Is there a nightly build that you would prefer?

In general these native code paths ( both Java and C code ) are the same between 11 and 22. I would think its very unlikely we would see any difference in the data here. Additionally we can deduce looking at the raw throughput numbers in the table located at https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/643#issuecomment-1687127475 for the Java 11 prototype and comparing that to the raw data located in the table at https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/643#issuecomment-1688338881 for the Java 22 prototype that the throughput is very similar ( these runs were done on the same physical machine ). This leads me to say that most likely Semeru 22 nightly builds being used as a baseline wont be all too much different performance wise then Java 11 baseline since the Semeru 22 prototype includes most of the Semeru 22 updates and we did not see a difference in raw throughput between the prototypes.

keithc-ca:
I guess I was expecting that you would build the baseline version yourself (in the same way you built the version with these changes). I'm not expecting the difference arising from comparing Java 11 with Java 22 will be huge, but it seems an unnecessary variable to include.

I do think it will be helpful to have performance information that is based on this change as opposed to an earlier form of the proposed change.

jasonkatonica:
OK i will rebuild with and without my changes on java next and try to get this data. I have validated this same / latest code on 11 but will post JDK 22 data here to be sure.

jasonkatonica:
I rebuilt both a baseline and a new prototype. They are both based upon JDK next / 22. I made use of OpenSSL 3.1 in all the below test scenarios for comparison. ( See JCL links below for exact commit that each build was built upon )

Prototype build:

>java -version
openjdk version "22-internal" 2024-03-19
OpenJDK Runtime Environment (build 22-internal-adhoc.jenkins.BuildJDKnextx86-64linuxPersonal)
Eclipse OpenJ9 VM (build master-b03193f9bd4, JRE 22 Linux amd64-64-Bit Compressed References 20230824_1278 (JIT enabled, AOT enabled)
OpenJ9   - b03193f9bd4
OMR      - ce636962eb8
JCL      - f4586b84efd based on jdk-22+10)

Vs 

Custom JDK Next baseline build ( a build based upon the commit just prior to my commits within this PR )

> java -version
openjdk version "22-internal" 2024-03-19
OpenJDK Runtime Environment (build 22-internal-adhoc.jenkins.BuildJDKnextx86-64linuxPersonal)
Eclipse OpenJ9 VM (build master-b03193f9bd4, JRE 22 Linux amd64-64-Bit Compressed References 20230824_1281 (JIT enabled, AOT enabled)
OpenJ9   - b03193f9bd4
OMR      - ce636962eb8
JCL      - 3fd8312629b based on jdk-22+10)

Performance for our micro benchmarks is showing the following:

| Payload  | Thread  | Baseline | Prototype  | 
| ------------- | ------------- | ------------- | ------------- | 
| 16  | 1  | 18,598,312  | 34,188,024  | 
| 512  | 1  | 244,089,680  | 303,138,784  | 
| 16  | 16  | 1,617,685  | 27,027,862 | 
| 512  | 16  | 48,773,344  | 238,811,472  | 

This lines up to a high degree with previous reported data in terms of overall performance gained.

keithc-ca:
> lines up to a high degree with previous reported data

The way I read the previous data, it appeared there was up to a 20% regression in single-threaded performance, however the new data show improvements across the board. I'm inclined to attach more weight to the newer results (with fewer differences in implementations).

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1042,14 +1081,26 @@ JNIEXPORT void JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestReset
...
+
+        if (NULL != context->ctx) {
+            (*OSSL_MD_CTX_free)(context->ctx);
+            context->ctx = NULL;
+        }
+
+        if (NULL != context->cachedInitializedDigestContext) {
+            (*OSSL_MD_CTX_free)(context->cachedInitializedDigestContext);
+            context->cachedInitializedDigestContext = NULL;
+        }
```
keithc-ca, line 1103: This function should have a return value to communicate problems back to Java code (like `DigestComputeAndReset()`).

closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
```diff
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2019 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
```
keithc-ca, line 27: While you're here, could you please delete the extra blank line (one of lines 30 or 31)?

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1037,20 +1076,36 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DigestComputeAnd
  * Method:    DigestReset
  * Signature: (J)V
```
keithc-ca, line 1077: Please update the signature to reflect the fact that this function now returns a value.

keithc-ca (changes requested):

closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
jasonkatonica: Updated in recent commit to remove one of the duplicate blank lines.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
jasonkatonica: Updated `DigestReset` signature to be `Signature: (J)I` to reflect that we are now returning an integer.

keithc-ca:
Please squash and rebase on head of openj9 branch.

jasonkatonica:
Rebased upon head and sqashed into a single commit.

keithc-ca:
Testing for xlinux timed out waiting for machines.
Grinder in progress for alinux64: https://openj9-jenkins.osuosl.org/job/Grinder/2868.

keithc-ca (approved):

PR #644, opened 2023-08-14
Parse -Xmso from OPENJ9_JAVA_OPTIONS for main thread

pshipton:
Issue https://github.com/eclipse-openj9/openj9/issues/16105

### DIFF
--- a/src/java.base/share/native/libjli/args.c
+++ b/src/java.base/share/native/libjli/args.c
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * ===========================================================================
+ */
+
 #include <stdlib.h>
 #include <stdio.h>
 #include <assert.h>
@@ -78,6 +84,7 @@ static jboolean expectingNoDashArg = JNI_FALSE;
 static size_t argsCount = 1;
 static jboolean stopExpansion = JNI_FALSE;
 static jboolean relaunch = JNI_FALSE;
+static jboolean parsingOpenJ9Args = JNI_FALSE;
 
 /*
  * Prototypes for internal functions.
@@ -422,6 +429,10 @@ JLI_PreprocessArg(const char *arg, jboolean expandSourceOpt) {
         return expandArg(arg);
     }
 
+    if (parsingOpenJ9Args && (JLI_StrCCmp(arg, "-Xoptionsfile=") == 0)) {
+        return expandArgFile(arg + 14);
+    }
+
     if (arg[0] != '@') {
         checkArg(arg);
         return NULL;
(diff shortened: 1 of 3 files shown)


src/java.base/share/native/libjli/java.c
```diff
@@ -326,6 +329,41 @@ JLI_Launch(int argc, char ** argv,              /* main argc, argv */
...
+                            threadStackSize = tmp;
+                            if (threadStackSize > 0 && threadStackSize < (jlong)STACK_SIZE_MINIMUM) {
+                                threadStackSize = STACK_SIZE_MINIMUM;
+                            }
+                        }
+                        JLI_MemFree(msoSize);
+                    }
+                    continue; /* str already incremented */
+                }
+                str++;
```
keithc-ca, line 362: I think this needs to be a little more sophisticated (with some of the features of `parseOptionsBuffer()`) to avoid mishandling quotes, etc. and values like this: `-Xmso64k -Dchild.stack=-Xmso32k` (the second occurrence of `-Xmso` should not apply).

pshipton: Updated.

keithc-ca: As discussed offline, this is better, but doesn't handle quotes (and escapes) like the VM. Quotes need not be just at the beginning and end of elements of the value (e.g. `-Xmso64k -Dchild.options="-Xint -Xmso32k -Xms4M"` would not be processed properly).

pshipton: I've updated it to re-use the parsing code in libjli/args.c

src/java.base/share/native/libjli/args.c
```diff
@@ -484,6 +495,43 @@ JLI_AddArgsFromEnvVar(JLI_List args, const char *var_name) {
     return expand(args, env, var_name);
 }
 
+JNIEXPORT jboolean JNICALL
+JLI_ParseOpenJ9ArgsFromEnvVar(JLI_List args, const char *var_name) {
+    char *env = getenv(var_name);
```
keithc-ca, line 500: The type could be `const`:
```c
    const char *env = getenv(var_name);
```

src/java.base/share/native/libjli/args.c
```diff
@@ -484,6 +495,43 @@ JLI_AddArgsFromEnvVar(JLI_List args, const char *var_name) {
...
+    firstAppArgIndex = savedFirstAppArgIndex;
+    expectingNoDashArg = savedExpectingNoDashArg;
+    argsCount = savedArgsCount;
+    stopExpansion = savedStopExpansion;
+    relaunch = savedRelaunch;
+
+    return result;
+}
+
+
```
keithc-ca, line 534: Just one blank line here, please.

src/java.base/share/native/libjli/java.c
```diff
@@ -58,6 +58,7 @@
 
 
 #include <assert.h>
+#include <ctype.h>
```
keithc-ca, line 61: This is no longer needed.

src/java.base/share/native/libjli/java.c
```diff
@@ -326,6 +329,33 @@ JLI_Launch(int argc, char ** argv,              /* main argc, argv */
...
+                        if (threadStackSize > 0 && threadStackSize < (jlong)STACK_SIZE_MINIMUM) {
+                            threadStackSize = STACK_SIZE_MINIMUM;
+                        }
+                    }
+                    foundMso = JNI_TRUE;
+                }
+                JLI_MemFree(openj9Args->elements[i - 1]);
+            }
+            JLI_MemFree(openj9Args->elements);
+            JLI_MemFree(openj9Args);
```
keithc-ca, line 355: Suggestions:
* decrement `i` at the beginning of the loop rather than use `i - 1` several places
* use `JLI_List_free()` instead of essentially repeating its implementation here via `JLI_MemFree()`
* remove `foundMso` and just say `break;` after finding `-Xmso`

```c
            size_t i = openj9Args->size;
            while (i > 0) {
                i -= 1;
                if (JLI_StrCCmp(openj9Args->elements[i], "-Xmso") == 0) {
                    jlong tmp = 0;
                    fprintf(stderr, "Found thread stack size: '%s'\n", openj9Args->elements[i]);
                    if (parse_size(openj9Args->elements[i] + 5, &tmp)) {
                        threadStackSize = tmp;
                        if (threadStackSize > 0 && threadStackSize < (jlong)STACK_SIZE_MINIMUM) {
                            threadStackSize = STACK_SIZE_MINIMUM;
                        }
                    }
                    break;
                }
            }
            JLI_List_free(openj9Args);
```

keithc-ca (changes requested):

pshipton:
I've made the suggested changes.

keithc-ca (approved):

keithc-ca:

PR #650, opened 2023-08-22
Optimize AES/CBC cipher initialization

jasonkatonica:
The EVP cipher initialization cost has been found to be expensive in the `OpenSSL 3.x` API compared to the `OpenSSL 1.x` API.

This update allows for two different types of initializations to occur. The first initialization type is a full initialization which sets the key, iv, and EVP cipher context. The second type of initialization sets just the key and iv and does NOT recreate and reinitialize the EVP context. The former of these two is required once per Cipher instance, the later of these two can be used whenever we are reusing a specific Java cipher object within methods such as `Cipher.doFinal()`.

Signed-off by: Jason Katonica katonica@us.ibm.com

### DIFF
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -1097,7 +1097,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DestroyContext
  */
 JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_CBCInit
   (JNIEnv *env, jclass thisObj, jlong c, jint mode, jbyteArray iv, jint iv_len,
-  jbyteArray key, jint key_len)
+  jbyteArray key, jint key_len, jboolean doReset)
 {
     EVP_CIPHER_CTX *ctx = (EVP_CIPHER_CTX*)(intptr_t) c;
     unsigned char* ivNative = NULL;
@@ -1108,18 +1108,20 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_CBCInit
         return -1;
     }
 
-    switch(key_len) {
-        case 16:
-            evp_cipher1 = (*OSSL_aes_128_cbc)();
-            break;
-        case 24:
-            evp_cipher1 = (*OSSL_aes_192_cbc)();
-            break;
-        case 32:
-            evp_cipher1 = (*OSSL_aes_256_cbc)();
-            break;
-        default:
-            break;
+    if (JNI_FALSE == doReset) {
+        switch (key_len) {
+            case 16:
+                evp_cipher1 = (*OSSL_aes_128_cbc)();
+                break;
+            case 24:
+                evp_cipher1 = (*OSSL_aes_192_cbc)();
+                break;
+            case 32:
+                evp_cipher1 = (*OSSL_aes_256_cbc)();
+                break;
+            default:
+                break;
+        }
     }
 
     ivNative = (unsigned char*)((*env)->GetByteArrayElements(env, iv, 0));
@@ -1140,7 +1142,9 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_CBCInit
         return -1;
     }
 
-    (*OSSL_CIPHER_CTX_set_padding)(ctx, 0);
+    if (JNI_FALSE == doReset) {
+        (*OSSL_CIPHER_CTX_set_padding)(ctx, 0);
+    }
 
     (*env)->ReleaseByteArrayElements(env, iv, (jbyte*)ivNative, JNI_ABORT);
     (*env)->ReleaseByteArrayElements(env, key, (jbyte*)keyNative, JNI_ABORT);
--- a/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeCipherBlockChaining.java
+++ b/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeCipherBlockChaining.java
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
  * ===========================================================================
  */
 
@@ -56,6 +56,7 @@ class NativeCipherBlockChaining extends FeedbackCipher  {
 
     private static final NativeCrypto nativeCrypto;
     private static final Cleaner contextCleaner;
+    private int previousKeyLength = -1;
 
     /*
      * Initialize the CBC context.
@@ -184,7 +185,12 @@ class NativeCipherBlockChaining extends FeedbackCipher  {
 
         int ret;
         synchronized (this) {
-            ret = nativeCrypto.CBCInit(nativeContext, mode, iv, iv.length, key, key.length);
+            if (previousKeyLength == key.length) {
+                ret = nativeCrypto.CBCInit(nativeContext, mode, iv, iv.length, key, key.length, true);
+            } else {
+                ret = nativeCrypto.CBCInit(nativeContext, mode, iv, iv.length, key, key.length, false);
+                previousKeyLength = key.length;
+            }
         }
         if (ret == -1) {
             throw new ProviderException("Error in Native CipherBlockChaining");
@@ -201,7 +207,7 @@ class NativeCipherBlockChaining extends FeedbackCipher  {
         System.arraycopy(iv, 0, r, 0, blockSize);
         int ret;
         synchronized (this) {
-            ret = nativeCrypto.CBCInit(nativeContext, mode, iv, iv.length, key, key.length);
+            ret = nativeCrypto.CBCInit(nativeContext, mode, iv, iv.length, key, key.length, true);
         }
         if (ret == -1) {
             throw new ProviderException("Error in Native CipherBlockChaining");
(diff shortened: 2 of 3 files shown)


closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1097,8 +1097,8 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_DestroyContext
  */
 JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_CBCInit
   (JNIEnv *env, jclass thisObj, jlong c, jint mode, jbyteArray iv, jint iv_len,
-  jbyteArray key, jint key_len)
-{
+  jbyteArray key, jint key_len, jboolean doReset) {
+
```
keithc-ca, line 1101: Please revert this, leaving the opening brace on a line by itself.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1108,18 +1108,22 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_CBCInit
...
-            break;
-        case 32:
-            evp_cipher1 = (*OSSL_aes_256_cbc)();
-            break;
-        default:
-            break;
+    if (JNI_TRUE == doReset) {
+        evp_cipher1 = NULL;
+    } else {
+        switch(key_len) {
```
keithc-ca, line 1114: `evp_cipher1` is already `NULL` here, booleans should only be compared to false (perhaps implicitly).
Please also add the missing space after `switch`.
```c
    if (!doReset) {
        switch (key_len) {
```

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
jasonkatonica: Reverted the location change for the `{` in latest commit.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
jasonkatonica: Made this update to check for JNI_FALSE and not set evp_cipher1 to NULL twice.

keithc-ca:
Please squash.

Do you understand the part of OpenSSL that leads to worse performance in version 3.x as compared with 1.x? If so, have you considered offering a change to OpenSSL that would improve the situation (assuming you think that's possible)?

jasonkatonica:
I just squashed this code into a single commit. 

Yes I do understand some aspects of what is making OpenSSL 3.x more expensive then OpenSSL 1.x. The OpenSSL community is also reporting similar performance issues. The ones that are most interesting  to me so far include:

- https://github.com/openssl/openssl/issues/17064
- https://github.com/openssl/openssl/issues/20625
- https://github.com/openssl/openssl/issues/20171

To summarize the main complaints here: all the various initialization and get/set parameter functions are more expensive when using OpenSSL 3.x. Any code that does not optimally minimize initialization or parameter setting will pay a large performance cost when moving to OpenSSL 3.x. When OpenSSL was rewritten/refactored for version 3.x the APIS associated with crypto were abstracted to provide a much cleaner interface between the crypto engines and the SSL/TLS code. The crypto engines were more of a stand alone crypto engine that has some additional overhead over 1.1.x APIs. For example the majority of the problems we are having include a large amount of locking activity. The majority of the regression we observed was due to this locking activity. I was able to remove the expensive locking code from OpenSSL 3 and build a test library. Semeru when running against this test library ran similar in performance to using OpenSSL 1.1. While this is not an acceptable/proper "fix" to the issue it did show me how much the locking was costing. 

I did consider the idea of trying to change openssl with a real fix but was unable to imagine a good solution that would be acceptable to the community. Given that there have been long standing issues open for quite some time with openssl this did not seem like an easy task.

keithc-ca:
Reading https://github.com/openssl/openssl/issues/17064, it seems the essential change here is to reduce the number of calls to `EVP_aes_NNN_cbc()`. I suggest an even better improvement would result if we simply called each of those functions once in `NativeCrypto.loadCrypto()` capturing the returned `const EVP_CIPHER*` which could be used in `NativeCrypto.CBCInit()`, without needing to change any other files. This would improve all uses of `NativeCrypto.CBCInit()`, including those for new `NativeCipherBlockChaining` instances.

jasonkatonica:
> Reading [openssl/openssl#17064](https://github.com/openssl/openssl/issues/17064), it seems the essential change here is to reduce the number of calls to `EVP_aes_NNN_cbc()`. I suggest an even better improvement would result if we simply called each of those functions once in `NativeCrypto.loadCrypto()` capturing the returned `const EVP_CIPHER*` which could be used in `NativeCrypto.CBCInit()`, without needing to change any other files. This would improve all uses of `NativeCrypto.CBCInit()`, including those for new `NativeCipherBlockChaining` instances.

One of the expensive parts here is calling `EVP_aes_NNN_cbc()` to get a OpenSSL cipher. Most of our testing using callgraphs has shown that the expensive part measured is calling `EVP_CipherInit_ex` to initialize a OpenSSL context with a OpenSSL cipher. It is in this method that we run into locking problems while initializing an OpenSSL context with a OpenSSL cipher. 

Currently we have a 1 to 1 to 1 relationship between OpenSSL contexts, OpenSSL ciphers,  and a Java cipher object. The OpenSSL contexts certainly could not be shared as they contain state data. As far as OpenSSL ciphers, there is a chance we could reuse one of those for the entire JDK.I am not confident that would be a safe thing to do with multiple threads. This would require quite a bit of thinking surrounding testing, switching from one key size to another, multiple threads using this single OpenSSL cipher. I am also concerned that this solution wont be any more performant then what is being done in this pull request for this algorithm.

We could prototype this at a later date to see if it makes any difference and deliver a new change at that time.

jasonkatonica:
Looking into the OpenSSL code some more there are some concerns here that we are unable to cleanup OpenSSL contexts without freeing underlying associated OpenSSL ciphers as well. We certainly think that we need a single OpenSSL context, since state data is stored within this context. If we were to free various OpenSSL contexts, which share a single reference to a OpenSSL cipher, the cipher would be freed as well when we free the OpenSSL context. See line:

https://github.com/openssl/openssl/blob/7a5f58b2cf0d7b2fa0451603a88c3976c657dae9/crypto/evp/evp_enc.c#L47

I don't believe that reusing OpenSSL ciphers allocated within our load functions would be as easy as sharing a single OpenSSL cipher with all OpenSSL contexts.

keithc-ca:
The comment https://github.com/openssl/openssl/issues/17064#issuecomment-973058052 suggests that it is safe to acquire the `const EVP_CIPHER*` pointers and reuse them for multiple cipher objects (and "similarly with digest fetches").

> the expensive part measured is calling `EVP_CipherInit_ex` to initialize a OpenSSL context with a OpenSSL cipher

This change doesn't remove that call, it just passes `NULL` as the `cipher` parameter, which does avoid some work, so I suggest you adopt my suggestion *in addition to* the changes here (as opposed to *instead of* them).

jasonkatonica:
> The comment [openssl/openssl#17064 (comment)](https://github.com/openssl/openssl/issues/17064#issuecomment-973058052) suggests that it is safe to acquire the `const EVP_CIPHER*` pointers and reuse them for multiple cipher objects (and "similarly with digest fetches").

In this example the user is only attempting to reuse a single OpenSSL cipher instance. Given that the code is a snippet we are unsure if he is making use of a single or multiple OpenSSL contexts using that single OpenSSL cipher. Additionally this code does not address freeing the context and the worries we mentioned [in the comment above](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/650#issuecomment-1693567771).

The user also did not find that this helped his performance problems, since he ran into the same locking issues that our team did using `EVP_CipherInit_ex`. He mentions this a few comments below in the issue:

https://github.com/openssl/openssl/issues/17064#issuecomment-974066092

It is the passing of an OpenSSL cipher to the `EVP_CipherInit_ex` method that makes that call expensive. We dont believe that we can entirely remove the calls to `EVP_CipherInit_ex` when either reusing a cipher or when using a cipher for the first time. Passing a NULL as is coded here only saves on the more expensive portion of the initialization of the OpenSSL cipher ( skips the locking and fetching of a new OpenSSL cipher ) that we have measured.

keithc-ca:
Jenkins test sanity.openjdk xlinux jdknext

jasonkatonica:
I was investigating this issue some more and started asking bigger picture questions in regard to how these Cipher are typically used. Typically applications should be instantiating Ciphers a minimal amounts of times possible. They are then reused by either reinitializing them or resetting them. 

Prior to this update we created one OpenSSL cipher for each Cipher instantiation AND created a new OpenSSL cipher for each Cipher reuse. This could add up to a substantial amount of OpenSSL ciphers being created, initialized, and destroyed.

By implementing the above update we were able to narrow down the number of OpenSSL ciphers being created to only a handful of OpenSSL ciphers. The typical use pattern with this update dramatically create, initalizes, and destroys less OpenSSL ciphers.

I acknowledge that there is additional value in extending this approach and attempting to create and share a single OpenSSL cipher for an entire application but I do see risk in doing this along with predicted performance gained.

keithc-ca:
Jenkins compile win jdknext

keithc-ca:
Testing depends on resolution of these issues (so the required `jtreg_7_3_1_1.tar.gz` is available):
* https://github.com/adoptium/ci-jenkins-pipelines/issues/794
* https://github.com/adoptium/ci-jenkins-pipelines/issues/795

keithc-ca (approved):
Tested on Windows (which bundles openssl 3.0.10) via grinder: https://openj9-jenkins.osuosl.org/job/Grinder/2861.
None of the test failures appear to be related to this change.

PR #657, opened 2023-09-05
Parse -Xmso from more locations

pshipton:
To match OpenJ9, look in the options.default file, JAVA_TOOL_OPTIONS, and  IBM_JAVA_OPTIONS.

Issue https://github.com/eclipse-openj9/openj9/issues/16105

### DIFF
--- a/src/java.base/share/native/libjli/args.c
+++ b/src/java.base/share/native/libjli/args.c
@@ -373,6 +373,12 @@ static JLI_List expandArgFile(const char *arg) {
 
     /* arg file cannot be opened */
     if (fptr == NULL || fstat(fileno(fptr), &st) != 0) {
+        if (parsingOpenJ9Args) {
+            if (fptr != NULL) {
+                fclose(fptr);
+            }
+            return NULL;
+        }
         JLI_ReportMessage(CFG_ERROR6, arg);
         exit(1);
     } else {
@@ -385,7 +391,7 @@ static JLI_List expandArgFile(const char *arg) {
     rv = readArgFile(fptr);
 
     /* error occurred reading the file */
-    if (rv == NULL) {
+    if (rv == NULL && !parsingOpenJ9Args) {
         JLI_ReportMessage(ARG_ERROR18, arg);
         exit(1);
     }
@@ -507,7 +513,7 @@ JLI_ParseOpenJ9ArgsFromEnvVar(JLI_List args, const char *var_name) {
     jboolean savedStopExpansion = stopExpansion;
     jboolean savedRelaunch = relaunch;
 
-     if (NULL == env) {
+    if (NULL == env) {
         return JNI_FALSE;
     }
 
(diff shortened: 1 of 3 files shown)


src/java.base/share/native/libjli/args.c
```diff
@@ -531,6 +531,37 @@ JLI_ParseOpenJ9ArgsFromEnvVar(JLI_List args, const char *var_name) {
...
+    jboolean savedRelaunch = relaunch;
+
+    parsingOpenJ9Args = JNI_TRUE;
+    firstAppArgIndex = NOT_FOUND;
+    expectingNoDashArg = JNI_FALSE;
+    argsCount = 1;
+    stopExpansion = JNI_FALSE;
+    relaunch = JNI_FALSE;
+
+    result = expandArgFile(filename);
```
keithc-ca, line 552: If the named file does not exist or cannot be opened, a message is printed and the process exits: Is that the behavior we want?

pshipton: Good point. It shouldn't happen, but it probably would. I've pushed a fix.

src/java.base/share/native/libjli/java.c
```diff
@@ -328,29 +366,17 @@ JLI_Launch(int argc, char ** argv,              /* main argc, argv */
...
-                        if (threadStackSize > 0 && threadStackSize < (jlong)STACK_SIZE_MINIMUM) {
-                            threadStackSize = STACK_SIZE_MINIMUM;
-                        }
-                    }
-                    break;
-                }
-            }
-            JLI_List_free(openj9Args);
-        }
+        char optionsfile[MAXPATHLEN*2]; /* The size needs to be bigger than jrepath to avoid a warning. */
```
keithc-ca, line 374: We shouldn't assume a relationship between `MAXPATHLEN` and the length of the suffix we are adding:
```c
        char optionsfile[sizeof(jrepath) + sizeof("/lib/options.default") - 1];
```

keithc-ca (changes requested):

pshipton:
Updated.

src/java.base/share/native/libjli/java.c
```diff
@@ -328,29 +366,18 @@ JLI_Launch(int argc, char ** argv,              /* main argc, argv */
...
-                            threadStackSize = STACK_SIZE_MINIMUM;
-                        }
-                    }
-                    break;
-                }
-            }
-            JLI_List_free(openj9Args);
-        }
+#define OPTIONS_DEFAULT_PATH "/lib/options.default"
+        char optionsfile[sizeof(jrepath) + sizeof(OPTIONS_DEFAULT_PATH) - 1]; /* The size needs to be bigger than jrepath to avoid a warning. */
```
keithc-ca, line 375: I'm not sure the comment is helpful. It isn't just a warning we're avoiding here, but a potential buffer overflow if `jrepath` is long.

keithc-ca:
As I was testing this locally, I found that `-Xoptionsfile` is not recognized if it appears directly on the command line; perhaps it should be.

pshipton:
Added support for -Xoptionsfile on the command line.

src/java.base/share/native/libjli/java.c
```diff
@@ -957,6 +984,9 @@ AddOption(char *str, void *info)
     /* In OpenJ9 -Xmso is used to set native stack size instead of -Xss. -Xss is used to
      * set Java thread size only, which is handled in the JVM code.
      */
+    if (JLI_StrCCmp(str, "-Xoptionsfile=") == 0) {
```
keithc-ca, line 987: I think the comment above should be expanded to point out that `-Xmso` might be found in an options file.

pshipton: It seems obvious from the code.

pshipton: Updated the other comment to remove the obvious.

keithc-ca: Maybe it would help if the function names included "Xmso" (e.g. `parseXmoFile()` would be `parseXmsoFile()`).
The part that's less obvious here, is that as we consider the file named in `-Xoptionsfile=`, we're only paying attention to `-Xmso` options we find.

pshipton: Updated.

keithc-ca:

keithc-ca (approved):

PR #658, opened 2023-09-07

JasonFengJ9:

Conflicts:
```
	src/java.base/share/classes/jdk/internal/foreign/CABI.java
	src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
	src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```


src/java.base/share/classes/jdk/internal/foreign/CABI.java
```diff
@@ -85,6 +86,8 @@ private static CABI computeCurrent() {
             } else if (arch.equals("ppc64")) {
                 if (OperatingSystem.isAix()) {
                     return AIX_PPC_64;
+                } else if (OperatingSystem.isLinux()) {
+                    return LINUX_PPC_64;
                 }
```
keithc-ca, line 91: Please reorder these so the lines from upstream are unchanged:
```java
                if (OperatingSystem.isLinux()) {
                    return LINUX_PPC_64;
                } else if (OperatingSystem.isAix()) {
                    return AIX_PPC_64;
                }
```

keithc-ca (changes requested):

src/java.base/share/classes/jdk/internal/foreign/CABI.java
JasonFengJ9: Re-ordered.

keithc-ca (approved):

PR #661, opened 2023-09-13
Allow ChaCha20 cipher reuse with same key/IV or without init after decryption

KostasTsiounis:
The expected behaviour for reuse of `NativeChaCha20Cipher` for decryption is altered to match the existing Sun implementation in `ChaCha20Cipher`, as updated in OpenJDK21.

The following scenarios are now allowed for decryption:
- Reinitialize with same key + nonce
- Update `AAD` after a successful `doFinal()` without reinitializing
- Update `AAD` after an `AEADBadTagException`

Those scenarios are not allowed (i.e., an exception is thrown) in OpenJDK20 and earlier.

This fix addresses the failures mentioned in issue https://github.com/eclipse-openj9/openj9/issues/17632

### DIFF
--- a/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
+++ b/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
@@ -92,9 +92,24 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
     private static final NativeCrypto nativeCrypto;
     private static final Cleaner contextCleaner;
     private final long context;
+    // The initialization state of the OpenSSL context.
+    private boolean opensslInitialized;
 
-    // The previous mode, initialized to an illegal value.
-    private int prevMode = Cipher.WRAP_MODE;
+    // ossl_mode:
+    // 0 : ChaCha20-Poly1305 decrypt
+    // 1 : ChaCha20-Poly1305 encrypt
+    // 2 : ChaCha20 streaming
+    private int ossl_mode = -1;
+
+    // openssl_iv is only used by OpenSSL, here is the format:
+    // Streaming mode: 16 bytes
+    //                 first 4 bytes is the block counter (little-endian unsigned 32 bit int)
+    //                 the last 12 bytes is the nonce
+    // Poly1305 mode:  12 bytes nonce
+    private byte[] openssl_iv;
+
+    // The previous OpenSSL mode, initialized to an illegal value.
+    private int prevMode = -1;
 
     private final ByteArrayOutputStream aadBuf;
 
@@ -557,25 +572,18 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
             throw new InvalidKeyException("Unknown opmode: " + opmode);
         }
 
-        // Make sure that the provided key and nonce are unique before
-        // assigning them to the object.
         byte[] newKeyBytes = getEncodedKey(key);
-        checkKeyAndNonce(newKeyBytes, newNonce);
+        if (opmode == Cipher.ENCRYPT_MODE) {
+            // Make sure that the provided key and nonce are unique before
+            // assigning them to the object.  Key and nonce uniqueness
+            // protection is for encryption operations only.
+            checkKeyAndNonce(newKeyBytes, newNonce);
+        }
+        if (this.keyBytes != null) {
+            Arrays.fill(this.keyBytes, (byte)0);
+        }
         this.keyBytes = newKeyBytes;
         nonce = newNonce;
-
-        // ossl_mode:
-        // 0 : ChaCha20-Poly1305 decrypt
-        // 1 : ChaCha20-Poly1305 encrypt
-        // 2 : ChaCha20 streaming
-        int ossl_mode = -1;
-
-        // openssl_iv is only used by OpenSSL, here is the format:
-        // Streaming mode: 16 bytes
-        //                 first 4 bytes is the block counter (little-endian unsigned 32 bit int)
-        //                 the last 12 bytes is the nonce
-        // Poly1305 mode:  12 bytes nonce
-        byte[] openssl_iv = null;
 
         if (mode == MODE_NONE) {
             engine = new EngineStreamOnly();
@@ -585,7 +593,6 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
             openssl_iv = new byte[16];
             System.arraycopy(counter_byte, 0, openssl_iv, 0, counter_byte.length /* 4 */);
             System.arraycopy(nonce, 0, openssl_iv, 4, nonce.length /* 12 */);
-
         } else if (mode == MODE_AEAD) {
             openssl_iv = nonce;
             if (opmode == Cipher.ENCRYPT_MODE) {
@@ -599,27 +606,23 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
             }
         }
 
+        boolean modeChanged = false;
+        if (prevMode != ossl_mode) {
+            modeChanged = true;
+            prevMode = ossl_mode;
+        }
+
+        ChaCha20Init(modeChanged);
+
         direction = opmode;
         aadDone = false;
         initialized = true;
-
-        // Optimize the initialization of chacha20-poly1305 when they have the same ossl_mode
-        // otherwise, we have to reinitialize based upon the ossl_mode.
-        int ret;
-        if (prevMode == ossl_mode) {
-            ret = nativeCrypto.ChaCha20Init(context, ossl_mode, openssl_iv, openssl_iv.length, keyBytes, keyBytes.length, true);
-        } else {
-            ret = nativeCrypto.ChaCha20Init(context, ossl_mode, openssl_iv, openssl_iv.length, keyBytes, keyBytes.length, false);
-            prevMode = ossl_mode;
-        }
-        if (ret == -1) {
-            throw new ProviderException("Error in Native ChaCha20Cipher");
-        }
     }
 
     /**
      * Check the key and nonce bytes to make sure that they do not repeat
-     * across reinitialization.
+     * across reinitialization. This method is only useful for encryption
+     * mode.
      *
      * @param newKeyBytes the byte encoding for the newly provided key
      * @param newNonce the new nonce to be used with this initialization
@@ -743,9 +746,8 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
         } catch (ShortBufferException | KeyException exc) {
             throw new RuntimeException(exc);
         } finally {
-            // Regardless of what happens, the cipher cannot be used for
-            // further processing until it has been freshly initialized.
-            initialized = false;
+            // Reset the cipher's state to appropriate values.
+            resetStartState();
         }
         return output;
     }
@@ -781,9 +783,8 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
         } catch (KeyException ke) {
             throw new RuntimeException(ke);
         } finally {
-            // Regardless of what happens, the cipher cannot be used for
-            // further processing until it has been freshly initialized.
-            initialized = false;
+            // Reset the cipher's state to appropriate values.
+            resetStartState();
         }
         return bytesUpdated;
     }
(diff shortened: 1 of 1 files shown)


closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -92,6 +92,21 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
     private static final NativeCrypto nativeCrypto;
     private static final Cleaner contextCleaner;
     private final long context;
+    // The initialization state of the OpenSSL context
```
keithc-ca, line 95: Please finish each sentence with a period.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -558,25 +573,18 @@ private void init(int opmode, Key key, byte[] newNonce)
         }
 
         // Make sure that the provided key and nonce are unique before
-        // assigning them to the object.
+        // assigning them to the object.  Key and nonce uniqueness
+        // protection is for encryption operations only.
```
keithc-ca, line 577: I think it makes more sense for this comment to be inside the `if` statement beginning on line 579.
Also, I suggest you update the comment for `checkKeyAndNonce()` to mention that the check is only relevant for encryption mode.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -857,6 +857,31 @@ private static byte[] intToLittleEndian (long i) {
...
+        int ret = nativeCrypto.ChaCha20Init(context, ossl_mode, openssl_iv, openssl_iv.length, keyBytes, keyBytes.length, !modeChanged);
+
+        if (ret == -1) {
+            throw new ProviderException("Error in Native ChaCha20Cipher");
+        }
+        opensslInitialized = true;
+    }
+
+    /**
+     * Reset the Cipher's state to the appropriate values.
```
keithc-ca, line 872: I don't think `Cipher` deserves to be capitalized here.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -857,6 +857,31 @@ private static byte[] intToLittleEndian (long i) {
...
+     *
+     * Note: The cipher's internal "initialized" field is set differently
+     * for ENCRYPT_MODE and DECRYPT_MODE in order to allow DECRYPT_MODE
+     * ciphers to reuse the key/nonce/counter values.  This kind of reuse
+     * is disallowed in ENCRYPT_MODE.
+     */
+    private void resetStartState() {
+        aadDone = false;
+        initialized = (direction == Cipher.DECRYPT_MODE);
```
keithc-ca, line 881: I think this should only (potentially) clear `initialized`. If it's already set, leave it that way:
```java
        if (direction != Cipher.DECRYPT_MODE) {
            initialized = false;
        }
```

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -982,15 +1015,16 @@ public synchronized int doUpdate(byte[] in, int inOff, int inLen, byte[] out,
                     throw new ShortBufferException("Output buffer too small");
                 }
 
-                Objects.checkFromIndexSize(inOff, inLen, in.length);
+                if (in != null) {
```
keithc-ca, line 1018: Why isn't there a similar test `(in != null)` in `EngineStreamOnly.doUpdate()`?

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -966,6 +995,10 @@ private final class EngineAEADEnc implements ChaChaEngine {
         public synchronized int doUpdate(byte[] in, int inOff, int inLen, byte[] out,
                 int outOff) throws ShortBufferException, KeyException {
             if (initialized) {
+                if (!opensslInitialized) {
+                    ChaCha20Init(false);
+                }
+
                 // If this is the first update since AAD updates, signal that
                 // we're done processing AAD info
```
keithc-ca, line 1003: Please finish this sentence with a period.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -982,15 +1015,16 @@ public synchronized int doUpdate(byte[] in, int inOff, int inLen, byte[] out,
...
-                Objects.checkFromIndexSize(inOff, inLen, in.length);
+                if (in != null) {
+                    Objects.checkFromIndexSize(inOff, inLen, in.length);
 
-                byte aadArray[] = aadBuf.toByteArray();
-                aadBuf.reset();
-                int ret = nativeCrypto.ChaCha20Update(context, in, inOff, inLen, out, outOff, aadArray, aadArray.length);
-                if (ret == -1) {
-                    throw new ProviderException("Error in Native CipherBlockChaining");
+                    byte aadArray[] = aadBuf.toByteArray();
```
keithc-ca, line 1021: Please use a more conventional declaration (`[]` in the type):
```java
                    byte[] aadArray = aadBuf.toByteArray();
```
Likewise on line 1141.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -1001,15 +1035,16 @@ public synchronized int doUpdate(byte[] in, int inOff, int inLen, byte[] out,
         @Override
         public synchronized int doFinal(byte[] in, int inOff, int inLen, byte[] out,
                 int outOff) throws ShortBufferException, KeyException {
+            if ((in == null) && (inLen != 0)) {
+                throw new ShortBufferException("Input buffer in null, while input length is non-zero.");
```
keithc-ca, line 1039: I think this should say "is" instead of "in".

keithc-ca (changes requested):

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -92,6 +92,21 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
...
+    // 1 : ChaCha20-Poly1305 encrypt
+    // 2 : ChaCha20 streaming
+    private int ossl_mode = -1;
+
+    // openssl_iv is only used by OpenSSL, here is the format:
+    // Streaming mode: 16 bytes
+    //                 first 4 bytes is the block counter (little-endian unsigned 32 bit int)
+    //                 the last 12 bytes is the nonce
+    // Poly1305 mode:  12 bytes nonce
+    private byte[] openssl_iv = null;
```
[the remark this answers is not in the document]
KostasTsiounis: Done.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -92,6 +92,21 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
...
 
     // The previous mode, initialized to an illegal value.
     private int prevMode = Cipher.WRAP_MODE;
```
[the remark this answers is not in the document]
KostasTsiounis: Changed.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
KostasTsiounis: Done.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
KostasTsiounis: Moved it and added a mention in the method description.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
KostasTsiounis: Changed.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
KostasTsiounis: Changed.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
KostasTsiounis: Yep. Thanks for the catch. Fixed it.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
KostasTsiounis: Done.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
KostasTsiounis: Changed.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
KostasTsiounis: You're right. Changed.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -1001,15 +1036,16 @@ public synchronized int doUpdate(byte[] in, int inOff, int inLen, byte[] out,
         @Override
         public synchronized int doFinal(byte[] in, int inOff, int inLen, byte[] out,
                 int outOff) throws ShortBufferException, KeyException {
+            if ((in == null) && (inLen != 0)) {
+                throw new ShortBufferException("Input buffer is null, while input length is non-zero.");
+            }
```
keithc-ca, line 1041: Why is this check done only for `EngineAEADEnc` (and not also for `EngineAEADDec` and `EngineStreamOnly`)?

KostasTsiounis: For `EngineAEADEnc`, if the user provides a null `in`, but a non-zero `inLen`, a number of bytes will be left empty in the output array before writing the tag. There will be no indication that this has happened or an exception of any kind. This is why I added the check.

As far as `EngineStreamOnly` goes, it doesn't really matter since the value of `inLen` is never used if `in == null`. The only issue could be with the return statement in the `else` scenario which I could change to zero.

Lastly, in `EngineAEADDec`, there is a scenario where an issue would occur if `doUpdate()` is never called and `doFinal()` is called with `(in == null) && (inLen != 0)`. In that case, the native method will return -1 and an exception will be thrown, even though it might not be very helpful indeed.

I can, for completeness, add the same check in the other two engines, if you feel that would be the right thing to do.

keithc-ca: Looking at `CipherSpi` it seems nulls are not expected in that API. Perhaps this checking belongs at that level instead of in the classes that implement `ChaChaEngine`?

KostasTsiounis: Input can be null, in case you just want to finalize without passing any further information. But you're right, this check would be more suited in the `engineDoFinal()` of `NativeChaCha20Cipher`. I will move it there.

KostasTsiounis: As you mentioned, the check for `(in == null) && (inLen != 0)` is not required, as it is a scenario that cannot occur. All those possibilities are checked as part of the calls to the methods of `Cipher`.

I have removed the check altogether.

keithc-ca:
Jenkins compile amac jdknext

keithc-ca:
Testing underway in https://openj9-jenkins.osuosl.org/job/Grinder/2945/.

keithc-ca (approved):
Testing passed.

PR #664, opened 2023-09-19
Look for -XX:[+/-]UseZlibNX on AIX

pshipton:
AIX puts zlibNX on the LIBPATH when running on P9 or later, but having this in the environment may cause some things, such as git clone, to fail when exec'ed from Java. For example,
`fatal: pack has bad object at offset 3872100: inflate returned -5`

### DIFF
--- a/src/java.base/unix/native/libjli/java_md.c
+++ b/src/java.base/unix/native/libjli/java_md.c
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2020, 2021 All Rights Reserved
+ * (c) Copyright IBM Corp. 2020, 2023 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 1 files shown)


src/java.base/unix/native/libjli/java_md.c
```diff
@@ -375,12 +375,33 @@ CreateExecutionEnvironment(int *pargc, char ***pargv,
...
          * followed by the user's previous effective LD_LIBRARY_PATH, if
          * any.
          */
 
+#ifdef AIX
+        int argc = *pargc;
+        char **argv = *pargv;
+        char *arg = NULL;
+        jboolean useZlibNX = JNI_TRUE;
+        while (argc > 0 && *(arg = *argv) == '-') {
```
keithc-ca, line 389: This needs to skip the first element of `argv` which is the name of the launcher:
```
Command line args:
argv[0] = ./jdk/bin/java
argv[1] = -XX:-UseZlibNX
argv[2] = -version
```
I suggest:
```c
        int argc = *pargc - 1;
        char **argv = *pargv + 1;
        jboolean useZlibNX = JNI_TRUE;
        for (; argc > 0; --argc, ++argv) {
            const char *arg = *argv;
            if ('-' != *arg) {
                break;
            } else if (JLI_StrCmp(arg, "-XX:+UseZlibNX") == 0) {
...
```

keithc-ca (changes requested):

src/java.base/unix/native/libjli/java_md.c
pshipton: I've updated it. The while loop is the same as used in java.c SelectVersion() and ParseArgument(), I think we can keep it as is?

src/java.base/unix/native/libjli/java_md.c
```diff
@@ -375,12 +375,33 @@ CreateExecutionEnvironment(int *pargc, char ***pargv,
...
+         *     o          ZLIBNX_PATH (for AIX P9 or newer systems with NX, unless -XX:-UseZlibNX is set)
          *
          * followed by the user's previous effective LD_LIBRARY_PATH, if
          * any.
          */
 
+#ifdef AIX
+        int argc = *pargc - 1; // skip the launcher name
+        char **argv = *pargv + 1;
+        char *arg = NULL;
```
keithc-ca, line 387: Write access is not needed:
```c
        const char *arg = NULL;
```

src/java.base/unix/native/libjli/java_md.c
```diff
@@ -375,12 +375,33 @@ CreateExecutionEnvironment(int *pargc, char ***pargv,
...
+            } else if (JLI_StrCmp(arg, "-XX:-UseZlibNX") == 0) {
+                useZlibNX = JNI_FALSE;
+            }
+            argc--;
+            argv++;
+        }
+        useZlibNX = useZlibNX && power_9_andup() && power_nx_gzip();
+        if (JLI_IsTraceLauncher()) {
+            printf("Add " ZLIBNX_PATH " to the LIBPATH: %s P9+ %s NX %s\n", useZlibNX ? "TRUE" : "FALSE",
+                power_9_andup() ? "TRUE" : "FALSE", power_nx_gzip() ? "TRUE" : "FALSE");
```
keithc-ca, line 401: Can you split this to make it easier to read?
```c
            printf("Add " ZLIBNX_PATH " to the LIBPATH: %s P9+ %s NX %s\n",
                useZlibNX ? "TRUE" : "FALSE",
                power_9_andup() ? "TRUE" : "FALSE",
                power_nx_gzip() ? "TRUE" : "FALSE");
```

pshipton: Updated for both comments.

keithc-ca:
Is something being done about the underlying issue (that breaks things like git)?

keithc-ca:
Jenkins test sanity aix,alinux64 jdknext

pshipton:
> Is something being done about the underlying issue (that breaks things like git)?

I believe they looked at it and fixed a problem, but we've had some recent machine updates and it's still not resolved in the version we have installed.
https://github.com/eclipse-openj9/openj9/issues/14948#issuecomment-1210725785

pshipton:
@zl-wang fyi this new option to control the behavior of using zlibNX.

keithc-ca (approved):

pshipton:
Doc issue https://github.com/eclipse-openj9/openj9-docs/issues/1178

PR #666, opened 2023-09-20
Register natives in ConstantPool init

tajila:
Depends on https://github.com/eclipse-openj9/openj9/pull/18169

### DIFF
--- a/src/java.base/share/classes/jdk/internal/reflect/ConstantPool.java
+++ b/src/java.base/share/classes/jdk/internal/reflect/ConstantPool.java
@@ -22,6 +22,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * ===========================================================================
+ */
 
 package jdk.internal.reflect;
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/jdk/internal/reflect/ConstantPool.java
```diff
@@ -22,7 +22,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
-
```
keithc-ca, line 25: Please insert the copyright notice before this line instead of replacing it.

keithc-ca (changes requested):

keithc-ca (approved):

keithc-ca:

tajila:
Only failures are:

```
FAILED: test_verifyJVMTIMacros
java.lang.AssertionError: JVMTI_VTHREAD_STATE_YIELDING (7) does not match VirtualThread.YIELDING (10)
	at org.testng.Assert.fail(Assert.java:96)
```

unrelated to this PR.

FYI @babsingh

keithc-ca:
I created https://github.com/eclipse-openj9/openj9/issues/18183 to address the test failure.

PR #672, opened 2023-09-29
Always revert to Java impl when OpenSSL fails

KostasTsiounis:
At the moment, if there is an OpenSSL failure when using `NativeECKeyPairGenerator`, a flag for that particular curve is set and the Sun Java implementation is used to complete the task. 

If another instance of `NativeECKeyPairGenerator` reaches the same failure, a `ProviderException` is thrown, as the issue with that particular curve should have been discovered during the call to `initialize()` thus creating and using an instance of `ECKeyPairGenerator`, and that particular point should never have been reached.

However, in the scenario where the second instance of `NativeECKeyPairGenerator` has been initialized earlier, the flag for the problematic curve will never be discovered.

This fix ensures that we always revert to the Sun Java implementation when an OpenSSL failure occurs.

### DIFF
--- a/closed/src/java.base/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
+++ b/closed/src/java.base/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
@@ -166,17 +166,13 @@ public final class NativeECKeyPairGenerator extends KeyPairGeneratorSpi {
             return this.javaImplementation.generateKeyPair();
         }
 
-        boolean absent;
         long nativePointer = NativeECUtil.encodeGroup(this.params);
 
         if (nativePointer == -1) {
-            absent = NativeECUtil.putCurveIfAbsent(this.curve, Boolean.FALSE);
-            if (!absent) {
-                throw new ProviderException("Could not encode group");
-            }
+            NativeECUtil.putCurveIfAbsent(this.curve, Boolean.FALSE);
             if (nativeCryptTrace) {
-                System.err.println(this.curve +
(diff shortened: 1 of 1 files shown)


closed/src/java.base/share/classes/sun/security/ec/NativeECKeyPairGenerator.java
```diff
@@ -170,13 +170,10 @@ public KeyPair generateKeyPair() {
         long nativePointer = NativeECUtil.encodeGroup(this.params);
 
         if (nativePointer == -1) {
-            absent = NativeECUtil.putCurveIfAbsent(this.curve, Boolean.FALSE);
```
keithc-ca, line 173: Please remove the (now unused) local variable `absent`.

KostasTsiounis: Done.

KostasTsiounis:
I updated the commit message too. I couldn't make it shorter and still explain what is going on.

keithc-ca:
> I couldn't make it shorter and still explain what is going on.

There's no practical limit to the length of the commit message. I think the description captures the issue well and deserved to be in the commit message. Thanks.

keithc-ca:

keithc-ca:
The failures are known; see https://github.com/eclipse-openj9/openj9/issues/18183.

keithc-ca (approved):

KostasTsiounis:
FYI @keithc-ca

PR #681, opened 2023-10-24
Revert "8266242: java/awt/GraphicsDevice/CheckDisplayModes.java failing on macOS 11 ARM"

keithc-ca:
This reverts commit 24daf37d2f71472821832dbfcfcb7bb396035dc5.

See https://github.com/eclipse-openj9/openj9/issues/18334.

PR #682, opened 2023-10-24
Disable JVMCI feature for OpenJ9

keithc-ca:
Avoid problem seen in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/681#issuecomment-1777312700.


pshipton:
jenkins compile alinux64 jdknext

pshipton (approved):

PR #684, opened 2023-10-27
Replace date utility usage with DateUtil for SOURCE_DATE

joransiu:
On some platforms, the date utility doesn't support the "%s" format specifier or supports neither the "-f" option nor the "-j" option. Use the boot jdk to run a java program instead.

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -517,7 +517,7 @@ run-preprocessors-j9 : stage-j9
 			OMR_DIR=$(OUTPUTDIR)/vm/omr \
 			SOURCETOOLS_DIR=$(call MixedPath,$(OPENJ9_TOPDIR))/sourcetools \
 			SPEC=$(OPENJ9_BUILDSPEC) \
-			UMA_OPTIONS_EXTRA="-buildDate $(shell date +'%Y%m%d')" \
+			UMA_OPTIONS_EXTRA="-buildDate $(shell $(JAVA_SMALL) $(TOPDIR)/make/src/classes/DateUtil.java --format=YYYYMMdd)" \
 			VERSION_MAJOR=$(VERSION_FEATURE) \
 			tools
 
--- a/closed/get_j9_source.sh
+++ b/closed/get_j9_source.sh
@@ -126,10 +126,7 @@ for i in "$@" ; do
 	esac
 done
 
-# clone OpenJ9 repos
-date '+[%F %T] Get OpenJ9 sources'
-START_TIME=$(date +%s)
-
+# Extract the various source repositories.
 for i in "${!git_urls[@]}" ; do
 	branch=${branches[$i]}
 
@@ -172,9 +169,6 @@ if [ ${pflag} = true ] ; then
 	# wait for all subprocesses to complete
 	wait
 fi
-
-END_TIME=$(date +%s)
-date "+[%F %T] OpenJ9 clone repositories finished in $(($END_TIME - $START_TIME)) seconds"
 
 for i in "${!git_urls[@]}" ; do
 	if [ -e /tmp/${i}.pid.rc ] ; then
--- a/make/autoconf/jdk-options.m4
+++ b/make/autoconf/jdk-options.m4
@@ -24,7 +24,7 @@
 #
 
 # ===========================================================================
-# (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
+# (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
 # ===========================================================================
 
 ###############################################################################
@@ -748,7 +748,7 @@ AC_DEFUN_ONCE([JDKOPT_SETUP_REPRODUCIBLE_BUILD],
       AC_MSG_RESULT([$SOURCE_DATE, from SOURCE_DATE_EPOCH])
     else
       # Tell makefiles to take the time from configure
-      SOURCE_DATE=$($DATE +"%s")
+      SOURCE_DATE=$($JAVA $TOPDIR/make/src/classes/DateUtil.java)
       AC_MSG_RESULT([$SOURCE_DATE, from 'current' (default)])
     fi
   elif test "x$with_source_date" = xupdated; then


keithc-ca:
While you're here, could you also fix the uses of `date +%...` in `closed/get_j9_source.sh` and `closed/OpenJ9.gmk`?

It's not clear where the boot JDK is as `get_j9_source.sh` runs, but using `awk` might be reasonable:
* `date "+[%F %T] message"` could be replaced by `echo "message" | awk '{print strftime("[%F %T]"), $0 }'`
* `date +'%s'` could be replaced by `awk </dev/null 'END { print systime() }'`

joransiu:
The awk functions don't seem to be supported on z/OS.

```
$ echo "message" | awk '{print strftime("[%F %T]"), $0 }'
awk: line 0 (NR=1): variable "strftime" cannot be used as a function
$ awk </dev/null 'END { print systime() }'
awk: line 0 (END): variable "systime" cannot be used as a function
```

keithc-ca:
> The awk functions don't seem to be supported on z/OS.

That's unfortunate. Do you have other suggestions?

keithc-ca:
Perhaps we should just remove those uses of `date` from `get_j9_source.sh`. I expect the vast majority of the time, jenkins is also reporting the time in which case the extra information printed is redundant.

For example, see https://openj9-jenkins.osuosl.org/job/Build_JDKnext_aarch64_linux_OpenJDK/410/consoleText
```
[2023-10-27T03:22:04.590Z] [2023-10-27 03:22:04] Get OpenJ9 sources
[2023-10-27T03:22:04.590Z] 
[2023-10-27T03:22:04.590Z] Clone repository: omr
[2023-10-27T03:22:04.590Z] 
[2023-10-27T03:22:04.590Z] Cloning into 'omr'...
[2023-10-27T03:22:09.753Z] 
[2023-10-27T03:22:09.754Z] Clone repository: openj9
[2023-10-27T03:22:09.754Z] 
[2023-10-27T03:22:09.754Z] Cloning into 'openj9'...
[2023-10-27T03:22:14.889Z] [2023-10-27 03:22:13] OpenJ9 clone repositories finished in 9 seconds
```
Each of the clone start time, the clone completion time and the elapsed time is presented or is computable (with greater precision if anyone is interested) from the prefix added by jenkins.

joransiu:
@keithc-ca :  I propose we remove the `[%F %T]`, as I agree it's redundant.   FWIW, I found that `date` does support the %F %T tokens on z/OS.

```sh
$ date "+[%F %T] message"
[2023-10-31 07:58:09] message
````


For the `date +"%s"` where it's used to calculate elapsed time, I think it'll be nice to have that there.  I found the following awk command using `srand()` to work.  

```sh
$ awk 'BEGIN {srand(); print srand()}'
1698764214
```

> srand([x])
> 
>     Set the starting point, or seed, for generating random numbers to the value x.
> ...
>     If the argument x is omitted, as in ‘srand()’, then the current date and time of day are used for a seed. This is the way to get random numbers that are truly unpredictable.

https://www.gnu.org/software/gawk/manual/html_node/Numeric-Functions.html

What do you think?

joransiu:
@keithc-ca : I've implemented what I proposed above. Can you take another look?  Thx!

closed/get_j9_source.sh
```diff
@@ -127,8 +127,8 @@ for i in "$@" ; do
 done
 
 # clone OpenJ9 repos
-date '+[%F %T] Get OpenJ9 sources'
-START_TIME=$(date +%s)
+echo "Get OpenJ9 sources"
+START_TIME=$(awk 'BEGIN {srand(); print srand()}')
```
keithc-ca, line 131: I'm surprised `awk` doesn't seem to read stdin, but today I learned that if the only pattern is `BEGIN`, `awk` doesn't read input (unless `getline` is used).

I don't think this output (nor the timing) is needed; I think it should simply be deleted (and the comment on line 129 corrected - it's not just about OpenJ9).

keithc-ca (changes requested):
There's still a use of `date +'%Y%m%d'` in `OpenJ9.gmk` that should be replaced.

joransiu:
@keithc-ca : I've removed the output messages from `get_j9_source.sh`, and updated the OpenJ9.gmk date usage to use DateUtil.  Thx.

closed/OpenJ9.gmk
```diff
@@ -517,7 +517,7 @@ run-preprocessors-j9 : stage-j9
 			OMR_DIR=$(OUTPUTDIR)/vm/omr \
 			SOURCETOOLS_DIR=$(call MixedPath,$(OPENJ9_TOPDIR))/sourcetools \
 			SPEC=$(OPENJ9_BUILDSPEC) \
-			UMA_OPTIONS_EXTRA="-buildDate $(shell date +'%Y%m%d')" \
+			UMA_OPTIONS_EXTRA="-buildDate $(shell $JAVA $TOPDIR/make/src/classes/DateUtil.java --format=yyyyMMdd)" \
```
keithc-ca, line 520: This must use `$(var)` to reference make macros; also suggest this can use `JAVA_SMALL`:
```
			UMA_OPTIONS_EXTRA="-buildDate $(shell $(JAVA_SMALL) $(TOPDIR)/make/src/classes/DateUtil.java --format=YYYYMMdd)" \
```

closed/get_j9_source.sh
```diff
@@ -126,10 +126,7 @@ for i in "$@" ; do
 	esac
 done
 
-# clone OpenJ9 repos
-date '+[%F %T] Get OpenJ9 sources'
-START_TIME=$(date +%s)
-
+# Get OpenJ9, OMR sources
```
keithc-ca, line 129: This comment should be more general (internally there are other repos); please finish the sentence with a period.

keithc-ca (changes requested):

closed/OpenJ9.gmk
joransiu: Changed as suggested under: [b5249ac](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/684/commits/b5249aced00439e5fb5a923d0cea7118f1316d37)

keithc-ca (approved):

PR #694, opened 2023-11-16
CRIU adds @NotCheckpointSafe for PrintStream.writeln(String)

JasonFengJ9:
`CRIU` adds `@NotCheckpointSafe` for `PrintStream.writeln(String)`

related
* https://github.com/eclipse-openj9/openj9/issues/18471

### DIFF
--- a/src/java.base/share/classes/java/io/PrintStream.java
+++ b/src/java.base/share/classes/java/io/PrintStream.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * ===========================================================================
+ */
+
 package java.io;
 
 import java.util.Formatter;
@@ -33,6 +39,10 @@ import java.nio.charset.UnsupportedCharsetException;
 import jdk.internal.access.JavaIOPrintStreamAccess;
 import jdk.internal.access.SharedSecrets;
 import jdk.internal.misc.InternalLock;
+
+/*[IF CRIU_SUPPORT]*/
+import openj9.internal.criu.NotCheckpointSafe;
+/*[ENDIF] CRIU_SUPPORT */
 
 /**
  * A {@code PrintStream} adds functionality to another output stream,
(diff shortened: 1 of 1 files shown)


tajila (approved):
Need to fix copywrites

JasonFengJ9:
@tajila This is ready for another look.

tajila:
jenkins test sanity zlinux jdknext

JasonFengJ9:
> https://openj9-jenkins.osuosl.org/job/Test_openjdknext_j9_sanity.functional_s390x_linux_Personal/42/tapResults/

A known issue
* https://github.com/eclipse-openj9/openj9/issues/17474

src/java.base/share/classes/java/io/PrintStream.java
```diff
@@ -818,6 +828,9 @@ private void implWrite(String s) throws IOException {
     // using println, but since subclasses could exist which depend on
     // observing a call to print followed by newLine we only use this if
     // getClass() == PrintStream.class to avoid compatibility issues.
+    /*[IF CRIU_SUPPORT]*/
+    @NotCheckpointSafe
```
keithc-ca, line 832: Why should this not apply to `writeln(char[])` as well?
What about other methods that invoke `lock.lock()`, e.g. `close()` and `flush()`?
Might a better place for this be on `InternalLock.lock()`?

JasonFengJ9: The `@NotCheckpointSafe` annotation is to delay taking a checkpoint (for a certain amount of attempts) while a non checkpoint thread owns the global resource.
While taking a checkpoint, single threaded mode is enabled, i.e., only current checkpoint thread is active, other Java threads are paused.
* https://github.com/eclipse-openj9/openj9/issues/18471

In this issue, non-checkpoint thread took the `lock` at https://github.com/ibmruntimes/openj9-openjdk-jdk21/blob/2278b32d9391765c7736be31ff94dac09fd56cbb/src/java.base/share/classes/java/io/PrintStream.java#L839, and was paused at single threaded mode.
When the checkpoint thread attempted to acquire the `lock` at https://github.com/ibmruntimes/openj9-openjdk-jdk21/blob/2278b32d9391765c7736be31ff94dac09fd56cbb/src/java.base/share/classes/java/io/PrintStream.java#L837, parked there indefinitely and a `JVMCheckpointException` was thrown.

keithc-ca: That isn't very reassuring: waiting until users have problems before fixing the code. Shouldn't we be more proactive?

JasonFengJ9: This issue is specific to the application, which is tailored to JVM CRIU currently. 
It is intermittent, and only occurs at CRIU checkpoint or restore single-threaded mode stage. It won't happen whenever the application starts running.
JVM throws `JVMCRIUException` to indicate the code causing the issue.

PR #701, opened 2023-11-23
Update RestrictedSecurity flags, alter debug comments and profile name

KostasTsiounis:
The original flag to enable FIPS (i.e., `-Dsemeru.fips=true`) remains the same, but the one allowing a user to set a custom profile is changed to `-Dsemeru.customprofile=<profile.version>`.

The debug messages have been altered a bit to only be enabled using the already known and used by similar components `-Djava.security.auth.debug` flag. The information for available profiles, as well as the profile used in the particular run, is printed as part of the debug messages, instead of specifying additional properties in the custom profile flag.

Further checks are added to ensure solutions are supported in the running platform and the profile is marked as FIPS compliant. 

The flag for the custom profile allows the user to either specify the full name of the profile to be used (e.g., `-Dsemeru.customprofile=NSS.FIPS140-2`), or specify the solution to be used (e.g., `-Dsemeru.customprofile=NSS`) and allow `RestrictedSecurity` to pick the default profile for that.

The naming of profiles has, also, been altered to abide by the `<solution.version>` template (e.g., `NSS.FIPS140-2`), instead of an integer.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -32,12 +32,16 @@ import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import sun.security.util.Debug;
 
@@ -50,41 +54,104 @@ public final class RestrictedSecurity {
 
     // Restricted security mode enable check, only supported on Linux x64.
     private static final boolean userEnabledFIPS;
-    private static final boolean userEnabledSecurity;
-    private static final boolean isSecuritySupported;
+    private static boolean isFIPSSupported;
+    private static boolean isFIPSEnabled;
+
+    private static final boolean isNSSSupported;
+    private static final boolean isOpenJCEPlusSupported;
+
+    private static final boolean userSetProfile;
     private static final boolean shouldEnableSecurity;
-    private static final String userSecuritySetting;
+    private static String selectedProfile;
+    private static String profileID;
 
     private static boolean securityEnabled;
 
-    private static int userSecurityNum;
-    private static boolean userSecurityTrace;
-    private static boolean userSecurityAudit;
-    private static boolean userSecurityHelp;
+    private static String userSecurityID;
 
     private static RestrictedSecurityProperties restricts;
 
-    private static final List<String> supportPlatforms = List.of("amd64", "ppc64le", "s390x");
+    private static final Map<String, List<String>> supportedPlatformsNSS = new HashMap<>();
+    private static final Map<String, List<String>> supportedPlatformsOpenJCEPlus = new HashMap<>();
 
     static {
+        supportedPlatformsNSS.put("Arch", List.of("amd64", "ppc64le", "s390x"));
+        supportedPlatformsNSS.put("OS", List.of("Linux"));
+
+        supportedPlatformsOpenJCEPlus.put("Arch", List.of("amd64", "ppc64"));
+        supportedPlatformsOpenJCEPlus.put("OS", List.of("Linux", "AIX", "Windows"));
+
         @SuppressWarnings("removal")
         String[] props = AccessController.doPrivileged(
                 new PrivilegedAction<>() {
                     @Override
                     public String[] run() {
                         return new String[] { System.getProperty("semeru.fips"),
-                                System.getProperty("semeru.restrictedsecurity"),
+                                System.getProperty("semeru.customprofile"),
                                 System.getProperty("os.name"),
                                 System.getProperty("os.arch") };
                     }
                 });
+
+        boolean isOsSupported, isArchSupported;
+        // Check whether the NSS FIPS solution is supported.
+        isOsSupported = false;
+        for (String os: supportedPlatformsNSS.get("OS")) {
+            if (props[2].contains(os)) {
+                isOsSupported = true;
+            }
+        }
+        isArchSupported = false;
+        for (String arch: supportedPlatformsNSS.get("Arch")) {
+            if (props[3].contains(arch)) {
+                isArchSupported = true;
+            }
+        }
+        isNSSSupported = isOsSupported && isArchSupported;
+
+        // Check whether the OpenJCEPlus FIPS solution is supported.
+        isOsSupported = false;
+        for (String os: supportedPlatformsOpenJCEPlus.get("OS")) {
+            if (props[2].contains(os)) {
+                isOsSupported = true;
+            }
+        }
+        isArchSupported = false;
+        for (String arch: supportedPlatformsOpenJCEPlus.get("Arch")) {
+            if (props[3].contains(arch)) {
+                isArchSupported = true;
+            }
+        }
+        isOpenJCEPlusSupported = isOsSupported && isArchSupported;
+
+        // Check the default solution to see if FIPS is supported.
+        isFIPSSupported = isNSSSupported;
+
         userEnabledFIPS = Boolean.parseBoolean(props[0]);
-        // If semeru.fips is true, then ignore semeru.restrictedsecurity, use userSecurityNum 1.
-        userSecuritySetting = userEnabledFIPS ? "1" : props[1];
-        userEnabledSecurity = !isNullOrBlank(userSecuritySetting);
-        isSecuritySupported = "Linux".equalsIgnoreCase(props[2])
-                && supportPlatforms.contains(props[3]);
-        shouldEnableSecurity = (userEnabledFIPS || userEnabledSecurity) && isSecuritySupported;
+
+        if (userEnabledFIPS) {
+            if (isFIPSSupported) {
+                // Set to default profile for the default FIPS solution.
+                selectedProfile = "NSS.140-2";
+            }
+        }
+
+        // If user has specified a profile, use that
+        if (props[1] != null) {
+            selectedProfile = props[1];
+            userSetProfile = true;
+        } else {
+            userSetProfile = false;
+        }
+
+        // Check if FIPS is supported on this platform without explicitly setting a profile.
+        if (userEnabledFIPS && !isFIPSSupported && !userSetProfile) {
+            printStackTraceAndExit("FIPS mode is not supported on this platform by default.\n"
+                    + " Use the semeru.customprofile system property to use an available FIPS-compliant profile.\n"
+                    + " Note: Not all platforms support FIPS at the moment.");
+        }
+
+        shouldEnableSecurity = (userEnabledFIPS && isFIPSSupported) || userSetProfile;
     }
 
     private RestrictedSecurity() {
@@ -95,8 +162,9 @@ public final class RestrictedSecurity {
      * Check if restricted security mode is enabled.
      *
      * Restricted security mode is enabled when, on supported platforms,
-     * the semeru.restrictedsecurity system property is set or the system
-     * property semeru.fips is true.
+     * the semeru.customprofile system property is used to set a
+     * specific security profile or the semeru.fips system property is
+     * set to true.
      *
      * @return true if restricted security mode is enabled
      */
@@ -139,13 +207,16 @@ public final class RestrictedSecurity {
     /**
      * Check if the FIPS mode is enabled.
      *
-     * FIPS mode will be enabled when the semeru.fips system property is true,
-     * or semeru.restrictedsecurity system property is set by using FIPS policy.
+     * FIPS mode will be enabled when the semeru.fips system property is
+     * true, and the RestrictedSecurity mode has been successfully initialized.
      *
      * @return true if FIPS is enabled
      */
     public static boolean isFIPSEnabled() {
-        return securityEnabled && (userSecurityNum == 1);
+        if (securityEnabled) {
+            return isFIPSEnabled;
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -198,6 +271,99 @@ public static boolean isProviderAllowed(Class<?> providerClazz) {
...
+                printStackTraceAndExit("No default profile was found for " + selectedProfile);
+            } else {
+                profileID = defaultMatch;
+            }
+        }
+    }
+
+    private static void checkIfKnownProfileSupported() {
+        if (profileID.contains("NSS") && !isNSSSupported) {
+            printStackTraceAndExit("NSS profiles are not supported in this platform.");
```
pshipton, line 343: Should be "on" instead of "in".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -198,6 +271,99 @@ public static boolean isProviderAllowed(Class<?> providerClazz) {
...
+                    if (key.startsWith(potentialProfileID)) {
+                        // Find the property for specifying if this is the default profile.
+                        if (key.endsWith(".desc.default")) {
+                            // Check if property is set to true.
+                            if (Boolean.parseBoolean(props.getProperty(key))) {
+                                // Check if multiple defaults exist and act accordingly.
+                                if (defaultMatch == null) {
+                                    defaultMatch = key.split("\\.desc")[0];
+                                } else {
+                                    printStackTraceAndExit("Multiple default profiles for " + selectedProfile);
```
pshipton, line 326: All these messages can be more explicit and contain "Restricted Security" or similar, some of them already do.
i.e. "Multiple Restricted Security default profiles for ".
We know the context while implementing and testing, but it may not be clear to a user, in particular if the command line options are not explicitly set by the user but inherited from somewhere.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -198,6 +271,99 @@ public static boolean isProviderAllowed(Class<?> providerClazz) {
...
+        }
+    }
+
+    private static void checkIfKnownProfileSupported() {
+        if (profileID.contains("NSS") && !isNSSSupported) {
+            printStackTraceAndExit("NSS profiles are not supported in this platform.");
+        }
+
+        if (profileID.contains("OpenJCEPlus") && !isOpenJCEPlusSupported) {
+            printStackTraceAndExit("OpenJCEPlus profiles are not supported in this platform.");
```
pshipton, line 347: "on" instead of "in". Pls check the rest, there are more.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -198,6 +271,99 @@ public static boolean isProviderAllowed(Class<?> providerClazz) {
...
+        }
+
+        if (debug != null) {
+            debug.println("Profile is supported in this platform.");
```
pshipton, line 351: This could print `selectedProfile`

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -198,6 +271,99 @@ public static boolean isProviderAllowed(Class<?> providerClazz) {
...
+        }
+    }
+
+    private static void checkFIPSCompatibility(Properties props) {
+        boolean isFIPSProfile = Boolean.parseBoolean(props.getProperty(profileID + ".desc.fips"));
+        if (isFIPSProfile) {
+            if (debug != null) {
+                debug.println("Profile specified as FIPS compliant.");
```
pshipton, line 359: This could print `selectedProfile`

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -887,101 +997,102 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
-                String desc = securityProps.getProperty("RestrictedSecurity" + num + ".desc.name");
-                if ((desc == null) || desc.trim().isEmpty()) {
-                    break;
+            System.out.println("Restricted Security Available Profiles' Info:");
+            System.out.println("=============================================");
+
+            Set<String> availableProfiles = new HashSet<>();
+            for(Object securityFileObject : securityProps.keySet()) {
+                if (securityFileObject instanceof String key) {
+                    Pattern profileNamePattern = Pattern.compile("^(RestrictedSecurity\\.\\S+)\\.desc\\.name");
```
pshipton, line 1008: This shouldn't be in the loop.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -887,101 +997,102 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                    Matcher profileMatcher = profileNamePattern.matcher(key);
+                    if(profileMatcher.matches()) {
```
pshipton, line 1010: Space after if

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -887,101 +997,102 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
-                        + securityProps.getProperty("RestrictedSecurity" + num + ".desc.name"));
-                System.out.println("RestrictedSecurity" + num + ".desc.number: "
-                        + parseProperty(securityProps.getProperty("RestrictedSecurity" + num + ".desc.number")));
-                System.out.println("RestrictedSecurity" + num + ".desc.policy: "
-                        + parseProperty(securityProps.getProperty("RestrictedSecurity" + num + ".desc.policy")));
-                System.out.println("RestrictedSecurity" + num + ".desc.sunsetDate: "
-                        + parseProperty(securityProps.getProperty("RestrictedSecurity" + num + ".desc.sunsetDate")));
-                System.out.println();
+            }
+            System.out.println("Those are the available Restricted Security profiles:\n");
```
pshipton, line 1015: Those -> These

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -251,7 +416,7 @@ public static boolean configure(Properties props) {
 
                 // Print out the Trace info.
                 if (userSecurityTrace) {
```
pshipton, line 418: I don't see anywhere that sets this to true. Simliarly userSecurityAudit and userSecurityHelp are still declared.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -887,101 +997,102 @@ boolean isRestrictedProviderAllowed(String providerName) {
         /**
          * List audit info if userSecurityAudit is true, default as false.
```
pshipton, line 998: userSecurityAudit is no longer used, pls update the comment.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
pshipton: This could be "The available Restricted Security profiles:"

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -95,8 +164,9 @@ private RestrictedSecurity() {
      * Check if restricted security mode is enabled.
      *
      * Restricted security mode is enabled when, on supported platforms,
-     * the semeru.restrictedsecurity system property is set or the system
-     * property semeru.fips is true.
+     * the restrictedsecurity.profile system property is used to set a
```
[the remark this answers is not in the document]
KostasTsiounis: Sorry, forgot to change it. It's fixed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -198,6 +271,99 @@ public static boolean isProviderAllowed(Class<?> providerClazz) {
...
     }
 
+    /**
+     * Figure out the full profile ID.
+     *
+     * Use the default or user selected profile and attempt to find
+     * an appropriate entry in the java.security properties.
+     *
+     * If a profile cannot be found, or multiple defaults are discovered
+     * for a single profile,an appropriate message is printed and the
```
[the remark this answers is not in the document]

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Fixed everywhere.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Added references to `RestrictedSecurity` for clarity.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Fixed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Moved outside the loop.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Fixed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Changed to your suggestion.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Those were part of the previous iteration and I missed them when cleaning up. Sorry about that. Removed all mentions and definitions.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Same.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -198,6 +268,106 @@ public static boolean isProviderAllowed(Class<?> providerClazz) {
...
+        if (selectedProfile.indexOf(".") != -1) {
+            /* The default profile is used, or the user specified the
+             * full <profile.version>.
+             */
+            if (debug != null) {
+                debug.println("Profile specified using full name (i.e., <profile.version>): "
+                        + selectedProfile);
+            }
+            for (Object keyObject : props.keySet()) {
+                // A matching property is found.
```
pshipton, line 295: This comment doesn't seem appropriate here.  I don't think it's needed at all. Same in the else case.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -198,6 +268,106 @@ public static boolean isProviderAllowed(Class<?> providerClazz) {
...
+                        + selectedProfile);
+            } else {
+                profileID = defaultMatch;
+            }
+        }
+    }
+
+    private static void checkIfKnownProfileSupported() {
+        if (profileID.contains("NSS") && !isNSSSupported) {
+            printStackTraceAndExit("NSS profiles are not supported on this platform.");
```
pshipton, line 344: RestrictedSecurity profiles ...

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -198,6 +268,106 @@ public static boolean isProviderAllowed(Class<?> providerClazz) {
...
+        }
+
+        if (profileID.contains("OpenJCEPlus") && !isOpenJCEPlusSupported) {
+            printStackTraceAndExit("OpenJCEPlus profiles are not supported on this platform.");
```
pshipton, line 348: RestrictedSecurity profiles ...

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -50,41 +54,103 @@ public final class RestrictedSecurity {
...
+        if (props[1] != null) {
+            selectedProfile = props[1];
+            userSetProfile = true;
+        } else {
+            userSetProfile = false;
+        }
+
+        // Check if FIPS is supported on this platform without explicitly setting a profile.
+        if (userEnabledFIPS && !isFIPSSupported && !userSetProfile) {
+            printStackTraceAndExit("FIPS mode is not supported on this platform by default.");
```
pshipton, line 150: This could provide additional help, such as "Set the semeru.customprofile system property."

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -198,6 +268,106 @@ public static boolean isProviderAllowed(Class<?> providerClazz) {
...
+        boolean isFIPSProfile = Boolean.parseBoolean(props.getProperty(profileID + ".desc.fips"));
+        if (isFIPSProfile) {
+            if (debug != null) {
+                debug.println("RestrictedSecurity profile " + profileID
+                        + " is specified as FIPS compliant.");
+            }
+            isFIPSEnabled = true;
+        } else {
+            printStackTraceAndExit("RestrictedSecurity profile " + profileID
+                    + " is not specified as FIPS compliant.");
```
pshipton, line 369: ", but the semeru.fips system propery is set."

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: I restructured the check for the property there and updated the comment to match it.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Added here too.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Added here too.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -593,17 +696,15 @@ private void initProviders() {
...
-                    debug.println(
-                            "Loaded restricted security provider: " + providers.get(pNum - 1)
-                                    + " with simple name: " + providerName);
-                }
             }
 
             if (providers.isEmpty()) {
                 printStackTraceAndExit(
-                        "Restricted security mode provider list empty, or no such restricted security policy in java.security file.");
+                        "No providers were provided as part of the Restricted Security profile.");
```
pshipton, line 703: How about "No providers are specified as part of ..."

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Problem is that in some platforms there is no available solution yet. So, there wouldn't be any custom profile that would allow them to go into FIPS mode. Think I should add that recommendation anyway?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
pshipton: The else case looks good now, but this comment in this place doesn't make sense to me. Maybe if you rewrote it. The next line is checking if the key is a String. The code seem readable enough without any comment here.

KostasTsiounis: It's about the if that comes after, so I can put it below that. I can also remove it altogether though.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
pshipton: You could add something like "Set the semeru.customprofile system property to an available FIPS profile, some platforms does not support FIPS."

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
pshipton: I'm fine with it removed, or if it's closer to the if that it's referring to.

KostasTsiounis: Removed altogether.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Changed to your suggestion.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
pshipton: The message is the error given to the customer, we want to give as much help as possible so they can solve their own problems rather than asking on Slack or opening a support ticket.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
pshipton: "are" instead of "were" is better.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
pshipton: You seem to have removed a comment from the else case, but the comment on line 295 which I was discussing is still there.

pshipton: Your updated comment in the else case was fine, up to you if you want to put it back.

pshipton:
There is a merge conflict now that needs to be resolved.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: It's fine as it is now.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Changed message.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Fixed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
pshipton: Are you going to remove the comment on line 296?

pshipton:
jenkins compile amac jdknext

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Yep. Done.

pshipton:
If all you did was remove the comment, we don't need to rerun the build. Do you need to run any testing on the latest changes? Otherwise I'll go ahead and merge it and you can start backporting.

KostasTsiounis:
No, nothing functional was changed so we should be fine.

pshipton (approved):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -50,41 +54,104 @@ public final class RestrictedSecurity {
...
+        }
+        isArchSupported = false;
+        for (String arch: supportedPlatformsNSS.get("Arch")) {
+            if (props[3].contains(arch)) {
+                isArchSupported = true;
+            }
+        }
+        isNSSSupported = isOsSupported && isArchSupported;
+
+        // Check whether the OpenJCEPlus FIPS solution is supported.
```
pshipton, line 112: This and the next are duplicated from above.

PR #702, opened 2023-11-26
Remove unnecessary interrupt

tajila:
The current implementation contains some code from the j9 threading implementation. With openjdk threading we dont need a deadinterrupt (which the old code is attempting to simulate) field as the interrupt state is tracked in the JCL rather than the JVM.

Also, the old code actually clears the interrupt state which is incorrect.


JasonFengJ9 (approved):
Verified that JDK21 w/ this change fixed the internal workitem 149819.

pshipton:
Created https://github.com/ibmruntimes/openj9-openjdk-jdk21/pull/77

PR #705, opened 2023-12-04
Enable ECDSA ciphers when running with FIPS cryptography

jasonkatonica:
Enable ECDSA ciphers when running with FIPS cryptography

In FIPS mode, certain cryptographic algorithms are typically disabled.
This includes the algorithm `SHA1withECDSA` since `SHA1` is no longer
recommended for general use.

The `JSSE`provider, responsible for managing TLS cipher specifications,
currently disables all `*ECDSA*` ciphers due to the absence of the
`SHA1withECDSA` algorithm on the system when running in FIPS mode.
This behavior is inaccurate, since other ECDSA-related
transformations, such as `SHA512withECDSA` or `SHA384withECDSA`, are
available in a FIPS environment and can be used for TLS connections.

With this update, `*ECDSA*` ciphers can now be enabled in FIPS mode. We
will allow for `*ECDSA*` ciphers to be enabled if any algorithm in the
ECDSA family is present such as `SHA512withECDSA`, `SHA384withECDSA`,
`SHA256withECDSA`, or `SHA224withECDSA`. We expect all these algorithms
to be present in the various FIPS solutions available within Semeru.

### DIFF
--- a/src/java.base/share/classes/sun/security/ssl/JsseJce.java
+++ b/src/java.base/share/classes/sun/security/ssl/JsseJce.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * ===========================================================================
+ */
+
 package sun.security.ssl;
 
 import java.math.BigInteger;
@@ -89,6 +95,26 @@ final class JsseJce {
      * JCA identifier string for ECDSA, i.e. a ECDSA with SHA-1.
      */
     static final String SIGNATURE_ECDSA = "SHA1withECDSA";
+
+    /**
+     * JCA identifier string for ECDSA, i.e. a ECDSA with SHA224.
+     */
+    static final String SIGNATURE_ECDSA_224 = "SHA224withECDSA";
+
+    /**
+     * JCA identifier string for ECDSA, i.e. a ECDSA with SHA256.
+     */
+    static final String SIGNATURE_ECDSA_256 = "SHA256withECDSA";
+
+    /**
+     * JCA identifier string for ECDSA, i.e. a ECDSA with SHA384.
+     */
+    static final String SIGNATURE_ECDSA_384 = "SHA384withECDSA";
+
+    /**
+     * JCA identifier string for ECDSA, i.e. a ECDSA with SHA512.
+     */
+    static final String SIGNATURE_ECDSA_512 = "SHA512withECDSA";
 
     /**
      * JCA identifier for EdDSA signatures.
@@ -162,10 +188,38 @@ final class JsseJce {
         // Is EC crypto available?
         private static final boolean isAvailable;
 
+        /**
+         * Checks if a particular signature algorithm is available.
+         *
+         * @param algorithm the algorithm we will attempt to instantiate to check if it is available
+         * @return true if the signature algorithm is found, false otherwise
+         */
+        private static boolean isSignatureAlgorithmAvailable(String algorithm) {
+            try {
+                // Attempt to create a Cipher instance with the specified algorithm.
+                Signature.getInstance(algorithm);
+                return true;
+            } catch (NoSuchAlgorithmException e) {
+                return false;
+            }
+        }
+
         static {
             boolean mediator = true;
             try {
-                Signature.getInstance(SIGNATURE_ECDSA);
+                // When running in FIPS mode, the signature "SHA1withECDSA" is not
+                // available by default. In this scenario we should still set EC
+                // availability to true since other algorithms in the ECDSA signature
+                // family are available for use in various ECDSA TLS ciphers. All
+                // FIPS solutions are expected to have an algorithm such as
+                // "SHA512withECDSA", "SHA384withECDSA", "SHA256withECDSA", or
+                // "SHA224withECDSA" available so we will also check for these algorithms.
+                mediator = isSignatureAlgorithmAvailable(SIGNATURE_ECDSA)
+                    || isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_224)
+                    || isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_256)
+                    || isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_384)
+                    || isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_512);
+
                 Signature.getInstance(SIGNATURE_RAWECDSA);
                 KeyAgreement.getInstance("ECDH");
                 KeyFactory.getInstance("EC");


pshipton:
The copyright and line endings checks are failing.

keithc-ca (changes requested):

src/java.base/share/classes/sun/security/ssl/JsseJce.java
```diff
@@ -165,7 +178,18 @@ private static class EcAvailability {
         static {
             boolean mediator = true;
             try {
-                Signature.getInstance(SIGNATURE_ECDSA);
+                // When running in FIPS mode the signature SHA1withECDSA is not
```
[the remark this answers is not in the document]
jasonkatonica: Updated using https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/705/commits/636ab7c9f8101d7dff306c3b08ce7b8322156508

src/java.base/share/classes/sun/security/ssl/JsseJce.java
```diff
@@ -165,7 +178,18 @@ private static class EcAvailability {
...
+                // available by default. In this scenario we should still set EC
+                // availability to true since other algorithms in the ECDSA signature
+                // family are available for use in various ECDSA TLS ciphers. All
+                // FIPS solutions are expected to have the algorithm "SHA512withECDSA"
+                // so we will check for this instead when running in FIPS mode.
+                if (RestrictedSecurity.isFIPSEnabled() == true) {
```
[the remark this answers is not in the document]
jasonkatonica: Updated using https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/705/commits/636ab7c9f8101d7dff306c3b08ce7b8322156508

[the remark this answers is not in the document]
jasonkatonica: This seems OK to me. I updated the code using Updated using https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/705/commits/636ab7c9f8101d7dff306c3b08ce7b8322156508 and https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/705/commits/b8f46f838661d9a48e3e3a08356c17ddfaf49bf0

src/java.base/share/classes/sun/security/ssl/JsseJce.java
```diff
@@ -31,6 +37,8 @@
 import java.security.spec.*;
 import javax.crypto.*;
 
+import openj9.internal.security.RestrictedSecurity;
```
keithc-ca, line 40: Please remove this unused import.

src/java.base/share/classes/sun/security/ssl/JsseJce.java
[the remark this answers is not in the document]
keithc-ca: Looking at this again, I see the commit message also makes reference to "SHA384withECDSA"; would it be worthwhile to test all "SHA*withECDSA" algorithms and declare EC support if _any_ of those algorithms are available?

keithc-ca (changes requested):

jasonkatonica:
> Looking at this again, I see the commit message also makes reference to "SHA384withECDSA"; would it be worthwhile to test all "SHA*withECDSA" algorithms and declare EC support if any of those algorithms are available?

Id expect that our current FIPS providers will always have all the SHA*withECDSA algorithms. I can change the logic to check if any of them are present. This too seems like it should be ok to do. I am currently testing commit https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/705/commits/df1e6b148e60d955593a8714b22837765aca02a8

src/java.base/share/classes/sun/security/ssl/JsseJce.java
jasonkatonica: Removed this unused import.

src/java.base/share/classes/sun/security/ssl/JsseJce.java
```diff
@@ -155,6 +181,22 @@ static RSAPublicKeySpec getRSAPublicKeySpec(PublicKey key) {
...
+     */
+    private static boolean isSignatureAlgorithmAvailable(String algorithm) {
+        try {
+            // Attempt to create a Cipher instance with the specified algorithm
+            Signature.getInstance(algorithm);
+            return true;
+        } catch (NoSuchAlgorithmException | javax.crypto.NoSuchPaddingException e) {
+            return false;
+        }
+    }
```
keithc-ca, line 198: This method can be within `EcAvailability` (it's only used there).

I think the description should say "available" rather than "exists" to match the name.

By convention `@` taglets do not end with a period, but the comment in the method should.

`NoSuchPaddingException` is imported, so need not be fully-qualified here.

src/java.base/share/classes/sun/security/ssl/JsseJce.java
```diff
@@ -165,7 +207,24 @@ private static class EcAvailability {
...
+                    // so we will check for this instead when running in FIPS mode.
+                    if (!(isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_224) ||
+                          isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_256) ||
+                          isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_384) ||
+                          isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_512)
+                         )) {
+                            throw NoSuchAlgorithmException("No expected ECDSA algorithms found.");
+                    }
+                }
+
```
keithc-ca, line 227: There's no need to create a new exception to throw, nor even to throw at all, and the try block is not needed with the new helper method; suggest:
```java
                mediator = isSignatureAlgorithmAvailable(SIGNATURE_ECDSA)
                        || isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_224)
                        || isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_256)
                        || isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_384)
                        || isSignatureAlgorithmAvailable(SIGNATURE_ECDSA_512);
```
Please update the comment to reflect that we're not only relying on `SHA512withECDSA` for backup.

keithc-ca (changes requested):
Also, please squash.

src/java.base/share/classes/sun/security/ssl/JsseJce.java
jasonkatonica: Moved the method to the `EcAvailability`

Changed the description to be `available` instead of exists.

Updated the taglets to no longer have a period at the end. Updated the comment in the method to have a period.

I found out during compilation i dont need `NoSuchPaddingException`. This was removed.

src/java.base/share/classes/sun/security/ssl/JsseJce.java
jasonkatonica: Yes this reads better. I updated the code and comment accordingly.

jasonkatonica:
Also squashed this PR.

src/java.base/share/classes/sun/security/ssl/JsseJce.java
```diff
@@ -162,10 +188,38 @@ private static class EcAvailability {
         // Is EC crypto available?
         private static final boolean isAvailable;
 
+        /**
+         * Checks if a particular signature algorithm is available.
+         *
+         * @param algorithm The algorithm we will attempt to instantiate to check if it is available
```
keithc-ca, line 194: nit: "The" -> "the"

src/java.base/share/classes/sun/security/ssl/JsseJce.java
```diff
@@ -162,10 +188,38 @@ private static class EcAvailability {
...
+         * @return true if the signature algorithm is found, false otherwise
+         */
+        private static boolean isSignatureAlgorithmAvailable(String algorithm) {
+            try {
+                // Attempt to create a Cipher instance with the specified algorithm.
+                Signature.getInstance(algorithm);
+                return true;
+            } catch (NoSuchAlgorithmException e) {
```
keithc-ca, line 202: An earlier version of this also caught `NoSuchPaddingException` (which surprised me). Please confirm that we don't need to catch those exceptions.

src/java.base/share/classes/sun/security/ssl/JsseJce.java
```diff
@@ -162,10 +188,38 @@ private static class EcAvailability {
...
             boolean mediator = true;
             try {
-                Signature.getInstance(SIGNATURE_ECDSA);
+                // When running in FIPS mode, the signature "SHA1withECDSA" is not
+                // available by default. In this scenario we should still set EC
+                // availability to true since other algorithms in the ECDSA signature
+                // family are available for use in various ECDSA TLS ciphers. All
+                // FIPS solutions are expected to have an algorithm such as
+                // "SHA512withECDSA", "SHA384withECDSA", "SHA256withECDSA", or
+                // "SHA224withECDSA" available so we will also check for these algorithms.
```
keithc-ca, line 216: The commit message should also be updated to spell "FIPS" consistently and make it clearer (like this comment does) that we don't only check for `SHA1withECDSA` and `SHA512withECDSA`.

src/java.base/share/classes/sun/security/ssl/JsseJce.java
jasonkatonica: Updated The to the.

src/java.base/share/classes/sun/security/ssl/JsseJce.java
jasonkatonica: Yes I did compile this and a `NoSuchPaddingException` is not allowable by the compiler so I removed it.

keithc-ca: I just saw your comment https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/705#discussion_r1414711935.

src/java.base/share/classes/sun/security/ssl/JsseJce.java
jasonkatonica: FIPS is now consistently upper cased. 

The wording for the commit message was also adjusted to list out various ECDSA algorithms.

The PR message was copied from the commit message to force them to match.

keithc-ca: Please fix the first paragraph of the commit message: finish the first sentence with a period and start the second with a capital.
```
In FIPS mode, certain cryptographic algorithms are typically disabled.
This includes the algorithm `SHA1withECDSA` since `SHA1` is no longer
recommended for general use.
```

jasonkatonica: This sentence was updated as per above.

keithc-ca:

keithc-ca:
The test failures with CRIU and virtual threads seem unrelated to this change.

keithc-ca (approved):

PR #708, opened 2023-12-06
Include new "access" shared library in test-image

keithc-ca:
Required by https://github.com/eclipse-openj9/openj9/pull/18571.


pshipton:
jenkins test sanity,extended amac jdk11 depends ibmruntimes/openj9-openjdk-jdk11#736

pshipton:
Yes, it was an accident, I thought I was running it in the other PR. We can ignore this one.

PR #717, opened 2024-01-04
Restore behavior on platforms where CRIU is supported but not in use

WilburZjh:
This PR aims to fix an issue on how to read random data and
get random bytes on platforms where CRIU is supported, but
not in use by enabling it via a --enable-criu-support JVM
option.

Issue https://github.com/eclipse-openj9/openj9/issues/18602

### DIFF
--- a/src/java.base/unix/classes/sun/security/provider/NativePRNG.java
+++ b/src/java.base/unix/classes/sun/security/provider/NativePRNG.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2024 All Rights Reserved
  * ===========================================================================
  */
 
@@ -508,8 +508,9 @@ public final class NativePRNG extends SecureRandomSpi {
                 if (in.readNBytes(data, 0, len) < len) {
                     throw new IOException("Could not read from file(s)");
                 }
+                return;
             }
-            /*[ELSE] CRIU_SUPPORT */
+            /*[ENDIF] CRIU_SUPPORT */
             int ofs = 0;
             while (len > 0) {
                 int k = in.read(data, ofs, len);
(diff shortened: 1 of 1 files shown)


pshipton:
This is to fix https://github.com/eclipse-openj9/openj9/issues/18602

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
```diff
@@ -508,8 +508,8 @@ private static void readFully(InputStream in, byte[] data, Variant variant)
                 if (in.readNBytes(data, 0, len) < len) {
                     throw new IOException("Could not read from file(s)");
                 }
-            }
-            /*[ELSE] CRIU_SUPPORT */
+            } else {
+            /*[ENDIF] CRIU_SUPPORT */
```
keithc-ca, line 512: Suggest using `return;` after line 510 eliminating the need for `else` and lines 525-527 below.

keithc-ca (changes requested):
I think the commit message and the summary here could be improved. Rather than describe *how* the problem was fixed, explain *what* problem is fixed.

keithc-ca:
We'll also want a copy of this (once finalized) for https://github.com/ibmruntimes/openj9-openjdk-jdk22.

WilburZjh:
> The content looks good, but it doesn't appear that you've addressed my concern about the commit message or the summary here. From my perspective, the essence of this is to restore the behavior (actually read random data) on platforms where CRIU is supported, but not in use.

Yes, I just updated the code. Now Im updating the commit message.

pshipton:
I see the commit message was updated. Is this ready to be removed from draft state?

WilburZjh:
> I see the commit message was updated. Is this ready to be removed from draft state?

I rerun the failed tests on the initial code without using the `return`, and the failed tests are passed. Now I am building the JDKs and rerun on the failed cases to ensure that the failed cases are passed.

keithc-ca (approved):
Approved, assuming testing will be successful.

pshipton:
The change looked correct so I went ahead and merged it for jdk21 and we started a Semeru 0.42 M3 build, which is looking (very) good.

WilburZjh:
Build is passed, and the failed cases are passed. Change from draft to ready for review.

@jasonkatonica FYI

pshipton (approved):

PR #720, opened 2024-01-08
JNI type incorrect when fetching PKCS11 NSS modules

jasonkatonica:
When obtaining the list of NSS modules available when running in FIPS
mode a PKCS11 method is called which returns a boolean value. The code
however does not correctly handle the type and is expecting a void
causing some tests to fail that are able to detect this condition.

This update makes the method call while expecting a boolean return
value.


pshipton:
Pls add an IBM copyright statement for the change.

src/jdk.crypto.cryptoki/share/native/libj2pkcs11/j2secmod.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2005, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2024, Oracle and/or its affiliates. All rights reserved.
```
pshipton, line 2: The Oracle copyright should not be modified, an IBM copyright should be added.

jasonkatonica: Added the IBM copyright and also reverted the line that i modified for the Oracle copyright.

jasonkatonica:
> Pls refer to the issue this is fixing in the description.

Added a link to the issue being fixed in both the commit and PR descriptions.

pshipton (approved):

pshipton:
Ideally we should try to get OpenJDK to accept this change so this patch will become obsolete.

keithc-ca (changes requested):
Please update the commit message. The summary should speak to what this change will do if it is merged (as opposed to the problem that exists now).

src/jdk.crypto.cryptoki/share/native/libj2pkcs11/j2secmod.c
```diff
@@ -23,6 +23,11 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ */
```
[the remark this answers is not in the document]
jasonkatonica: Added a blank line below the copyright statement.

jasonkatonica:
> Please update the commit message. The summary should speak to what this change will do if it is merged (as opposed to the problem that exists now).

Updated the commit and PR descriptions with a new approach.

keithc-ca:
The summary line of the commit message still describes the problem instead of the change. Instead of
> JNI signature type incorrect when building NSS modules ArrayList

I suggest it should say something like
> Use correct JNI function to call ArrayList.add()

keithc-ca:
Jenkins compile amac jdknext

keithc-ca:
Please back-port to jdk22, jdk21, jdk17, jdk11 and jdk8.

PR #722, opened 2024-01-10
Add provider name and class name mapping in Restricted Security mode

taoliult:
For the Java security providers, for example, the SunSASL, its class name is com.sun.security.sasl.Provider. From the provider class name, can not get the provider name which defined in its construction method. So, add the mapping between the provider name and its class name in Restricted Security mode.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2024 All Rights Reserved
  * ===========================================================================
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -687,11 +687,19 @@ public final class RestrictedSecurity {
 
                 // Remove the provider's optional arguments if present.
                 pos = providerName.indexOf(' ');
-                providerName = (pos < 0) ? providerName.trim() : providerName.substring(0, pos).trim();
-                // Remove the provider's class package names if present.
-                pos = providerName.lastIndexOf('.');
-                providerName = (pos < 0) ? providerName : providerName.substring(pos + 1, providerName.length());
-                // Provider without arguments and package names.
+                if (pos >= 0) {
+                    providerName = providerName.substring(0, pos);
+                }
+                providerName = providerName.trim();
+
+                // Remove argument, e.g. -NSS-FIPS, if present.
+                pos = providerName.indexOf('-');
+                if (pos >= 0) {
+                    providerName = providerName.substring(0, pos);
+                }
+
+                // Provider name defined in provider construction method.
+                providerName = getProvidersSimpleName(providerName);
                 providersSimpleName.add(pNum - 1, providerName);
             }
 
@@ -963,11 +971,12 @@ public final class RestrictedSecurity {
 
             // Remove argument, e.g. -NSS-FIPS, if present.
             int pos = providerName.indexOf('-');
-            providerName = (pos < 0) ? providerName : providerName.substring(0, pos);
+            if (pos >= 0) {
+                providerName = providerName.substring(0, pos);
+            }
 
-            // Remove the provider class package name if present.
-            pos = providerName.lastIndexOf('.');
-            providerName = (pos < 0) ? providerName : providerName.substring(pos + 1, providerName.length());
+            // Provider name defined in provider construction method.
+            providerName = getProvidersSimpleName(providerName);
 
             // Check if the provider is in restricted security provider list.
             // If not, the provider won't be registered.
@@ -990,6 +999,27 @@ public final class RestrictedSecurity {
                 }
             }
             return false;
+        }
+
+        /**
+         * Get the provider name defined in provider construction method.
+         *
+         * @param providerName provider name or provider with packages
+         * @return provider name defined in provider construction method
+         */
+        private static String getProvidersSimpleName(String providerName) {
+            if (providerName.equals("com.sun.security.sasl.Provider")) {
+                // The main class for the SunSASL provider is com.sun.security.sasl.Provider.
+                return "SunSASL";
+            } else {
+                // Remove the provider's class package names if present.
+                int pos = providerName.lastIndexOf('.');
+                if (pos >= 0) {
+                    providerName = providerName.substring(pos + 1);
+                }
+                // Provider without package names.
+                return providerName;
+            }
         }
 
         /**


pshipton:
Is this targeted to the April release?

taoliult:
> Is this targeted to the April release?

@pshipton We would like to get this into the Feb GA because this is a customer requirement that was opened. And we just get the customer positive feedback today.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -992,6 +982,33 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+        /**
+         * Get the provider name defined in provider construction method.
+         *
+         * @param providerName provider name or provider with packages or arguments
+         * @return provider name defined in provider construction method
+         */
+        private static String getProvidersSimpleName(String providerName) {
+            // Remove the provider's optional arguments if present.
+            int pos = providerName.indexOf(' ');
+            providerName = (pos < 0) ? providerName.trim() : providerName.substring(0, pos).trim();
```
keithc-ca, line 994: I'm not a fan of statements that can effectively assign a variable to itself. Please separate the `substring()` and `trim()` steps:
```java
            if (pos >= 0) {
                providerName = providerName.substring(0, pos);
            }
            providerName = providerName.trim();
```
Similarly line 998 should be:
```java
            if (pos >= 0) {
                providerName = providerName.substring(0, pos);
            }
```
and line 1006 should be:
```java
                if (pos >= 0) {
                    providerName = providerName.substring(0, pos + 1);
                }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -992,6 +982,33 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+
+            // Remove argument, e.g. -NSS-FIPS, if present.
+            pos = providerName.indexOf('-');
+            providerName = (pos < 0) ? providerName : providerName.substring(0, pos);
+
+            if (providerName.equals("com.sun.security.sasl.Provider")) {
+                // The main class for the SunSASL provider is com.sun.security.sasl.Provider
```
keithc-ca, line 1001: Please finish each sentence with a period.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -992,6 +982,33 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                return "SunSASL";
+            } else {
+                // Remove the provider's class package names if present.
+                pos = providerName.lastIndexOf('.');
+                providerName = (pos < 0) ? providerName : providerName.substring(pos + 1);
+                // Provider without arguments and package names.
```
keithc-ca, line 1007: nit: "and" should be "or" here.

keithc-ca (changes requested):
How can we be sure there are no other providers that need treatment similar to SunSASL?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -685,13 +685,8 @@ private void initProviders() {
...
 
-                // Remove the provider's optional arguments if present.
-                pos = providerName.indexOf(' ');
-                providerName = (pos < 0) ? providerName.trim() : providerName.substring(0, pos).trim();
-                // Remove the provider's class package names if present.
-                pos = providerName.lastIndexOf('.');
-                providerName = (pos < 0) ? providerName : providerName.substring(pos + 1, providerName.length());
-                // Provider without arguments and package names.
+                // Provider name defined in provider construction method.
+                providerName = getProvidersSimpleName(providerName);
```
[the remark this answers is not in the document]
taoliult: Previously, we did not adding a new method for trimming the provider name when there are packages or arguments. But when we back port the Restricted Security mode to JDK8, some providers name(name defined in the construction method) and its class name in JDK8 are not same, so only from the class name, we can not get the provider name. Then, we add the new method(getProvidersSimpleName) for mapping the provider name and its class name. And during those code reviews on JDK8, you suggested, we can move those provider name trimming codes into this new method. So, that is why, those trimming codes and provider name, class name mapping codes, are in this new method. 

But previously, these codes are only exist in JDK8. For this PR, we have to add the provider name and its class name mapping for provider SunSASL. So, I copy those same codes from JDK8 to this JDKNext PR, and also adding the mapping for SunSASL.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -685,13 +685,8 @@ private void initProviders() {
                 // Provider with argument (provider name + optional argument).
                 providers.add(pNum - 1, providerName);
```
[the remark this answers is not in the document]
taoliult: SunSASL is not a FIPS provider, so we can not, by default, adding "RestrictedSecurity.NSS.140-2.jce.provider.5 = SunSASL" into java.security file. Clients need to make their own decision to add this SunSASL or not. We can update the FIPS document to mention this.

taoliult:
> How can we be sure there are no other providers that need treatment similar to SunSASL?

Ideally, in FIPS mode, we cannot allow the non-FIPS providers in the FIPS provider list. In Semeru NSS FIPS, only SunPKCS11-NSS-FIPS, Sun, SunEC and SunJSSE, these four providers can be allowed. So, in this ideal case, after JDK11, we don’t need such mapping in Restricted Security mode codes. 

But, the clients always have requirements, to use the non-FIPS providers in the Sememe NSS FIPS mode. Clients need to make their own decision to make sure their environment is FIPS compliance, when they want to use non-FIPS provider.

In this case, for the providers which the provider name(defined in the construction method) and its class name are not same, there is no way for us to get the provider name only from its class name, before initializing the provider (In the Restricted Security mode review, it requires, the non-FIPS provider can not be initialized). So, we have to add this method for the mapping between the provider name and its class name, if they are not the same. For the providers which the provider name and its class name are same, there is no such issue.

And unfortunately, we can’t be sure, there are no other providers that need similar treatment, if the provider name and its class name are not same.

taoliult:
@keithc-ca Codes updated according to code review. Please help to review and advise.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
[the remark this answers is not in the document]
keithc-ca: Suppose a client includes either of these lines in `java.security`:
```
RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.jce.provider.4 = SunSASL-NSS-FIPS
RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.jce.provider.4 = com.sun.security.sasl.Provider
```
If I understand the intent of this change, either should result in
```
security.provider.4=SunSASL
```
being stored in the `Properties` object by `RestrictedSecurity.setProviders()`, but the code currently will not do that. Should it?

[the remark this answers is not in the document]
keithc-ca: It's not clear that you have answered my questions:
* Given a provider name with an argument (e.g. `-NSS_FIPS`) should it be removed or not when constructing the value of `security.provider.N`?
* If the provider name differs from the class name (like `com.sun.security.sasl.Provider`) should the value of `security.provider.N` be the provider name or the class name?

Comments in `java.security` discuss the different situations where the provider name vs. the class name are used. Did you test both?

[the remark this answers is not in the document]
taoliult: Given a provider name with an argument (e.g. -NSS_FIPS), we need to remove the argument when add it into variable `private final List<String> providersSimpleName;`. 

For creating the `security.provider.N` list, it will use the variable `private final List<String> providers`, which comes from the "`RestrictedSecurity.NSS.140-2.jce.provider.N`". So, if "`RestrictedSecurity.NSS.140-2.jce.provider.N=provider name`", then use provider name. If "`RestrictedSecurity.NSS.140-2.jce.provider.N=class name`", then use class name.

If the provider name differs from the class name (like com.sun.security.sasl.Provider), when creating the `security.provider.N` list, same as above first question, it will use the variable `private final List<String> providers`, which comes from the "`RestrictedSecurity.NSS.140-2.jce.provider.N`".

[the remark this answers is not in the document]
keithc-ca: I understand how `providers` and `providersSimpleName` are used. You still haven't answered my question. If this change is to have any value, a user will need to specify
```
RestrictedSecurity.some-profile.jce.provider.N = something
```
What should "_something_" be? Should it be `SunSASL` or `com.sun.security.sasl.Provider`?

[`java.security`](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/src/java.base/share/conf/security/java.security#L75) suggests that `security.provider.N=SunSASL` is appropriate, so if "_something_" is `com.sun.security.sasl.Provider` we don't get the desired result.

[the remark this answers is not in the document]
taoliult: It should be `SunSASL`.  Example as follow:

`RestrictedSecurity.NSS.140-2.jce.provider.5 = SunSASL`

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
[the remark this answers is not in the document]
keithc-ca: I'm not sure you actually answered my question. I now think it was an oversight that the code did not previously trim the argument to form the simple name. Please confirm.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -992,6 +982,40 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+         * @param providerName provider name or provider with packages or arguments
+         * @return provider name defined in provider construction method
+         */
+        private static String getProvidersSimpleName(String providerName) {
+            // Remove the provider's optional arguments if present.
+            int pos = providerName.indexOf(' ');
+            if (pos >= 0) {
+                providerName = providerName.substring(0, pos);
+            }
+            providerName = providerName.trim();
```
keithc-ca, line 997: I don't think this belongs here. Removing optional arguments should only occur in `initProviders()` so it does not alter the behavior of `isRestrictedProviderAllowed(String providerName)` in unwanted ways (if "SunPKCS11" is allowed, that doesn't mean "SunPKCS11 [some random text]" should be accepted there).

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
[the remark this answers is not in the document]
taoliult: I wouldn’t say it was an oversight that the code did not previously trim the argument (after -). Because, in the `initProviders()` method, it didn’t have the argument trim for `“-”`, that is because no need to trim `“-”`. The provider lists which the initProviders() method will deal with are as below:

```
RestrictedSecurity.NSS.140-2.jce.provider.1 = SunPKCS11 ${java.home}/conf/security/nss.fips.cfg
RestrictedSecurity.NSS.140-2.jce.provider.2 = SUN [{CertificateFactory, X.509, ImplementedIn=Software}, \
    {CertStore, Collection, ImplementedIn=Software}, \
    {CertStore, com.sun.security.IndexedCollection, ImplementedIn=Software}, \
    {Policy, JavaPolicy, *}, {Configuration, JavaLoginConfig, *}, \
    {CertPathBuilder, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}, \
    {CertPathValidator, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}, \
    {KeyStore, PKCS12, *}]
RestrictedSecurity.NSS.140-2.jce.provider.3 = SunEC [{KeyFactory, EC, ImplementedIn=Software: \
RestrictedSecurity.NSS.140-2.jce.provider.4 = SunJSSE
```

The provider `SunPKCS11`, before it initialized using the configure file ${java.home}/conf/security/nss.fips.cfg. Its name is `SunPKCS11`, not `SunPKCS11-NSS-FIPS` yet.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: Ok, I will move the below codes, back to the method `initProviders()`

```
            // Remove the provider's optional arguments if present.
            int pos = providerName.indexOf(' ');
            if (pos >= 0) {
                providerName = providerName.substring(0, pos);
            }
            providerName = providerName.trim();
```

And do you think the below trim for "-", will also need to move back to the `isRestrictedProviderAllowed()` ?

            // Remove argument, e.g. -NSS-FIPS, if present.
            pos = providerName.indexOf('-');
            if (pos >= 0) {
                providerName = providerName.substring(0, pos);
            }

keithc-ca: I think the removal of an argument (e.g. `-NSS-FIPS`) belongs in `initProviders()` as well, before the call to `getProvidersSimpleName()` instead of within `getProvidersSimpleName()`.

I don't think `isRestrictedServiceAllowed(Service service)` should be expecting arguments and so should not include code to remove them. However, it should be calling `getProvidersSimpleName()` so `SunSASL` is handled properly.

taoliult: > I think the removal of an argument (e.g. `-NSS-FIPS`) belongs in `initProviders()` as well, before the call to `getProvidersSimpleName()` instead of within `getProvidersSimpleName()`.
> 

The removal of an argument (e.g. -NSS-FIPS) can belongs in initProviders(). I will add the check before the call to `getProvidersSimpleName()` instead of within `getProvidersSimpleName()`.

> I don't think `isRestrictedServiceAllowed(Service service)` should be expecting arguments and so should not include code to remove them. However, it should be calling `getProvidersSimpleName()` so `SunSASL` is handled properly.

In the `isRestrictedServiceAllowed(Service service)`, it will get the provider name by the following code:

`String providerName = service.getProvider().getName();`

This `providerName` will always be the provider name defined in its construction method, for the `SunSASL` provider, it will alway return `SunSASL` as the `providerName`. So, I don't think it need to call the `getProvidersSimpleName()`.

keithc-ca: > In the `isRestrictedServiceAllowed(Service service)`, it will get the provider name by the following code:
> 
> `String providerName = service.getProvider().getName();`
> 
> This `providerName` will always be the provider name defined in its construction method, for the `SunSASL` provider, it will alway return `SunSASL` as the `providerName`. So, I don't think it need to call the `getProvidersSimpleName()`.

Agreed, but I think the code to remove arguments should be removed from `isRestrictedServiceAllowed(Service service)`.

taoliult: > > In the `isRestrictedServiceAllowed(Service service)`, it will get the provider name by the following code:
> > `String providerName = service.getProvider().getName();`
> > This `providerName` will always be the provider name defined in its construction method, for the `SunSASL` provider, it will alway return `SunSASL` as the `providerName`. So, I don't think it need to call the `getProvidersSimpleName()`.
> 
> Agreed, but I think the code to remove arguments should be removed from `isRestrictedServiceAllowed(Service service)`.

Here are the codes in isRestrictedServiceAllowed(Service service)

```
        boolean isRestrictedServiceAllowed(Service service) {
            String providerName = service.getProvider().getName();

            // Provider with argument, remove argument.
            // e.g. SunPKCS11-NSS-FIPS, remove argument -NSS-FIPS.
            int pos = providerName.indexOf('-');
            providerName = (pos < 0) ? providerName : providerName.substring(0, pos);

            Constraint[] constraints = providerConstraints.get(providerName);
            .....
```

I don't think the remove argument (e.g. remove -NSS-FIPS) can be removed from `isRestrictedServiceAllowed(Service service)`.

Because, for the provider like `SunPKCS11`, the provider name comes from two parts, one is the provider name defined in its construction method, which is `SunPKCS11`. Another part comes from its configuration file `${java.home}/conf/security/nss.fips.cfg`, which is` NSS-FIPS`, listed below.

```
name = NSS-FIPS
nssLibraryDirectory = /usr/lib64
nssSecmodDirectory = /etc/pki/nssdb
nssDbMode = readOnly
nssModule = fips
```

So, once the `SunPKCS11` is initialized using its configuration file, its provider name will be `SunPKSC11-NSS-FIPS`.

But when we generate the `providersSimpleName` list in method `initProviders()`. The provider name we get from the `java.security` file, which is 

`RestrictedSecurity.NSS.140-2.jce.provider.1 = SunPKCS11 ${java.home}/conf/security/nss.fips.cfg`. 

At that time, the provider is not initialized yet, and there is no way we can know the provider name will be `SunPKCS11-NSS-FIPS` at that time. So, we remove the configure file part ( which is `${java.home}/conf/security/nss.fips.cfg`) and use `SunPKCS11` as the `providersSimpleName` and store it into the `providersSimpleName` list.

But in the `isRestrictedServiceAllowed(Service service)`, the `providerName` returned in this method, is the provider which is already initialized, so the `providerName` will be `SunPKCS-NSS-FIPS`.  So, that is why we need the remove argument (e.g. remove -NSS-FIPS) here in this `isRestrictedServiceAllowed(Service service)`.

keithc-ca: I thought I saw a provider with a hyphen in its name, but when I looked again, I couldn't find it. So, yes, you're right `isRestrictedServiceAllowed(Service service)` should not change.

taoliult: Ok, then, I will move the below codes, out of `getProvidersSimpleName()`,  back to the method `initProviders()`

```
            // Remove the provider's optional arguments if present.
            int pos = providerName.indexOf(' ');
            if (pos >= 0) {
                providerName = providerName.substring(0, pos);
            }
            providerName = providerName.trim();

            // Remove argument, e.g. -NSS-FIPS, if present.
            pos = providerName.indexOf('-');
            if (pos >= 0) {
                providerName = providerName.substring(0, pos);
            }
```

And also, move the below code, back to the method `isRestrictedProviderAllowed()`.

            // Remove argument, e.g. -NSS-FIPS, if present.
            pos = providerName.indexOf('-');
            if (pos >= 0) {
                providerName = providerName.substring(0, pos);
            }

taoliult: @keithc-ca Codes updated according to above. Please help to review and advise.

keithc-ca:

keithc-ca (approved):

PR #724, opened 2024-01-10
Account for different NSS error code in s390x

KostasTsiounis:
A difference in behaviour arises if the `SunPKCS11` provider, and the underlying `NSS` library, is used for verifying certificates in `s390x`. 

When the signature is incorrect and the verification of a certificate fails, the library returns a `CKR_DATA_LEN_RANGE` instead of a `CKR_SIGNATURE_LEN_RANGE` error.

The test is updated to account for this difference.

Issue https://github.com/eclipse-openj9/openj9/issues/18562

### DIFF
--- a/test/jdk/sun/security/pkcs11/ec/ReadCertificates.java
+++ b/test/jdk/sun/security/pkcs11/ec/ReadCertificates.java
@@ -22,6 +22,12 @@
  */
 
 /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ */
+
+/*
  * @test
  * @bug 6405536 6414980 8051972
  * @summary Make sure that we can parse certificates using various named curves
@@ -41,6 +47,7 @@ import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.security.NoSuchProviderException;
 import java.security.Provider;
+import java.security.ProviderException;
 import java.security.PublicKey;
 import java.security.SecureRandom;
 import java.security.SignatureException;
@@ -182,6 +189,16 @@ public class ReadCertificates extends PKCS11Test {
                 }
             } catch (SignatureException | InvalidKeyException e) {
                 System.out.println("OK: " + e);
+            } catch (ProviderException pe) {
+                if (pe.getMessage().contains("cancel failed")
+                    && "s390x".equals(System.getProperty("os.arch"))
+                ) {
+                    System.out.println("NSS error code is different in s390x.");
+                    pe.printStackTrace();
+                    return;
+                }
+
+                throw pe;
             }
         }
 


pshipton:
Pls add the IBM copyright.

test/jdk/sun/security/pkcs11/ec/ReadCertificates.java
```diff
@@ -55,6 +62,7 @@
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+
```
keithc-ca, line 65: Please don't make any unnecessary changes like adding this blank line.

KostasTsiounis: Changed back.

pshipton:
When trying to reproduce the failure on jdk23 we discovered it already appears to be fixed in OpenJDK from 21. The better solution is to backport the fix to 17, and earlier if applicable.

keithc-ca:
> appears to be fixed in OpenJDK from 21

Is that "fix" a change to the signature validation code or a change to the test?
Can someone share a link to that fix, please?

pshipton:
It's a code change. Kostas mentioned https://github.com/openjdk/jdk21/commit/83e6a4c0e9e4a9474ae0c1252378b1a09d1d2df0 but I don't think we should all of that. Kostas is looking at it. This could be closed but I left it open in draft until we figure it out.

pshipton:
Kostas discovered the real reason the test passes in jdk21+ is because it's skipped.
https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/8366eeab847e3c5eb4ea89ed0f86d1cabb6e8678

keithc-ca:
I agree with Peter's [comment](https://github.com/eclipse-openj9/openj9/issues/18562#issuecomment-1858179332) that the fix needs to be in the code that a user might invoke, not changing the test to tolerate the difference.

pshipton:
I don't want to rush into a code fix that we'd be stuck maintaining. We could have backported an existing fix, but since an OpenJDK code fix doesn't seem to exist, the original plan was to consult with the nss project to see if the fix should be done there for z, since z seems to have different behavior from x, p. In the short term we probably want this change in jdk17, jdk21 to avoid test failures for a known problem. We don't need this change in jdk22+ because of https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/8366eeab847e3c5eb4ea89ed0f86d1cabb6e8678, which added the requirement to set `-Djdk.test.lib.artifacts.nsslib-<platform>=<nss location>` in order for the tests to run (otherwise they are skipped).

i.e.
`jtreg.SkippedException: Warning: unsupported OS: Linux-s390x-64, please initialize NSS library location, skipping test`
`jtreg.SkippedException: Warning: unsupported OS: Linux-amd64-64, please initialize NSS library location, skipping test`

Even for jdk21 and earlier the test can be skipped, I think only on certain OSes where the nss location is not known, or if nss is not installed.
```
Beginning test run ReadCertificates...
Warning: can't find NSS librarys on this machine, skipping test
```

keithc-ca:
I wasn't suggest we rush anything.

If we don't need this in jdk22+ should we simply close this?

pshipton:
Created https://github.com/adoptium/aqa-tests/issues/4962 for the test issue.

PR #727, opened 2024-01-16
Remove z/OS from platforms copying static jvm/jsig libaries into image

joransiu:
Remove z/OS from platforms copying static jvm/jsig libaries into image


keithc-ca (commented):
This isn't about building those libraries, but rather about including them in an image.

joransiu:
> This isn't about building those libraries, but rather about including them in an image.

You're right.  I've updated the commit message and PR title to reflect this.

keithc-ca (approved):

PR #736, opened 2024-01-23
[FFI/JDKnext] Perf improvement via NativeMethodHandle/linkToNative

ChengJin01:
The changes aim to generate the NativeMethodHandle instance to ensure the invocation path goes via linkToNative.


ChengJin01:
Reviewer: @tajila
FYI: @keithc-ca, @pshipton

src/java.base/share/classes/java/lang/invoke/MethodHandleImpl.java
```diff
@@ -25,15 +25,14 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
```
keithc-ca, line 28: Copyrights should say 2024.

ChengJin01: Fixed everywhere.

src/java.base/share/classes/java/lang/invoke/NativeMethodHandle.java
```diff
@@ -23,34 +23,48 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
```
keithc-ca, line 28: The first year was correct: you started this change in 2023; same for `JavaLangInvokeAccess`.

src/java.base/share/classes/java/lang/invoke/NativeMethodHandle.java
```diff
@@ -74,11 +88,12 @@ private static boolean hasIllegalType(MethodType type) {
...
-              || pType == float.class
-              || pType == double.class
-              || pType == void.class
-              || pType == Object.class);
+                || pType == int.class
+                || pType == float.class
+                || pType == double.class
+                || pType == void.class
+                || pType == MemorySegment.class // Skip checking the argument preset in OpenJ9 for the moment.
+                || pType == SegmentAllocator.class); // Skip checking the argument preset OpenJ9 for the moment.
```
keithc-ca, line 96: Please retain the indentation from upstream.

src/java.base/share/classes/java/lang/invoke/MethodHandleImpl.java
```diff
@@ -25,15 +25,14 @@
...
- * (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2024 All Rights Reserved
  * ===========================================================================
  */
 
 package java.lang.invoke;
 
 import jdk.internal.access.JavaLangInvokeAccess;
 import jdk.internal.access.SharedSecrets;
-import jdk.internal.foreign.abi.NativeEntryPoint;
```
keithc-ca, line 36: I think you should leave this alone; it isn't used, but there's no harm keeping it.

src/java.base/share/classes/java/lang/invoke/NativeMethodHandle.java
```diff
@@ -23,34 +23,48 @@
...
+ * ===========================================================================
+ */
+
 package java.lang.invoke;
 
 import jdk.internal.vm.annotation.ForceInline;
-import jdk.internal.foreign.abi.NativeEntryPoint;
+
+import java.lang.foreign.MemorySegment;
+import java.lang.foreign.SegmentAllocator;
```
keithc-ca, line 37: I think these imports belong before line 34.

src/java.base/share/classes/java/lang/invoke/NativeMethodHandle.java
ChengJin01: Fixed.

src/java.base/share/classes/java/lang/invoke/NativeMethodHandle.java
ChengJin01: Adjusted as suggested above.

src/java.base/share/classes/java/lang/invoke/MethodHandleImpl.java
ChengJin01: Agreed and fixed.

src/java.base/share/classes/java/lang/invoke/NativeMethodHandle.java
ChengJin01: Fixed.

src/java.base/share/classes/java/lang/invoke/NativeMethodHandle.java
```diff
@@ -23,34 +23,48 @@
...
  * which is used to capture the characteristics of the native call (such as calling convention to be used,
  * or whether a native transition is required) and a <em>fallback</em> method handle, which can be used
  * when intrinsification of this method handle is not possible.
  */
 /*non-public*/ final class NativeMethodHandle extends MethodHandle {
-    final NativeEntryPoint nep;
+    /* The native entry point is literally the bounded MethodHandle implemented in OpenJ9. */
+    final MethodHandle nep;
+
+    /* The cache array for the bounded MethodHandle stores MemberName and appendix which are populated
```
keithc-ca, line 54: In these comments, I think "bound" is preferable to "bounded".

ChengJin01: Fixed everywhere.

keithc-ca (changes requested):

src/java.base/share/classes/java/lang/invoke/NativeMethodHandle.java
```diff
@@ -78,7 +92,9 @@ private static boolean isIllegalType(Class<?> pType) {
               || pType == float.class
               || pType == double.class
               || pType == void.class
-              || pType == Object.class);
+              || pType == Object.class
+              || pType == MemorySegment.class // Skip checking the argument preset in OpenJ9 for the moment.
+              || pType == SegmentAllocator.class); // Skip checking the argument preset OpenJ9 for the moment.
```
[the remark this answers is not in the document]
ChengJin01: > Please explain why these classes need to be allowed here. 

The leading 3 arguments of runNativeMethod() (`MemorySegment downcallAddr, SegmentAllocator segmtAllocator, 
MemorySegment stateSegmt`) at https://github.com/eclipse-openj9/openj9/blob/7bbeb0f43dcf2ac8677928512e1c134b14ae4f46/jcl/src/java.base/share/classes/openj9/internal/foreign/abi/InternalDowncallHandler.java#L712
```
	Object runNativeMethod(MemorySegment downcallAddr, SegmentAllocator segmtAllocator, <------
                                                  MemorySegment stateSegmt,  <------
long[] args) throws IllegalArgumentException, IllegalStateException {
...
}
```
must be allowed to ensure it passes the type check on the arguments of LambdaForm-generated bytecode in OpenJDK.

>  The comments suggest it's temporary: when will the exceptions be removed?

The prototype only supports primitives for the time being as we have not yet determined whether they must always be there for struct/union (which won't be supported until we figure out everything with primitives) from VM & JIT perspective and how to deal with this situation. So we will need to leave them in there till everything is implemented in our code and get back to check whether they need to be removed or not at that time.

src/java.base/share/classes/java/lang/invoke/NativeMethodHandle.java
```diff
@@ -23,34 +23,48 @@
...
+ * ===========================================================================
+ */
+
 package java.lang.invoke;
 
+import java.lang.foreign.MemorySegment;
+import java.lang.foreign.SegmentAllocator;
+
 import jdk.internal.vm.annotation.ForceInline;
-import jdk.internal.foreign.abi.NativeEntryPoint;
```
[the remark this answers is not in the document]
ChengJin01: Fixed.

src/java.base/share/classes/java/lang/invoke/NativeMethodHandle.java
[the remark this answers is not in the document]
keithc-ca: Perhaps the exceptions should only be made for parameter types and not also the return type by modifying `hasIllegalType()` instead?
```java
    private static boolean hasIllegalType(MethodType type) {
        if (isIllegalType(type.returnType()))
            return true;

        for (Class<?> pType : type.ptypes()) {
            // Specifically allow these parameter types for OpenJ9 for the moment.
            if (pType == MemorySegment.class || pType == SegmentAllocator.class) {
                continue;
            }
            if (isIllegalType(pType))
                return true;
        }

        return false;
    }
```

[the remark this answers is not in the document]
ChengJin01: That's correct for the moment but won't be necessary as ultimately the check on `MemorySegment` & `SegmentAllocator` should be moved to `isIllegalType()` (the same as the current changes) once we confirm everything in the calling path of the MH invocation from VM & JIT perspective.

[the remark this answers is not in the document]
keithc-ca: Based on your answer, we don't need to allow the return type to be `MemorySegment` or `SegmentAllocator`, so it doesn't belong in `isIllegalType()`.

Perhaps we should be even more restrictive, allowing those types only for parameters of `runNativeMethod()`?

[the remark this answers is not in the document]
ChengJin01: >  we don't need to allow the return type to be MemorySegment or SegmentAllocator, so it doesn't belong in isIllegalType().

`MemorySegment` is used for the returned `struct/union` (which we have not ye supported for now)

`SegmentAllocator` might be removed from the return type check (as it is intended to allocate the memory for `struct/union`)

Both types should be reserved for the parameter type check as they are specific to `runNativeMethod`.

> Perhaps we should be even more restrictive, allowing those types only for parameters of runNativeMethod()?

No need to worry about this given the passed-in nep (namely the bound MH) already indicates the method is `runNativeMethod`.

Overall, we don't have to emphasize the difference of the parameter & return type on the type check so as to minimize the changes to adapt them to the upstream in OpenJDK; otherwise, we have to treat them separately with different methods for the parameter & return type.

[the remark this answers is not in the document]
keithc-ca: My concern is that `hasIllegalType()` is used for other situations where we don't want those extra types to be allowed.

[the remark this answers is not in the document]
ChengJin01: There is no other case to do the type check via `hasIllegalType()` except `nativeMH.make()` (actually check  the types in `runNativeMethod` for the bound MH so as to generate the native MH) given the `nativeMH/LinkToNative` based solution is only specific to OpenJ9 rather then OpenJDK//Hotspot unless the related code are changed in the future in OpenJDK (in which case we will have to figure out other way around this situation).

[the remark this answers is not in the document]
keithc-ca: You haven't convince me that the additional types should be allowed for the return-type of the method handle being checked.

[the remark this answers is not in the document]
ChengJin01: I have no problem with the suggestion at https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/736#discussion_r1466528442 give the return type of `runNativeMethod` is `Object` (will be converted to the specific value against the actual return type) in our implementation.
```
Object runNativeMethod(MemorySegment downcallAddr, SegmentAllocator segmtAllocator, ...
```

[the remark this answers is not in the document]
ChengJin01: Updated the code against the suggestion at https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/736#discussion_r1466528442.

keithc-ca (commented):
This change looks good, but the "JDKnext" part of the title and commit message seems redundant (it's implied by the repo).

ChengJin01:
> This change looks good, but the "JDKnext" part of the title and commit message seems redundant (it's implied by the repo).

Removed `JDKnext` as suggested.

tajila (approved):

keithc-ca:
Jenkins test sanity amac,zlinux,win jdknext depends https://github.com/eclipse-openj9/openj9/pull/18799

ChengJin01:
The test failure with `LoadLibrary` on Windows at https://openj9-jenkins.osuosl.org/job/Test_openjdknext_j9_sanity.functional_x86-64_windows_Personal/21/tapResults/ has nothing to do with our changes.
```
TESTING:
*** Starting test suite: LoadLibrary test intended for CMVC 201408 ***
Testing: LoadLibrary testcase with invalid paths specified in the library path list
Test start time: 2024/02/01 19:39:36 Eastern Standard Time
Running command: C:/Users/jenkins/workspace/Test_openjdknext_j9_sanity.functional_x86-64_windows_Personal_testList_1/jdkbinary/j2sdk-image\bin\java -Djava.library.path=".;'C:\\Program Files (x86)\\ibm\\tivoli';C:\\Users\\j9build;C:\\dev;." -cp C:/Users/jenkins/workspace/Test_openjdknext_j9_sanity.functional_x86-64_windows_Personal_testList_1/aqa-tests///..//jvmtest\functional\cmdLineTests\loadLibraryTest\loadLibraryTest.jar org.openj9.test.loadLibrary.TestLoadLibrary
Time spent starting: 78 milliseconds
Time spent executing: 1093 milliseconds
Test result: FAILED
```

keithc-ca:
The only test failure was in `cmdLineTester_loadLibraryTests`, but I don't think that's related to this change.

keithc-ca (approved):

PR #737, opened 2024-01-24
Update Java Restricted Security Mode comment in java.security

taoliult:
The Java Restricted Security Mode comment in java.security file should be presented on all platforms that have either NSS FIPS 140-2 or OpenJCEPlus FIPS 140-3 supported.

### DIFF
--- a/src/java.base/share/conf/security/java.security
+++ b/src/java.base/share/conf/security/java.security
@@ -85,10 +85,12 @@ security.provider.tbd=Apple
 #endif
 security.provider.tbd=SunPKCS11
 
-#if defined linux-x86 || defined linux-ppc || defined linux-s390
+#if defined aix-ppc || defined linux-ppc || defined linux-s390 || defined linux-x86 || defined windows
 #
 # Java Restricted Security Mode
 #
+#endif
+#if defined linux-ppc || defined linux-s390 || defined linux-x86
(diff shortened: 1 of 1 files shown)


src/java.base/share/conf/security/java.security
```diff
@@ -85,10 +85,11 @@ security.provider.tbd=Apple
 #endif
 security.provider.tbd=SunPKCS11
 
-#if defined linux-x86 || defined linux-ppc || defined linux-s390
+#if defined aix-ppc || defined linux-ppc || defined linux-s390 || defined linux-x86 || defined windows
 #
 # Java Restricted Security Mode
 #
+#if defined linux-ppc || defined linux-s390 || defined linux-x86
```
keithc-ca, line 92: With this addition should there not also be an added `#endif`?

keithc-ca: `MakeJavaSecurity` doesn't appear to support nested conditions, so I think that new `#endif` belongs before line 92 so conditions are not nested here. Otherwise I think it's misleading (the similarity with the C preprocessor might lead readers to think nesting is supported leading to a misunderstanding of meaning here).

taoliult: Yes, the `#endif` is needed here. I updated the codes and right now, rebuilding it on Jenkins.

keithc-ca (approved):

JinhangZhang:
Backport to Semeru JDK11 PR https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/761

PR #742, opened 2024-02-08
Fix comment and static method use

keithc-ca:


JasonFengJ9 (approved):
Validated in a grinder https://openj9-jenkins.osuosl.org/job/Grinder/3288

PR #744, opened 2024-02-13
Avoid overlapping buffers in native ChaCha20

jasonkatonica:
When using the ChaCha20 algorithms and an input
and output buffer overlaps we should ensure that
a copy of the input buffer is made before
encrypting or decrypting. This ensures that
OpenSSL is able to process this data. OpenSSL
does not allow for overlapping input and output
buffers when performing operations on data for the ChaCha20 algorithm.

The values returned when getting the output
size for a crypto operation were also found to be
incorrect. This update matches the logic that is
within the ChaCha20Cipher.java file for
calculating the output sizes.

Encoded key material was found to be left in
memory under the right conditions. This memory
should be zeroed to avoid a copy of the key from
residing in memory for longer than necessary.
This addition was noticed when comparing the
NativeChaCha20Cipher class to the ChaCha20Cipher
class. The method getEncodedKey now will zero out
the copy of the key before throwing an exception.

Each of these changes fix the two failing tests
reported in issue [18703](https://github.com/eclipse-openj9/openj9/issues/18703).

### DIFF
--- a/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
+++ b/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2024 All Rights Reserved
  * ===========================================================================
  */
 package com.sun.crypto.provider;
@@ -214,20 +214,7 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
      */
     @Override
     protected int engineGetOutputSize(int inputLen) {
-        int outLen = 0;
-
-        if (mode == MODE_NONE) {
-            outLen = inputLen;
-        } else if (mode == MODE_AEAD) {
-            if (direction == Cipher.ENCRYPT_MODE) {
-                outLen = Math.addExact(inputLen, TAG_LENGTH);
-            } else {
-                outLen = Integer.max(inputLen, TAG_LENGTH) - TAG_LENGTH;
-                outLen = Math.addExact(outLen, engine.getCipherBufferLength());
-            }
-        }
-
-        return outLen;
+        return engine.getOutputSize(inputLen, true);
     }
 
     /**
@@ -659,6 +646,9 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
         }
         byte[] encodedKey = key.getEncoded();
         if (encodedKey == null || encodedKey.length != 32) {
+            if (encodedKey != null) {
+                Arrays.fill(encodedKey, (byte)0);
+            }
             throw new InvalidKeyException("Key length must be 256 bits");
         }
         return encodedKey;
@@ -677,11 +667,11 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
      */
     @Override
     protected byte[] engineUpdate(byte[] in, int inOfs, int inLen) {
-        byte[] out = new byte[inLen];
+        byte[] out = new byte[engine.getOutputSize(inLen, false)];
         try {
             int size = engine.doUpdate(in, inOfs, inLen, out, 0);
             // Special case for EngineAEADDec, doUpdate only buffers the input
-            // So the output array must be empty since no encryption happened yet
+            // so the output array must be empty since no encryption has happened yet.
             if (size == 0) {
                 return new byte[0];
             }
@@ -740,7 +730,7 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
     @Override
     protected byte[] engineDoFinal(byte[] in, int inOfs, int inLen)
             throws AEADBadTagException {
-        byte[] output = new byte[engineGetOutputSize(inLen)];
+        byte[] output = new byte[engine.getOutputSize(inLen, true)];
         try {
             engine.doFinal(in, inOfs, inLen, output, 0);
         } catch (ShortBufferException | KeyException exc) {
@@ -890,6 +880,17 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
      */
     interface ChaChaEngine {
         /**
+         * Size an output buffer based on the input and where applicable
+         * the current state of the engine in a multipart operation.
+         *
+         * @param inLength the input length.
+         * @param isFinal true if this is invoked from a doFinal call.
+         *
+         * @return the recommended size for the output buffer.
+         */
+        int getOutputSize(int inLength, boolean isFinal);
+
+        /**
          * Perform a multi-part update for ChaCha20.
          *
          * @param in the input data.
@@ -935,11 +936,48 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
         * @return the number of unprocessed bytes left.
         */
         int getCipherBufferLength();
+
+        /**
+         * Determines if two arrays have any overlapping memory.
+         *
+         * @param input The input array.
+         * @param inputStart The index of the input arrays start.
+         * @param inputEnd The index of the input arrays end.
+         * @param output The output array.
+         * @param outputStart The index of the output arrays start.
+         * @param outputEnd The index of the output arrays end.
+         * @return true if any memory overlaps, false otherwise.
+         */
+        static boolean arraysOverlap(byte[] input, int inputStart, int inputEnd, byte[] output, int outputStart, int outputEnd) {
+
+            // Check if there is potential overlap if input and output buffers have same address.
+            if (input != output) {
+                return false;
+            }
+
+            // If the start of input is anywhere in the range of the output array, there is an overlap.
+            if ((outputStart <= inputStart) && (inputStart < outputEnd)) {
+                return true;
+            }
+
+            // If the start of output is anywhere in the range of the input array, there is an overlap.
+            if ((inputStart <= outputStart) && (outputStart < inputEnd)) {
+                return true;
+            }
+
+            return false; // Memory does not overlap, return false.
+        }
     }
 
     private final class EngineStreamOnly implements ChaChaEngine {
 
         EngineStreamOnly() { }
+
+        @Override
+        public int getOutputSize(int inLength, boolean isFinal) {
+            // The isFinal parameter is not relevant in this kind of engine.
+            return inLength;
+        }
 
         @Override
         public synchronized int doUpdate(byte[] in, int inOff, int inLen, byte[] out,
@@ -962,7 +1000,20 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
 
                 if (in != null) {
                     Objects.checkFromIndexSize(inOff, inLen, in.length);
-                    int ret = nativeCrypto.ChaCha20Update(context, in, inOff, inLen, out, outOff, /*aadArray*/ null, /*aadArray.length*/ 0);
+
+                    // Check if there is any overlap with the input and output
+                    // buffers. If so, create a temporary copy of the input such
+                    // that it does not overlap while performing an update computation
+                    // within the openssl library.
+                    int ret = -1;
+                    if (ChaChaEngine.arraysOverlap(in, inOff, inOff + inLen, out, outOff, outOff + getOutputSize(inLen, false))) {
+                        byte[] newInArray = Arrays.copyOfRange(in, inOff, inOff + inLen);
+                        ret = nativeCrypto.ChaCha20Update(context, newInArray, 0, inLen, out, outOff, /*aadArray*/ null, /*aadArray.length*/ 0);
+                        Arrays.fill(newInArray, (byte)0);
+                    } else {
+                        ret = nativeCrypto.ChaCha20Update(context, in, inOff, inLen, out, outOff, /*aadArray*/ null, /*aadArray.length*/ 0);
+                    }
+
                     if (ret == -1) {
                         throw new ProviderException("Error in Native ChaCha20Cipher");
                     }
@@ -990,6 +1041,11 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
 
         EngineAEADEnc() throws InvalidKeyException {
             counter = 1;
+        }
+
+        @Override
+        public int getOutputSize(int inLength, boolean isFinal) {
+            return (isFinal ? Math.addExact(inLength, TAG_LENGTH) : inLength);
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -935,12 +936,49 @@ int doFinal(byte[] in, int inOff, int inLen, byte[] out, int outOff)
...
+                return false;
+            }
+
+            // If the start of input is anywhere in the range of the output array, there is an overlap.
+            if (outputStart <= inputStart && inputStart <= outputEnd) {
+                return true;
+            }
+
+            // If the start of output is anywhere in the range of the input array, there is an overlap.
+            if (inputStart <= outputStart && outputStart <= inputEnd) {
```
keithc-ca, line 964: Please add parentheses and `<` should be used instead of `<=` for the end part of the conditions:
```java
            if ((outputStart <= inputStart) && (inputStart < outputEnd)) {
```

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -935,12 +936,49 @@ int doFinal(byte[] in, int inOff, int inLen, byte[] out, int outOff)
...
+        }
     }
 
     private final class EngineStreamOnly implements ChaChaEngine {
 
         EngineStreamOnly() { }
 
+        @Override
+        public int getOutputSize(int inLength, boolean isFinal) {
+            // The isFinal parameter is not relevant in this kind of engine
```
keithc-ca, line 978: Please finish each sentence with a period.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -962,7 +1000,23 @@ public synchronized int doUpdate(byte[] in, int inOff, int inLen, byte[] out,
...
-                    int ret = nativeCrypto.ChaCha20Update(context, in, inOff, inLen, out, outOff, /*aadArray*/ null, /*aadArray.length*/ 0);
+
+                    // Check if there is any overlap with the input and output
+                    // buffers. If so, create a temporary copy of the input such
+                    // that it does not overlap while performing an update computation
+                    // within the openssl library.
+                    int ret = -1;
+                    if (ChaChaEngine.arraysOverlap(in, inOff, inOff + inLen, out, outOff, outOff + getOutputSize(inLen, false))) {
+                        byte[] newInArray = new byte[inLen];
+                        System.arraycopy(in, inOff, newInArray, 0, inLen);
```
keithc-ca, line 1011: Suggest using `Arrays.copyOfRange()` (here and on line 1090):
```java
                        byte[] newInArray = Arrays.copyOfRange(in, inOff, inOff + inLen);
```

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -962,7 +1000,23 @@ public synchronized int doUpdate(byte[] in, int inOff, int inLen, byte[] out,
...
+                        ret = nativeCrypto.ChaCha20Update(context, newInArray, 0, inLen, out, outOff, /*aadArray*/ null, /*aadArray.length*/ 0);
+                        if (newInArray != null) {
```
keithc-ca, line 1013: Please remove this test which could never fail (`OutOfMemoryError` would have been thrown); also line 1093.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -992,6 +1046,11 @@ private final class EngineAEADEnc implements ChaChaEngine {
             counter = 1;
         }
 
+        @Override
+        public int getOutputSize(int inLength, boolean isFinal) {
+            return (isFinal ? Math.addExact(inLength, TAG_LENGTH) : inLength);
```
keithc-ca, line 1051: Please omit the outermost parentheses.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -1069,6 +1144,18 @@ private final class EngineAEADDec implements ChaChaEngine {
...
+        @Override
+        public int getOutputSize(int inLen, boolean isFinal) {
+            // If we are performing a decrypt-update we should always return
+            // zero length since we cannot return any data until the tag has
+            // been consumed and verified.  CipherSpi.engineGetOutputSize will
+            // always set isFinal to true to get the required output buffer
+            // size.
+            return (isFinal ?
+                    Integer.max(Math.addExact((inLen - TAG_LENGTH),
+                            getCipherBufferLength()), 0) : 0);
```
keithc-ca, line 1156: This isn't the same as the code it replaces for all values of `inLen`, and the outermost parentheses are unwanted:
```java
            int outLen = 0;
            if (isFinal) {
                outLen = Integer.max(inLen, TAG_LENGTH) - TAG_LENGTH;
                outLen = Math.addExact(outLen, getCipherBufferLength());
            }
            return outLen;
```

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -1155,7 +1242,10 @@ public synchronized int doFinal(byte[] in, int inOff, int inLen, byte[] out,
 
         @Override
         public int getCipherBufferLength() {
-            return cipherBuf.size();
+            if (cipherBuf != null) {
+                return cipherBuf.size();
+            }
+            return 0;
```
keithc-ca, line 1248: Suggest using the ternary operator here:
```java
            return (cipherBuf != null) ? cipherBuf.size() : 0;
```

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -677,7 +667,7 @@ private static byte[] getEncodedKey(Key key) throws InvalidKeyException {
      */
     @Override
     protected byte[] engineUpdate(byte[] in, int inOfs, int inLen) {
-        byte[] out = new byte[inLen];
+        byte[] out = new byte[engine.getOutputSize(inLen, false)];
         try {
             int size = engine.doUpdate(in, inOfs, inLen, out, 0);
             // Special case for EngineAEADDec, doUpdate only buffers the input
```
keithc-ca, line 673: Can you please fix this comment: I think it's one sentence, so it should end with a period and the second line should not start with a capital.

keithc-ca (changes requested):
Please also correct the commit message and the description:
* algorothms -> algorithms
* longer then necessary -> longer than necessary

jasonkatonica:
> Please also correct the commit message and the description:
> 
>     * algorothms -> algorithms
> 
>     * longer then necessary -> longer than necessary

Updated both the commit message and the github PR description as per above.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: I added parentheses in latest code.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: Added a period to this sentence.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: Updated to use the method suggested instead of `System.arraycopy`.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: Removed this test here and on line 1093.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: This code matches what is in the non native ChaCha20 class. Id prefer to keep it such that the classes can be more easily diffed but made this change anyway. If its ok id like to change it back to match the non native class for maintenance reasons.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: This code matches what is in the non native ChaCha20 class. Id prefer to keep it such that the classes can be more easily diffed but made this change anyway. If its ok id like to change it back to match the non native class for maintenance reasons.

Yes this is part of what i believe is a bug in our Native Cha Cha code. The non native code determines the output buffer size differently. Matching that code allowed for the failing test to pass without any other unit or openjdk tests failing that i could find.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: This code matches what is in the non native ChaCha20 class. Id prefer to keep it such that the classes can be more easily diffed but made this change anyway. If its ok id like to change it back to match the non native class for maintenance reasons.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: Fixed this sentence as suggested. I lower cased the comment on the second line and added a period.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: I think this method would not work as intended if we changed the comparison to `<` instead of `<=`. The case I have in mind is the following:
- Same buffer being used
- inputStart = 1
- inputEnd = 4
- outputStart = 4
- outputEnd = 8
We want this to return true to indicate an overlay with this set of parameters to this method.

In this case the inputEnd and outputStart overlap by just a single byte, I think we would need `<=` to catch this condition at the end of line 964.

keithc-ca: No, based on how this function is used, the end offsets are exclusive. So if input is `[1,4)` its length is 3 and only bytes at indices 1, 2 and 3 are accessed which doesn't overlap with an output beginning at offset 4 or larger.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
keithc-ca: That's a reasonable plan, but I see many other unnecessary differences between `NativeChaCha20Cipher` and `ChaCha20Cipher`. Please create an issue to eliminate as many of those differences as possible.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
keithc-ca: It may be equivalent for real-world values of `inLen` (i.e. `inLen >= Integer.MIN_VALUE + TAG_LENGTH`), but I see no need to settle for an expression that doesn't work for all values.

keithc-ca: Ok, I think I've convinced myself that part of the problem was that the old code was incorrect; see the [upstream change](https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/cce883c22e8e15ecc1ba7a395f8ca698c781c3f3). We want something equivalent to what you had when you created this pull request.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: OK yes does seem like this is the case in how this method is used. I updated the code using commit [f7f939c](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/744/commits/f7f939c52246361b747e1bdae5f32e515e753c66) to reflect this.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: Added `has` to this sentence using [04f96c5](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/744/commits/04f96c5d650a6ff8fdf18ea9d4a1ce2ce4095843)

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: I opened issue https://github.ibm.com/runtimes/jit-crypto/issues/362 to track these future efforts to sync up this code further.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
jasonkatonica: Given we have decided to not make any changes to these lines I believe the code as it stands now has the [original logic present in the ChaChaCipher.java file ](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/fc9b90c93d840c89b162fbf70f508e15616c6fa9/src/java.base/share/classes/com/sun/crypto/provider/ChaCha20Cipher.java#L1541) besides the change in parentheses that we made.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
keithc-ca: Ideally that issue would exist in a public repository, either this one or in eclipse-openj9/openj9.

keithc-ca:
If you like, those parentheses I asked you to remove could be restored (in line with the code in `ChaCha20Cipher.java`), then this should be squashed.

jasonkatonica:
Reverted just a few portions of this code associated with the parentheses and use of a ternary operator to have it match `ChaCha20Cipher`. If this looks OK i will proceed with a rebase into a single commit.

jasonkatonica:
> I'm happy with the result; please squash.

Rebase/squash complete.

keithc-ca:
Jenkins copyright check

keithc-ca:
Jenkins line endings check

keithc-ca (approved):

keithc-ca:
Tested locally.

PR #746, opened 2024-02-20

JasonFengJ9:

Conflicts:
```
```


src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1723,20 +1723,25 @@ public void interrupt() {
...
+            Interruptible blocker;
+            synchronized (interruptLock) {
+                blocker = nioBlocker;
+                if (blocker != null) {
+                    blocker.interrupt(this);
+                }
+            }
             if (blocker != null) {
-                blocker.interrupt(this);
+                blocker.postInterrupt();
```
keithc-ca, line 1742: I think this method (in its entirety) should be identical to upstream.

keithc-ca (changes requested):

src/java.base/share/classes/java/lang/Thread.java
JasonFengJ9: Matched `Thread.interrupt()` to the upstream, also `Thread.getAndClearInterrupt()`.

keithc-ca (approved):

PR #747, opened 2024-02-26
Initialize VarHandle.methodHandleTable and its entries with atomic CAS

jdmpapin:
This prevents double initialization when multiple threads are racing through `getMethodHandle()`. Once a non-null reference has been written to one of these locations, any subsequent write to the same location is a problem for the OpenJ9 JIT given its current handling of known objects.

Issue: eclipse-openj9/openj9#18882

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/VarHandle.java
+++ b/src/java.base/share/classes/java/lang/invoke/VarHandle.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ */
+
 package java.lang.invoke;
 
 import java.lang.constant.ClassDesc;
@@ -35,6 +41,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
+import jdk.internal.misc.Unsafe;
 import jdk.internal.vm.annotation.DontInline;
 import jdk.internal.vm.annotation.ForceInline;
 import jdk.internal.vm.annotation.IntrinsicCandidate;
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/lang/invoke/VarHandle.java
```diff
@@ -2198,15 +2206,32 @@ public Optional<VarHandleDesc> describeConstable() {
...
     final MethodHandle getMethodHandle(int mode) {
         MethodHandle[] mhTable = methodHandleTable;
         if (mhTable == null) {
-            mhTable = methodHandleTable = new MethodHandle[AccessMode.COUNT];
+            mhTable = new MethodHandle[AccessMode.COUNT];
+            if (!UNSAFE.compareAndSetReference(this, METHOD_HANDLE_TABLE_OFFSET, null, mhTable)) {
+                // We lost the race. Use the winning table instead.
+                mhTable = methodHandleTable;
+                assert mhTable != null;
+            }
```
keithc-ca, line 2224: Suggest this should use `compareAndExchangeReference()` which avoids an unnecessary access and makes it clear the result is non-null:
```java
            Object other = UNSAFE.compareAndExchangeReference(this, METHOD_HANDLE_TABLE_OFFSET, null, mhTable);
            if (other != null) {
                // We lost the race. Use the winning table instead.
                mhTable = (MethodHandle[])other;
            }
```
Likewise for element access below:
```java
            long offset = Unsafe.ARRAY_OBJECT_BASE_OFFSET + (Unsafe.ARRAY_OBJECT_INDEX_SCALE * mode);
            Object other = UNSAFE.compareAndExchangeReference(mhTable, offset, null, mh);
            if (other != null) {
                // We lost the race. Use the winning table instead.
                mh = (MethodHandle)other;
            }
```

src/java.base/share/classes/java/lang/invoke/VarHandle.java
```diff
@@ -23,8 +23,16 @@
...
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ */
+
 package java.lang.invoke;
 
+import jdk.internal.misc.Unsafe;
```
keithc-ca, line 34: Please put this import with others for jdk.internal.

keithc-ca (changes requested):

keithc-ca:
> a problem for the OpenJ9 JIT given its current handling of known objects

I'm not sure the problem is confined to OpenJ9: the assertion is generated by [InvokerBytecodeGenerator](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/src/java.base/share/classes/java/lang/invoke/InvokerBytecodeGenerator.java#L720).

jdmpapin:
Yes, but a subsequent write only causes the assertion to fail because of the way we deal with known objects. Otherwise, we'd have gotten the LF from the MH at runtime and it would match

keithc-ca:
> Yes, but a subsequent write only causes the assertion to fail because of the way we deal with known objects. Otherwise, we'd have gotten the LF from the MH at runtime and it would match

I'm not sure whether to infer from that response that future changes to OpenJ9 are anticipated which will eliminate the need for this patch.

jdmpapin:
Yes, we anticipate future OpenJ9 changes will allow this to be reverted, in particular the implementation of https://github.com/eclipse-openj9/openj9/issues/16616. See https://github.com/eclipse-openj9/openj9/issues/18882#issuecomment-1958173709 and https://github.com/eclipse-openj9/openj9/issues/18882#issuecomment-1960277863. It will be some time though. I have some not-yet-contributed partial progress toward it, but none of that is on the code generation / GC interactions, which I expect to be the bulk of the work

jdmpapin:
Updated to address review comments

keithc-ca:
Jenkins test sanity.openjdk zlinux jdknext

keithc-ca:
Jenkins copyright check

keithc-ca:
The failures seem to only be with respect to `java.lang.Class.toGenericString()` because `sealed` is missing; see https://github.com/eclipse-openj9/openj9/issues/19033.

keithc-ca (approved):

PR #754, opened 2024-03-07
[FFI/JTreg] Remove the pragma lines for AIX in test suites

ChengJin01:
The intention of the changes is to avoid conflicts
with our own code with 4-byte alignment for double on
AIX which already supports both the default setting
by the compiler and the natural alignment for double.

### DIFF
--- a/test/jdk/java/foreign/nested/libNested.c
+++ b/test/jdk/java/foreign/nested/libNested.c
@@ -21,13 +21,16 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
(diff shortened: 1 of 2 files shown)


ChengJin01:
Reviewer: @keithc-ca
FYI: @tajila, @pshipton

keithc-ca (approved):

keithc-ca:
I don't think testing this is worthwhile, it essentially reverts
* [8318175: AIX PPC64: Handle alignment of double in structs](https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/a05f0a3f0b68bb7e1e3d00d3933f4e896b2db6f2)

and tests were passing before that change.

PR #756, opened 2024-03-14
Set property com.ibm.fips.mode based upon active profile

jasonkatonica:
When loading a restricted security mode profile we need to set the property value `com.ibm.fips.mode` to the value contained within the active profile.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -472,6 +472,12 @@ public final class RestrictedSecurity {
         propsMapping.put("jdk.tls.legacyAlgorithms", restricts.jdkTlsLegacyAlgorithms);
         propsMapping.put("jdk.certpath.disabledAlgorithms", restricts.jdkCertpathDisabledAlgorithms);
         propsMapping.put("jdk.security.legacyAlgorithm", restricts.jdkSecurityLegacyAlgorithm);
+        String fipsMode = System.getProperty("com.ibm.fips.mode");
+        if (fipsMode == null) {
+            System.setProperty("com.ibm.fips.mode", restricts.jdkFipsMode);
+        } else if (!fipsMode.equals(restricts.jdkFipsMode)) {
+            printStackTraceAndExit("Property com.ibm.fips.mode is incompatible with semeru.customprofile and semeru.fips properties");
+        }
 
         for (Map.Entry<String, String> entry : propsMapping.entrySet()) {
             String jdkPropsName = entry.getKey();
@@ -592,6 +598,8 @@ public final class RestrictedSecurity {
         // For SecureRandom.
         String jdkSecureRandomProvider;
         String jdkSecureRandomAlgorithm;
+
+        String jdkFipsMode;
 
         // Provider with argument (provider name + optional argument).
         private final List<String> providers;
@@ -749,6 +757,8 @@ public final class RestrictedSecurity {
                     securityProps.getProperty(profileID + ".securerandom.provider"));
             jdkSecureRandomAlgorithm = parseProperty(
                     securityProps.getProperty(profileID + ".securerandom.algorithm"));
+            jdkFipsMode = parseProperty(
+                    securityProps.getProperty(profileID + ".fips.mode"));
 
             if (debug != null) {
                 debug.println("\tProperties of restricted security profile successfully loaded.");
@@ -1068,6 +1078,8 @@ public final class RestrictedSecurity {
                     securityProps.getProperty(profileToPrint + ".desc.default"));
             printProperty(profileToPrint + ".desc.fips: ",
                     securityProps.getProperty(profileToPrint + ".desc.fips"));
+            printProperty(profileToPrint + ".fips.mode: ",
+                    securityProps.getProperty(profileToPrint + ".fips.mode"));
             printProperty(profileToPrint + ".desc.number: ",
                     parseProperty(securityProps.getProperty(profileToPrint + ".desc.number")));
             printProperty(profileToPrint + ".desc.policy: ",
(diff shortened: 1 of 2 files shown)


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -472,6 +472,7 @@ private static void setProperties(Properties props) {
         propsMapping.put("jdk.tls.legacyAlgorithms", restricts.jdkTlsLegacyAlgorithms);
         propsMapping.put("jdk.certpath.disabledAlgorithms", restricts.jdkCertpathDisabledAlgorithms);
         propsMapping.put("jdk.security.legacyAlgorithm", restricts.jdkSecurityLegacyAlgorithm);
+        System.setProperty("com.ibm.fips.mode", restricts.jdkFipsMode);
```
keithc-ca, line 475: What if a user also says `-Dcom.ibm.fips.mode=140-n`? It seems wrong to ignore it if it's incompatible with the value specified in `semeru.customprofile`.

src/java.base/share/conf/security/java.security
```diff
@@ -94,6 +94,7 @@ security.provider.tbd=SunPKCS11
 RestrictedSecurity.NSS.140-2.desc.name = Red Hat Enterprise Linux 8 NSS Cryptographic Module FIPS 140-2
 RestrictedSecurity.NSS.140-2.desc.default = true
 RestrictedSecurity.NSS.140-2.desc.fips = true
+RestrictedSecurity.NSS.140-2.fips.mode = 140-2
```
keithc-ca, line 97: I think this should be added after the set of descriptive properties; likewise for FIPS140-3.

jasonkatonica: Makes sense to me too. Moved this property for both the NSS and OpenJCEPlus profiles to the end of the description section. I also added a new line right after the description section in the OpenJCEPlus profile to segment that profile similar to NSS.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
jasonkatonica: Yes it is true that this value will override any existing values for `com.ibm.fips.mode` when a restricted security mode profile is loaded. Given that this property will be based entirely upon the profile being set I am unsure what other behavior to recommend here. Perhaps we should add a trace to this section to indicate that a user defined value was overwritten?

keithc-ca: I suggest use of `printStackTraceAndExit()` is appropriate. In my view, this is like asking to enable FIPS on a platform where it's not supported. Other code may have already inspected `com.ibm.fips.mode` and made decisions based on the value seen then; I don't think it's reasonable to just overwrite the property. Perhaps we want something like this:
```java
        String fipsMode = System.getProperty("com.ibm.fips.mode");
        if (fipsMode == null) {
            System.setProperty("com.ibm.fips.mode", restricts.jdkFipsMode);
        } else if (!fipsMode.equals(restricts.jdkFipsMode)) {
            printStackTraceAndExit("com.ibm.fips.mode incompatible with semeru.customprofile");
        }
```

jasonkatonica: Yes agreed that we can make this fatal. 

I updated the code as suggested, only I changed the error message to `Property com.ibm.fips.mode is incompatible with semeru.customprofile and semeru.fips properties.` This property should not be set in the case where a user is using either default ( just `-Dsemeru.fips=true` ) or non default ( `-Dsemeru.fips=true` and `-Dsemeru.customprofile` ) properties and the restricted security mode.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -472,6 +472,12 @@ private static void setProperties(Properties props) {
         propsMapping.put("jdk.tls.legacyAlgorithms", restricts.jdkTlsLegacyAlgorithms);
         propsMapping.put("jdk.certpath.disabledAlgorithms", restricts.jdkCertpathDisabledAlgorithms);
         propsMapping.put("jdk.security.legacyAlgorithm", restricts.jdkSecurityLegacyAlgorithm);
+        String fipsMode = System.getProperty("com.ibm.fips.mode");
+        if (fipsMode == null) {
+            System.setProperty("com.ibm.fips.mode", restricts.jdkFipsMode);
+        } else if (!fipsMode.equals(restricts.jdkFipsMode)) {
+            printStackTraceAndExit("Property com.ibm.fips.mode is incompatible with semeru.customprofile and semeru.fips properties.");
```
keithc-ca, line 479: Other uses of `printStackTraceAndExit()` don't include the trailing period; suggest this shouldn't either.

keithc-ca (changes requested):
Please also squash and rebase on the head of the openj9 branch.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
jasonkatonica: I did search the code base for this and about 1/3 used periods and the rest did not. I removed the period from the message just now in this update.

keithc-ca: Consistency would have been nice, but it's probably not critical (nor worth dealing with in this pull request).

keithc-ca:
Jenkins test sanity zlinux  jdknext

keithc-ca (approved):
The test failure is unrelated to this change.

pshipton:
@keithc-ca the backports for jdk11+ are created. 

@jasonkatonica should this change go into the April release?

PR #758, opened 2024-03-18
Support provider fully-qualified class name in Restricted Security mode

taoliult:
This PR is for updating the codes to only support the fully-qualified class name when setting the providers list in restricted security profiles. For example, the NSS FIPS140-2 provider list will be:

```
RestrictedSecurity.NSS.140-2.jce.provider.1 = sun.security.pkcs11.SunPKCS11 ${java.home}/conf/security/nss.fips.cfg
RestrictedSecurity.NSS.140-2.jce.provider.2 = sun.security.provider.Sun [ \
    {CertificateFactory, X.509, ImplementedIn=Software}, \
    {CertStore, Collection, ImplementedIn=Software}, \
    {CertStore, com.sun.security.IndexedCollection, ImplementedIn=Software}, \
    {Policy, JavaPolicy, *}, {Configuration, JavaLoginConfig, *}, \
    {CertPathBuilder, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}, \
    {CertPathValidator, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}, \
    {KeyStore, PKCS12, *}]
RestrictedSecurity.NSS.140-2.jce.provider.3 = sun.security.ec.SunEC [{KeyFactory, EC, ImplementedIn=Software: \
RestrictedSecurity.NSS.140-2.jce.provider.4 = sun.security.ssl.SunJSSE
```

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -28,6 +28,7 @@ import java.security.AccessController;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.security.PrivilegedAction;
+import java.security.Provider;
 import java.security.Provider.Service;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
@@ -251,6 +252,12 @@ public final class RestrictedSecurity {
      */
     public static boolean isProviderAllowed(String providerName) {
         if (securityEnabled) {
+            // Remove argument, e.g. -NSS-FIPS, if present.
+            int pos = providerName.indexOf('-');
+            if (pos >= 0) {
+                providerName = providerName.substring(0, pos);
+            }
+
             return restricts.isRestrictedProviderAllowed(providerName);
         }
         return true;
@@ -264,17 +271,17 @@ public final class RestrictedSecurity {
      */
     public static boolean isProviderAllowed(Class<?> providerClazz) {
         if (securityEnabled) {
-            String providerName = providerClazz.getName();
+            String providerClassName = providerClazz.getName();
 
             // Check if the specified class extends java.security.Provider.
             if (java.security.Provider.class.isAssignableFrom(providerClazz)) {
-                return restricts.isRestrictedProviderAllowed(providerName);
+                return restricts.isRestrictedProviderAllowed(providerClassName);
             }
 
             // For a class that doesn't extend java.security.Provider, no need to
             // check allowed or not allowed, always return true to load it.
             if (debug != null) {
-                debug.println("The provider class " + providerName + " does not extend java.security.Provider.");
+                debug.println("The provider class " + providerClassName + " does not extend java.security.Provider.");
             }
         }
         return true;
@@ -647,27 +654,6 @@ public final class RestrictedSecurity {
     }
 
     /**
-     * Get the provider name defined in provider construction method.
-     *
-     * @param providerName provider name or provider with packages
-     * @return provider name defined in provider construction method
-     */
-    private static String getProvidersSimpleName(String providerName) {
-        if (providerName.equals("com.sun.security.sasl.Provider")) {
-            // The main class for the SunSASL provider is com.sun.security.sasl.Provider.
-            return "SunSASL";
-        } else {
-            // Remove the provider's class package names if present.
-            int pos = providerName.lastIndexOf('.');
-            if (pos >= 0) {
-                providerName = providerName.substring(pos + 1);
-            }
-            // Provider without package names.
-            return providerName;
-        }
-    }
-
-    /**
      * This class is used to save and operate on restricted security
      * properties which are loaded from the java.security file.
      */
@@ -700,7 +686,7 @@ public final class RestrictedSecurity {
         // Provider with argument (provider name + optional argument).
         private final List<String> providers;
         // Provider without argument.
-        private final List<String> providersSimpleName;
+        private final List<String> providersFullyQualifiedClassName;
         // The map is keyed by provider name.
         private final Map<String, Constraint[]> providerConstraints;
 
@@ -732,7 +718,7 @@ public final class RestrictedSecurity {
             this.jdkFipsMode = parser.getProperty("jdkFipsMode");
 
             this.providers = new ArrayList<>(parser.providers);
-            this.providersSimpleName = new ArrayList<>(parser.providersSimpleName);
+            this.providersFullyQualifiedClassName = new ArrayList<>(parser.providersFullyQualifiedClassName);
             this.providerConstraints = parser.providerConstraints
                                              .entrySet()
                                              .stream()
@@ -754,30 +740,26 @@ public final class RestrictedSecurity {
          * @return true if the Service is allowed
          */
         boolean isRestrictedServiceAllowed(Service service) {
-            String providerName = service.getProvider().getName();
+            Provider provider = service.getProvider();
+            String providerClassName = provider.getClass().getName();
 
             if (debug != null) {
-                debug.println("Checking service " + service.toString() + " offered by provider " + providerName + ".");
+                debug.println("Checking service " + service.toString() + " offered by provider " + providerClassName + ".");
             }
 
-            // Provider with argument, remove argument.
-            // e.g. SunPKCS11-NSS-FIPS, remove argument -NSS-FIPS.
-            int pos = providerName.indexOf('-');
-            providerName = (pos < 0) ? providerName : providerName.substring(0, pos);
-
-            Constraint[] constraints = providerConstraints.get(providerName);
+            Constraint[] constraints = providerConstraints.get(providerClassName);
 
             if (constraints == null) {
                 // Disallow unknown providers.
                 if (debug != null) {
                     debug.println("Security constraints check."
-                            + " Disallow unknown provider: " + providerName);
+                            + " Disallow unknown provider: " + providerClassName);
                 }
                 return false;
             } else if (constraints.length == 0) {
                 // Allow this provider with no constraints.
                 if (debug != null) {
-                    debug.println("No constraints for provider " + providerName + ".");
+                    debug.println("No constraints for provider " + providerClassName + ".");
                 }
                 return true;
             }
@@ -821,7 +803,7 @@ public final class RestrictedSecurity {
                         debug.println("The following service:"
                                 + "\n\tService type: " + type
                                 + "\n\tAlgorithm: " + algorithm
-                                + "\nis allowed in provider: " + providerName);
+                                + "\nis allowed in provider: " + providerClassName);
                     }
                     return true;
                 }
@@ -851,7 +833,7 @@ public final class RestrictedSecurity {
                                         + "\n\tService type: " + type
                                         + "\n\tAlgorithm: " + algorithm
                                         + "\n\tAttribute: " + cAttribute
-                                        + "\nis NOT allowed in provider: " + providerName);
+                                        + "\nis NOT allowed in provider: " + providerClassName);
                         }
                         return false;
                     }
@@ -865,7 +847,7 @@ public final class RestrictedSecurity {
                                 + "\n\tService type: " + type
                                 + "\n\tAlgorithm: " + algorithm
                                 + "\n\tAttribute: " + cAttribute
-                                + "\nis allowed in provider: " + providerName);
+                                + "\nis allowed in provider: " + providerClassName);
                 }
                 return true;
             }
@@ -876,7 +858,7 @@ public final class RestrictedSecurity {
                 debug.println("The following service:"
                             + "\n\tService type: " + type
                             + "\n\tAlgorithm: " + algorithm
-                            + "\nis NOT allowed in provider: " + providerName);
+                            + "\nis NOT allowed in provider: " + providerClassName);
             }
             return false;
         }
@@ -884,34 +866,25 @@ public final class RestrictedSecurity {
         /**
          * Check if the provider is allowed in restricted security mode.
          *
-         * @param providerName the provider to check
+         * @param providerClassName the provider to check
          * @return true if the provider is allowed
          */
-        boolean isRestrictedProviderAllowed(String providerName) {
+        boolean isRestrictedProviderAllowed(String providerClassName) {
             if (debug != null) {
-                debug.println("Checking the provider " + providerName + " in restricted security mode.");
+                debug.println("Checking the provider " + providerClassName + " in restricted security mode.");
             }
 
-            // Remove argument, e.g. -NSS-FIPS, if present.
-            int pos = providerName.indexOf('-');
-            if (pos >= 0) {
-                providerName = providerName.substring(0, pos);
-            }
-
-            // Provider name defined in provider construction method.
-            providerName = getProvidersSimpleName(providerName);
-
-            // Check if the provider is in restricted security provider list.
-            // If not, the provider won't be registered.
-            if (providersSimpleName.contains(providerName)) {
+            // Check if the provider fully-qualified cLass name is in restricted
+            // security provider list. If not, the provider won't be registered.
+            if (providersFullyQualifiedClassName.contains(providerClassName)) {
                 if (debug != null) {
-                    debug.println("The provider " + providerName + " is allowed in restricted security mode.");
+                    debug.println("The provider " + providerClassName + " is allowed in restricted security mode.");
                 }
                 return true;
             }
 
             if (debug != null) {
-                debug.println("The provider " + providerName + " is not allowed in restricted security mode.");
+                debug.println("The provider " + providerClassName + " is not allowed in restricted security mode.");
 
                 debug.println("Stack trace:");
                 StackTraceElement[] elements = Thread.currentThread().getStackTrace();
@@ -950,8 +923,8 @@ public final class RestrictedSecurity {
             for (int providerPosition = 0; providerPosition < providers.size(); providerPosition++) {
                 printProperty(profileID + ".jce.provider." + (providerPosition + 1) + ": ",
                         providers.get(providerPosition));
-                String providerSimpleName = providersSimpleName.get(providerPosition);
-                for (Constraint providerConstraint : providerConstraints.get(providerSimpleName)) {
+                String providerFullyQualifiedClassName = providersFullyQualifiedClassName.get(providerPosition);
+                for (Constraint providerConstraint : providerConstraints.get(providerFullyQualifiedClassName)) {
                     System.out.println("\t" + providerConstraint.toString());
                 }
             }
@@ -994,7 +967,7 @@ public final class RestrictedSecurity {
         // Provider with argument (provider name + optional argument).
         private final List<String> providers;
         // Provider without argument.
-        private final List<String> providersSimpleName;
+        private final List<String> providersFullyQualifiedClassName;
         // The map is keyed by provider name.
         private final Map<String, List<Constraint>> providerConstraints;
 
@@ -1022,7 +995,7 @@ public final class RestrictedSecurity {
             profileProperties = new HashMap<>();
 
             providers = new ArrayList<>();
-            providersSimpleName = new ArrayList<>();
+            providersFullyQualifiedClassName = new ArrayList<>();
             providerConstraints = new HashMap<>();
 
             profilesHashes = new HashMap<>();
@@ -1180,21 +1153,13 @@ public final class RestrictedSecurity {
             }
             providerName = providerName.trim();
 
-            // Remove argument, e.g. -NSS-FIPS, if present.
-            pos = providerName.indexOf('-');
-            if (pos >= 0) {
-                providerName = providerName.substring(0, pos);
-            }
-
-            // Provider name defined in provider construction method.
-            providerName = getProvidersSimpleName(providerName);
             boolean providerChanged = false;
             if (update) {
-                String previousProviderName = providersSimpleName.get(providerPos - 1);
+                String previousProviderName = providersFullyQualifiedClassName.get(providerPos - 1);
                 providerChanged = !previousProviderName.equals(providerName);
-                providersSimpleName.set(providerPos - 1, providerName);
+                providersFullyQualifiedClassName.set(providerPos - 1, providerName);
             } else {
-                providersSimpleName.add(providerPos - 1, providerName);
+                providersFullyQualifiedClassName.add(providerPos - 1, providerName);
             }
 
             if (debug != null) {
@@ -1210,14 +1175,14 @@ public final class RestrictedSecurity {
                 debug.println("\t\tRemoving provider in position " + providerPos);
             }
 
-            int numOfExistingProviders = providersSimpleName.size();
+            int numOfExistingProviders = providersFullyQualifiedClassName.size();
 
             // If this is the last provider, remove from all lists.
             if (providerPos == numOfExistingProviders) {
                 if (debug != null) {
                     debug.println("\t\t\tLast provider. Only one to be removed.");
                 }
-                String providerRemoved = providersSimpleName.remove(providerPos - 1);
+                String providerRemoved = providersFullyQualifiedClassName.remove(providerPos - 1);
                 providers.remove(providerPos - 1);
                 providerConstraints.remove(providerRemoved);
 
@@ -1241,7 +1206,7 @@ public final class RestrictedSecurity {
                 }
 
                 // Remove all of the providers that are set to empty.
-                String providerRemoved = providersSimpleName.remove(i - 1);
+                String providerRemoved = providersFullyQualifiedClassName.remove(i - 1);
                 providers.remove(i - 1);
                 providerConstraints.remove(providerRemoved);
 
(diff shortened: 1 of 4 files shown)


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -857,6 +863,63 @@ private void initConstraints() {
...
+                    debug.println("\tNo profile attributes.");
+                }
+                return;
+            }
+
+            try {
+                // Locate and parse providerattributes section.
+                pAttributes = pAttributes.replaceAll("\\s", "");
+                int startIndex = pAttributes.indexOf("<providerattributes>");
+                int endIndex = pAttributes.indexOf("</providerattributes>");
```
keithc-ca, line 885: I suggest that xml is not needed, nor is it very user-friendly in a properties file (i.e. `java.security`). The example in the description could be expressed more simply and clearly as:
```
RestrictedSecurity.NSS.140-2.attributes = \
    SunJGSS=sun.security.jgss.SunProvider, \
    SunSASL=com.sun.security.sasl.Provider
```
(possibly using a name other than `attributes`)

There are several references in comments to `providerattributes`; did you mean "provider attributes"?

keithc-ca (changes requested):
Please correct the typo in the commit message: "a attribute" -> "an attribute".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -593,9 +593,11 @@ private static final class RestrictedSecurityProperties {
         String jdkSecureRandomProvider;
         String jdkSecureRandomAlgorithm;
 
+        private final Map<String, String> providerClassNameMap;
```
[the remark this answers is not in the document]
taoliult: This Map. the provider class name is the key, and the provider name defined in its constructor is the value. If we call it "providerSimpleNamesMap", is that more appropriate?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: I considered this simply format(you suggested) before, but after discussed with the team, we think of a more general format. Right now, this XML tag “<providerattributes>” is used for the mapping between provider name and its class name. But, in the future, we can extend this "<restrictedsecurityattributes>" tag, to add other XML tags inside this “<restrictedsecurityattributes>” tag.

keithc-ca: Ok, but if you want to use xml, you should be using existing xml parsing classes.

keithc-ca: I take that back: java.base can't assume, for example, that the java.xml module is present and so can't use that code to parse xml.

Perhaps there's a better solution that doesn't require dealing with the "simple" name of a provider. Most of the uses of `isProviderAllowed(String)` could be switched to use `isProviderAllowed(Class)`. Could the remaining cases use the latter, or at least the (fully-qualfied) class name? If so, we don't need to discuss the format of the `attributes` property value.

taoliult: @keithc-ca 

If we only use provider class name to check allow or not-allow. That means, in java.security file, we have to use provider class name to configure the FIPS provider list. For example, when we want to add the SunSASL in the list, it will be like below:

`RestrictedSecurity.NSS.140-2.jce.provider.5=com.sun.security.sasl.Provider`

And then, in the `ProviderConfig` class, when loading this non-java.base provider, it will use `ServiceLoader` to load it, and due to “com.sun.security.sasl.Provider” is allowed (it can pass the `isProviderAllowed(Class)` check), so this provider class can be loaded. 

But after that, in `ProviderLoader.load()` method, it will use the provider name “SunSASL” to find the provider class object and return it [1]. But due to we did not use provider name in the java.security FIPS provider list, so the provider “SunSASL” can not be found and return (even this provider class is already loaded).

[1] - https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/src/java.base/share/classes/sun/security/jca/ProviderConfig.java#L375

And also, in this `ProviderLoader` class, it do had another method called `legacyLoad()`. It is designed to load the provider by using the provider class name. But when I tried this method for "SunSASL", I get the below error:

> ProviderList: provider configuration: [com.sun.security.sasl.Provider]
> ProviderConfig: Loading legacy provider: com.sun.security.sasl.Provider
> ProviderConfig: Error loading legacy provider com.sun.security.sasl.Provider
> java.security.PrivilegedActionException: java.lang.IllegalAccessException: Class sun/security/jca/ProviderConfig$ProviderLoader$1(java.base) can not access class com/sun/security/sasl/Provider(java.security.sasl) because module java.security.sasl does not export package com/sun/security/sasl to module java.base

So, if the provider module did not export package to java.base module, then this `legacyLoad()` won't work for that provider class.

taoliult: @keithc-ca Any update or advice on this PR?

taoliult:
@keithc-ca I replied the above questions, and any update or advice on this PR?

taoliult:
@keithc-ca 

I updated the codes, to avoid the legacyLoad() method from the ProviderConfig class. For now, for the providers which provider name and its class name are same, we still use the provider name as the provider list setting. For the SunSASL, since we already release this to the client by using the provider name for the setting, so I did not remove this provider mapping. For the other providers, which the provider name and its class name are different, we can only use the provider class as the provider list setting. For example, the SunJGSS and XMLDSig, we can set them as below:

```
RestrictedSecurity.NSS.140-2.jce.provider.1 = sun.security.jgss.SunProvider
RestrictedSecurity.NSS.140-2.jce.provider.2 = org.jcp.xml.dsig.internal.dom.XMLDSigRI
```

Please help to review and advise.

@jasonkatonica FYI.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -941,6 +940,11 @@ boolean isRestrictedServiceAllowed(Service service) {
 
             Constraint[] constraints = providerConstraints.get(providerName);
 
+            if (constraints == null) {
+                // Continue checking if the class name is the key.
+                constraints = providerConstraints.get(providerClass);
+            }
```
keithc-ca, line 946: Suggest this should check `providerClassName` first.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,12 +1084,12 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+         * Get the provider name without package name.
          *
-         * @param providerName provider name or provider with packages
-         * @return provider name defined in provider construction method
+         * @param providerName provider name with package name
+         * @return provider name without package name
          */
-        private static String getProvidersSimpleName(String providerName) {
+        private static String getProvidersNameWithoutPackage(String providerName) {
             if (providerName.equals("com.sun.security.sasl.Provider")) {
```
keithc-ca, line 1093: I don't think we need or want to make SunSASL a special case any more.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca check providerClassName first, you mean like below?

```
            Constraint[] constraints = providerConstraints.get(providerClassName);

            if (constraints == null) {
                // Continue checking if the provider name is the key.
                constraints = providerConstraints.get(providerName);
            }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca We already released a version which suggested the client to set using SunSASL in their java.security provider list. So, that is why I did not remove this SunSASL provider mapping.

taoliult:
@keithc-ca Codes updated according to code review, others replied the questions. Please help to review and advise.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: With the change to `ProviderConfig.load()`, it should it not recognize both `SunSASL` and `com.sun.security.sasl.Provider`? I think so, which eliminates the need for a special case here.

src/java.base/share/classes/sun/security/jca/ProviderConfig.java
```diff
@@ -190,6 +185,14 @@ Provider getProvider() {
             if (!shouldLoad()) {
                 return null;
             }
+            // There is RestrictedSecurity check in the ServiceLoader before the
+            // provider is initialized. So the check has already occurred for the
+            // provider where provider != null.
```
keithc-ca, line 190: I think a better place for this comment is before line 176.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca 

If we set `RestrictedSecurity.NSS.140-2.jce.provider.N=com.sun.security.sasl.Provider,` this works. Because in the ServiceLoader, before the provider initialized, it will use the package + class name to check, and due to we use the package + class name in the java.security provider list, the check will pass.

But if we set `RestrictedSecurity.NSS.140-2.jce.provider.N=SunSASL`, this won't work. Because in the ServiceLoader, before the provider initialized, it will use the class name to check, and due to we use the provider name in the java.security provider list, and for SunSASL, its provider name is different with its class name(after removing the package), so it won't pass the check in ServiceLoader. 

For the non-java.base providers, the check happens in the ServiceLoader using its package + class name. So, if the provider name is different with its class name(after removing the package). We have to set its package + class name in the java.security RestrictedSecurity provider list.

taoliult:
@keithc-ca Codes updated according to code review, others replied the questions. Please help to review and advise.

taoliult:
@keithc-ca Codes updated according to today's meeting. For NSS FIPS 140-2, both provider name and fully-qualified class name are accepted. For OpenJCEPlus FIPS 140-3, only fully-qualified class name is accepted. And also, removed the mapping for SunSASL.

And, because of the ProviderConfig updates need to call the method "RestrictedSecurity.isFIPS1402()" from RestrictedSecurity.java, so this PR depends on the PR https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/793

@jasonkatonica FYI.

src/java.base/share/classes/sun/security/jca/ProviderConfig.java
```diff
@@ -372,8 +375,14 @@ public Provider load(String pn) {
...
-                        return p;
+                    if (RestrictedSecurity.isFIPSEnabled() && !RestrictedSecurity.isFIPS1402()) {
+                        if (p.getClass().getName().equals(pn)) {
+                            return p;
+                        }
+                    } else {
+                        if (pName.equals(pn) || p.getClass().getName().equals(pn)) {
+                            return p;
+                        }
                     }
```
keithc-ca, line 386: Suggest checking the class name first:
```java
                    if (p.getClass().getName().equals(pn)) {
                        return p;
                    } else if (pName.equals(pn)) {
                        if (!RestrictedSecurity.isFIPSEnabled() || RestrictedSecurity.isFIPS1402()) {
                            return p;
                        }
                    }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -932,7 +931,9 @@ private void initConstraints() {
          * @return true if the Service is allowed
          */
         boolean isRestrictedServiceAllowed(Service service) {
-            String providerName = service.getProvider().getName();
+            Provider provider = service.getProvider();
+            String providerName = provider.getName();
+            String providerClassName = provider.getClass().getName();
```
keithc-ca, line 936: Suggest updating the local in `isProviderAllowed(Class<?>)` from `providerName` to `providerClassName` to be consistent with this method.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: Yes.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: In retrospect, it was a mistake to try to support `RestrictedSecurity.NSS.140-2.jce.provider.N=SunSASL`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1043,10 +1049,20 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+            // security provider list without removing the package name.
+            if (providersSimpleName.contains(providerName)) {
+                if (debug != null) {
+                    debug.println("The provider " + providerName + " class is allowed in restricted security mode.");
+                }
+                return true;
+            }
+
+            // Provider name without package name.
+            providerName = getProvidersNameWithoutPackage(providerName);
```
keithc-ca, line 1063: I don't think we want this any more. Only the name as specified in the properties file (less any argument) should be used. The name of the field, `providersSimpleName`, should be changed: perhaps `providers` and `providersSimpleName`
should be `providerWithArg` and `providerWithoutArg`, respectively.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca 

I remember in our last meeting, for the NSS FIPS 140-2 provider list, we still use the provider name (the name from its constructor) in the provider list. Only the OpenJCEPlus FIPS 140-3 and any new providers which need to be added into the NSS FIPS 140-2 provider list, for example the SunSASL, will use the provider fully qualified name.

```
RestrictedSecurity.NSS.140-2.jce.provider.1 = SunPKCS11 ${java.home}/conf/security/nss.fips.cfg
RestrictedSecurity.NSS.140-2.jce.provider.2 = SUN [{CertificateFactory, X.509, ImplementedIn=Software}, \
    {CertStore, Collection, ImplementedIn=Software}, \
    {CertStore, com.sun.security.IndexedCollection, ImplementedIn=Software}, \
    {Policy, JavaPolicy, *}, {Configuration, JavaLoginConfig, *}, \
    {CertPathBuilder, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}, \
    {CertPathValidator, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}, \
    {KeyStore, PKCS12, *}]
RestrictedSecurity.NSS.140-2.jce.provider.3 = SunEC [{KeyFactory, EC, ImplementedIn=Software: \
RestrictedSecurity.NSS.140-2.jce.provider.4 = SunJSSE
```

So, for example, the SunEC, when the ServiceLoader check the sun.security.ec.SunEC, we still need to get its provider name SunEC from removing the package. So, that is why I did not remove this getProvidersNameWithoutPackage() method.

@jasonkatonica FYI.

taoliult:
@keithc-ca 

Codes updated according to code review, others replied the questions. Please help to review and advise.

And I also added "isFIPS1402()" method from Allow extension of RestrictedSecurity profile PR https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/793. So, this PR won’t have the dependency on Allow extension of RestrictedSecurity profile PR.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: Please update the description of this pull request: the discussion of unqualified names and XML are no longer relevant.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: My understanding is that we agreed that only fully-qualified class names would be allowed when specifying a restricted security profile (otherwise the restrictions can be bypassed). Yes, this may mean some migration work for users, but the longer we wait to make that change, the greater number of users will be impacted.

taoliult:
@keithc-ca As we talked today, I removed the checks using the provider name, only use the fully qualified class name when checking the providers and services, for all the restricted security profiles. The changes I made as below:

1. Removed the method isFIPS1402() since we don’t need it.
2. Changed the variable name providersSimpleName to providersFullyQualifiedClassName.
3. Removed the codes about “remove argument -NSS-FIPS”, since we only use the fully qualified class name.
4. Removed the method getProvidersNameWithoutPackage() since we don’t need to remove the package.

Please help to review and advise.

@jasonkatonica FYI.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -758,15 +759,7 @@ private void initProviders() {
                 }
                 providerName = providerName.trim();
 
-                // Remove argument, e.g. -NSS-FIPS, if present.
-                pos = providerName.indexOf('-');
-                if (pos >= 0) {
-                    providerName = providerName.substring(0, pos);
-                }
```
keithc-ca, line 765: Why don't we need to keep this?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1037,18 +1026,9 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
-                providerName = providerName.substring(0, pos);
-            }
-
-            // Provider name defined in provider construction method.
-            providerName = getProvidersSimpleName(providerName);
-
-            // Check if the provider is in restricted security provider list.
-            // If not, the provider won't be registered.
-            if (providersSimpleName.contains(providerName)) {
+            // Check if the provider fully qualified cLass name is in restricted
```
keithc-ca, line 1029: Missing hyphen in "fully-qualified".

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: Revert them, done.

taoliult:
@keithc-ca Codes updated according to review, please help to review and advise.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -932,7 +931,8 @@ private void initConstraints() {
          * @return true if the Service is allowed
          */
         boolean isRestrictedServiceAllowed(Service service) {
-            String providerName = service.getProvider().getName();
+            Provider provider = service.getProvider();
+            String providerName = provider.getClass().getName();
 
             // Provider with argument, remove argument.
             // e.g. SunPKCS11-NSS-FIPS, remove argument -NSS-FIPS.
```
keithc-ca, line 938: This no longer makes sense (`providerName` is a fully-qualified class name).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1043,12 +1043,9 @@ boolean isRestrictedProviderAllowed(String providerName) {
                 providerName = providerName.substring(0, pos);
```
keithc-ca, line 1043: The parameter of this method should be renamed to `providerClassName` to more accurately reflect its meaning.

The removal of the argument does not belong here, but should instead be in `isProviderAllowed(String)` (the only place it might not already be a fully-qualified class name).

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: Yes, because we get the provider class name from the `Service`, so no need to remove argument. I removed these lines. 

And I also removed the "Remove argument" codes from the `initProviders()` method, because we only set the  fully-qualified class name in the java.security profile, so no need to remove the argument.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: Yes, and I also renamed to `providerClassName` in method `isRestrictedServiceAllowed(Service service)`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: What about providers that must specify an argument, like in
```
RestrictedSecurity.NSS.140-2.securerandom.provider = SunPKCS11-NSS-FIPS
```
The reference to `SunPKCS11` should probably use the fully-qualified name, but `-NSS-FIPS` still needs to be removed to derive the class name.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: I don't think we can set as below:

`RestrictedSecurity.NSS.140-2.securerandom.provider = SunPKCS11-NSS-FIPS`

The provider name SunPKCS11-NSS-FIPS comes from two parts, provider name defined in its constructor + name defined in its configure file, the "${java.home}/conf/security/nss.fips.cfg" file. The SunPKCS11 constructor codes below:

```
    // Used by Secmod
    SunPKCS11(Config c) {
        super("SunPKCS11-" + c.getName(), PROVIDER_VER, c.getDescription());
        this.config = c;
```

When this provider is initialized with the configure file, for example the "nss.fips.cfg" file, the name will be SunPKCS11-NSS-FIPS. Without configure file, the name will be SunPKCS11.

The argument "-NSS-FIPS" is not a part of the name in its constructor. So I don't think we can specify the argument (without using the configure file) in the java.security provider list.

keithc-ca:
> Please update the description of this pull request: the discussion of unqualified names and XML are no longer relevant.

See https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/758#discussion_r1643165104.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: My point is that the entry added to `providersFullyQualifiedClassName` is expected to be a fully-qualified class name, which would be `sun.security.pkcs11.SunPKCS11`, not `sun.security.pkcs11.SunPKCS11-NSS-FIPS`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: Description updated.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: Yes, that is correct. For the SunPKCS11 in NSS FIPS140-2 profile, it would be:

`RestrictedSecurity.NSS.140-2.jce.provider.1 = sun.security.pkcs11.SunPKCS11 ${java.home}/conf/security/nss.fips.cfg`

keithc-ca: Perhaps you misunderstand my concern. An existing properties file it might say:
```
RestrictedSecurity.NSS.140-2.jce.provider.1 = SunPKCS11-NSS-FIPS ...
```
I expected that this change will require that the class name will be fully-qualified, but the argument would still be present:
```
RestrictedSecurity.NSS.140-2.jce.provider.1 = sun.security.pkcs11.SunPKCS11-NSS-FIPS ...
```
Are you saying that must the argument be omitted?
```
RestrictedSecurity.NSS.140-2.jce.provider.1 = sun.security.pkcs11.SunPKCS11 ...
```
If so, it would seem there's no way to disallow other configurations of a given provider.

It's also not clear to me how the existing `java.properties` file could work properly for `SunPKCS11` because it does not specify the argument and so `RestrictedSecurity.isProviderAllowed("SunPKCS11-NSS-FIPS")` would return `false`.

keithc-ca (changes requested):

keithc-ca:
Please squash.

taoliult:
@keithc-ca Squashed, please help to review and advise.

keithc-ca:

keithc-ca (approved):

taoliult:
> @taoliult Please confirm that this change need not be back-ported to https://github.com/ibmruntimes/openj9-openjdk-jdk8.

@keithc-ca We need to back-ported it to Semeru OpenJDK8, working on it and will create the PR shortly.

PR #765, opened 2024-03-27
Add support for MD5 digest using native OpenSSL library

KostasTsiounis:
A new class, namely `MD5.java`, is created and the corresponding options are added in other places to allow for the use of native `OpenSSL` library to perform `MD5` digests.

### DIFF
--- a/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2024 All Rights Reserved
  * ===========================================================================
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -47,6 +47,7 @@ public class NativeCrypto {
     public static final int SHA2_256 = 2;
     public static final int SHA5_384 = 3;
     public static final int SHA5_512 = 4;
+    public static final int MD5 = 5;
 
     /* Define constants for the EC field types. */
     public static final int ECField_Fp = 0;
@@ -200,6 +201,8 @@ public class NativeCrypto {
     /* Native digest interfaces */
 
     private static final native long loadCrypto(boolean trace);
+
+    public static final native boolean isMD5Available();
 
     public final native long DigestCreateContext(long nativeBuffer,
                                                  int algoIndex);
--- a/closed/src/java.base/share/classes/sun/security/provider/NativeMD5.java
+++ b/closed/src/java.base/share/classes/sun/security/provider/NativeMD5.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2003, 2014, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ */
+
+package sun.security.provider;
+
+import jdk.crypto.jniprovider.NativeCrypto;
+
+public final class NativeMD5 extends NativeDigest {
+
+    public NativeMD5() {
+        super("MD5", 16, NativeCrypto.MD5);
+    }
+}
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2024 All Rights Reserved
  * ===========================================================================
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -200,6 +200,7 @@ static OSSL_CRYPTO_THREADID_set_callback_t* OSSL_CRYPTO_THREADID_set_callback = 
 static OSSL_CRYPTO_set_locking_callback_t* OSSL_CRYPTO_set_locking_callback = NULL;
 
 /* Define pointers for OpenSSL functions to handle Message Digest algorithms. */
+OSSL_sha_t* OSSL_md5;
 OSSL_sha_t* OSSL_sha1;
 OSSL_sha_t* OSSL_sha256;
 OSSL_sha_t* OSSL_sha224;
@@ -439,6 +440,7 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     }
 
     /* Load the function symbols for OpenSSL Message Digest algorithms. */
+    OSSL_md5 = (OSSL_sha_t*)find_crypto_symbol(crypto_library, "EVP_md5");
     OSSL_sha1 = (OSSL_sha_t*)find_crypto_symbol(crypto_library, "EVP_sha1");
     OSSL_sha256 = (OSSL_sha_t*)find_crypto_symbol(crypto_library, "EVP_sha256");
     OSSL_sha224 = (OSSL_sha_t*)find_crypto_symbol(crypto_library, "EVP_sha224");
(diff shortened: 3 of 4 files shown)


closed/src/java.base/share/classes/sun/security/provider/NativeMD5.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (c) 2003, 2014, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
```
keithc-ca, line 4: Perhaps an Oracle copyright notice is inappropriate here?
@pshipton What do you think?

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -103,6 +103,9 @@ public final class SunEntries {
      */
     private static final boolean useNativeDigest = NativeCrypto.isAlgorithmEnabled("jdk.nativeDigest", "MessageDigest");
 
+    // Flag indicating whether the operating system is AIX.
+    private static final boolean isAIX = "AIX".equals(GetPropertyAction.privilegedGetProperty("os.name"));
+
```
keithc-ca, line 108: Rather than add yet another O/S test, suggest `import jdk.internal.util.OperatingSystem;` and use `OperatingSystem.isAix()` below.

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -287,6 +291,16 @@ public final class SunEntries {
          * Set the digest provider based on whether native crypto is
          * enabled or not.
          */
+        /* Disabling OpenSSL usage in AIX due to perfomance regression observed */
```
keithc-ca, line 294: This sentence should end with a period; suggest "on AIX" instead of "in AIX"
```java
        /* Disable use of OpenSSL on AIX due to an observed perfomance regression. */
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -601,6 +603,7 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     if ((NULL == OSSL_error_string) ||
         (NULL == OSSL_error_string_n) ||
         (NULL == OSSL_get_error) ||
+        (NULL == OSSL_md5) ||
```
keithc-ca, line 606: Support for md5 is optional in OpenSSL, so this should probably not be a fatal error that disables use of all other algorithms.

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: I'm not sure what you mean by optional. Are there versions of OpenSSL that don't have `MD5`?

closed/src/java.base/share/classes/sun/security/provider/NativeMD5.java
pshipton: It depends. Was any code from OpenJDK copied to this class, or copied and modified? Or is it all freshly written code that has no roots from OpenJDK code?

pshipton: If it's inappropriate here, then it likely shouldn't be in the other subclasses: NativeSHA, NativeSHA2, NativeSHA5. I'm not sure if it belongs in NativeDigest either.

src/java.base/share/classes/sun/security/provider/SunEntries.java
keithc-ca: Yes, we can try to find something equivalent for jdk17 and older versions when we get there.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: I was referring to the configuration flag `OPENSSL_NO_MD5` used in [`crypto/evp/c_alld.c`](https://github.com/openssl/openssl/blob/openssl-3.1/crypto/evp/c_alld.c#L22).
We shouldn't sacrifice all other uses of OpenSSL just because the library that's available was configured without md5 support.

closed/src/java.base/share/classes/sun/security/provider/NativeMD5.java
KostasTsiounis: I modified `NativeSHA2` (https://github.com/KostasTsiounis/openj9-openjdk-jdk/blob/4df64384b4ba176308e9fe2595a0c812ab490e04/closed/src/java.base/share/classes/sun/security/provider/NativeSHA2.java), so not sure what the appropriate course of action is here.

pshipton: Right, the easiest thing is to leave it as-is. I'm not sure it's a big deal either way.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: According to [this](https://manpages.opensuse.org/Tumbleweed/libressl-devel-doc/EVP_add_digest.3ssl.en.html), the `EVP_add_digest()` call adds `MD5` to a global lookup table so that it can be retrieved with `EVP_get_cipherbyname()`.

But in the code I get it directly through `EVP_md5()`. I am not sure the library loading will fail, even if `OPENSSL_NO_MD5` is set. What do you think?

KostasTsiounis: We use `EVP_md5()` (https://www.openssl.org/docs/man3.0/man3/EVP_md5.html) which is not deprecated. `MD5()` (https://www.openssl.org/docs/man3.0/man3/MD5.html) is the one that's actually marked for deprecation.

You are right, however, in that someone could build with `no-md5` and force OpenSSL to not export those, so I can definitely add a check for that.

Only problem is that same thing could be done for any of `aes`, `aria`, `bf`, `camellia`, `cast`, `des`, `dh`, `dsa`, `ec`, `hmac`, `idea`, `md2`, `md5`, `mdc2`, `rc2`, `rc4`, `rc5`, `ripemd`, `seed`, `sha`, `sm2`, `sm3`, `sm4`. Do you think we should make a note to add check for each of them in the future if possible rather than failing to load?

src/java.base/share/classes/sun/security/provider/SunEntries.java
KostasTsiounis: Changed to suggestion.

src/java.base/share/classes/sun/security/provider/SunEntries.java
KostasTsiounis: Fixed.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: I added an additional method to check whether MD5 is available instead of failing to load OpenSSL altogether.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -201,6 +202,8 @@ public void run() {
 
     private static final native long loadCrypto(boolean trace);
 
+    public static final native boolean isMD5Available();
```
keithc-ca, line 205: Suggest you generalize this as
```java
public static final native boolean isDigestAvailable(int algoIndex);
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: I don't think we're using any of the other algorithms listed above (yet?), so I don't think we need to do anything right now. If and when we add support for other algorithms, we should consider whether they may become deprecated and thus not available, or similarly, new algorithms that might not be present in older libraries.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: This is just for `MD5`. There are other digests like `SHA` that are available. Unless you mean adding checks for them too and not failing if they are not present.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: We do use `SHA` and `EC`. And it's not about deprecation, but rather building OpenSSL with `-no-<cipher>` like you mentioned for `MD5`.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
keithc-ca: Yes, I was suggesting that use of other digest algorithms should be conditional on their availability in the library present on the runtime system.

KostasTsiounis: Ok, so let's keep just for `MD5` for now and we'll add further functionality in future changes.

keithc-ca: In my opinion, keeping it as is represents a missed opportunity to avoid some future work, but I won't insist.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -855,6 +857,18 @@ JNIEXPORT void JNICALL JNI_OnUnload(JavaVM * vm, void * reserved)
...
+/* Check whether MD5 is available.
+ *
+ * Class:     jdk_crypto_jniprovider_NativeCrypto
+ * Method:    isMD5Available
+ * Signature: ()Z
+ */
+JNIEXPORT jboolean JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_isMD5Available
+  (JNIEnv *env, jclass thisObj)
+{
+    return (NULL != OSSL_md5);
```
keithc-ca, line 869: The name `thisObj` is inappropriate (this related to a `static` method); suggest 'clazz` or `thisClass`.

The return expression should explicitly use constants appropriate for `jboolean`:
```c
    return (NULL != OSSL_md5) ? JNI_TRUE : JNI_FALSE;
```

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -287,6 +289,17 @@ public final class SunEntries {
          * Set the digest provider based on whether native crypto is
          * enabled or not.
          */
+        /* Disabling OpenSSL usage on AIX due to perfomance regression observed. */
```
keithc-ca, line 292: The comment should have the same tone as the one above it, suggest (with typo corrected):
```
        /* Don't use native MD5 on AIX due to an observed performance regression. */
```

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c

src/java.base/share/classes/sun/security/provider/SunEntries.java
KostasTsiounis: Changed.

keithc-ca:

keithc-ca:
Grinder for jdk_security1 passed at https://openj9-jenkins.osuosl.org/job/Grinder/3447.

keithc-ca (approved):

KostasTsiounis:
FYI @keithc-ca

PR #773, opened 2024-04-15
Check setting of security property when in FIPS mode

KostasTsiounis:
When in `FIPS` mode, calling `Security.setProperty()` should not be allowed when the security property that the user is trying to override is set in the active `RestrictedSecurity` profile. A `SecurityException` is thrown if such an attempt is made.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -52,10 +52,12 @@ public final class RestrictedSecurity {
 
     private static final Debug debug = Debug.getInstance("semerufips");
 
-    // Restricted security mode enable check, only supported on Linux x64.
+    // Restricted security mode enable check.
     private static final boolean userEnabledFIPS;
     private static boolean isFIPSSupported;
     private static boolean isFIPSEnabled;
+
+    private static final boolean allowSetProperties;
 
     private static final boolean isNSSSupported;
     private static final boolean isOpenJCEPlusSupported;
@@ -70,6 +72,8 @@ public final class RestrictedSecurity {
     private static String userSecurityID;
 
     private static RestrictedSecurityProperties restricts;
+
+    private static final Set<String> unmodifiableProperties = new HashSet<>();
 
     private static final Map<String, List<String>> supportedPlatformsNSS = new HashMap<>();
     private static final Map<String, List<String>> supportedPlatformsOpenJCEPlus = new HashMap<>();
@@ -89,7 +93,8 @@ public final class RestrictedSecurity {
                         return new String[] { System.getProperty("semeru.fips"),
                                 System.getProperty("semeru.customprofile"),
                                 System.getProperty("os.name"),
-                                System.getProperty("os.arch") };
+                                System.getProperty("os.arch"),
+                                System.getProperty("semeru.fips.allowsetproperties") };
                     }
                 });
 
@@ -128,6 +133,7 @@ public final class RestrictedSecurity {
         isFIPSSupported = isNSSSupported;
 
         userEnabledFIPS = Boolean.parseBoolean(props[0]);
+        allowSetProperties = Boolean.parseBoolean(props[4]);
 
         if (userEnabledFIPS) {
             if (isFIPSSupported) {
@@ -368,6 +374,53 @@ public final class RestrictedSecurity {
     }
 
     /**
+     * Check whether a security property can be set.
+     *
+     * A security property that is set by a RestrictedSecurity profile,
+     * while FIPS security mode is enabled, cannot be reset programmatically.
+     *
+     * Every time an attempt to set a security property is made, a check is
+     * performed. If the above scenario holds true, a SecurityException is
+     * thrown.
+     *
+     * One can override this behaviour and allow the user to set any security
+     * property through the use of {@code -Dsemeru.fips.allowsetproperties=true}.
+     *
+     * @param key the security property that the user wants to set
+     * @throws SecurityException
+     *         if the security property is set by the profile and cannot
+     *         be altered
+     */
+    public static void checkSetSecurityProperty(String key) {
+        if (debug != null) {
+            debug.println("RestrictedSecurity: Checking whether property '"
+                    + key + "' can be set.");
+        }
+
+        /*
+         * Only disallow setting of security properties that are set by the active profile,
+         * if FIPS has been enabled.
+         *
+         * Allow any change, if the 'semeru.fips.allowsetproperties' flag is set to true.
+         */
+        if (unmodifiableProperties.contains(key)) {
+            if (debug != null) {
+                debug.println("RestrictedSecurity: Property '" + key + "' cannot be set.");
+                debug.println("If you want to override the check and allow all security"
+                        + "properties to be set, use '-Dsemeru.fips.allowsetproperties=true'.");
+                debug.println("BEWARE: You might not be FIPS compliant if you select to override!");
+            }
+            throw new SecurityException("FIPS mode: User-specified '" + key
+                    + "' cannot override profile definition.");
+        }
+
+        if (debug != null) {
+            debug.println("RestrictedSecurity: Property '"
+                    + key + "' can be set without issue.");
+        }
+    }
+
+    /**
      * Remove the security providers and only add restricted security providers.
      *
      * @param props the java.security properties
@@ -468,10 +521,10 @@ public final class RestrictedSecurity {
         // JDK properties name as key, restricted security properties value as value.
         propsMapping.put("jdk.tls.disabledNamedCurves", restricts.jdkTlsDisabledNamedCurves);
         propsMapping.put("jdk.tls.disabledAlgorithms", restricts.jdkTlsDisabledAlgorithms);
-        propsMapping.put("jdk.tls.ephemeralDHKeySize", restricts.jdkTlsDphemeralDHKeySize);
+        propsMapping.put("jdk.tls.ephemeralDHKeySize", restricts.jdkTlsEphemeralDHKeySize);
         propsMapping.put("jdk.tls.legacyAlgorithms", restricts.jdkTlsLegacyAlgorithms);
         propsMapping.put("jdk.certpath.disabledAlgorithms", restricts.jdkCertpathDisabledAlgorithms);
-        propsMapping.put("jdk.security.legacyAlgorithm", restricts.jdkSecurityLegacyAlgorithm);
+        propsMapping.put("jdk.security.legacyAlgorithms", restricts.jdkSecurityLegacyAlgorithms);
         String fipsMode = System.getProperty("com.ibm.fips.mode");
         if (fipsMode == null) {
             System.setProperty("com.ibm.fips.mode", restricts.jdkFipsMode);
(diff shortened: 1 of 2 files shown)


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -57,6 +57,8 @@ public final class RestrictedSecurity {
     private static boolean isFIPSSupported;
     private static boolean isFIPSEnabled;
 
+    private static final boolean allowSetSecurityProps;
```
keithc-ca, line 60: Suggest `allowSetProperties` is a better name; it more closely matches the name of the property from which it is derived.

While you're here, please correct the comment on line 55 (this applies to more than just Linux x64).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -71,6 +73,8 @@ public final class RestrictedSecurity {
 
     private static RestrictedSecurityProperties restricts;
 
+    private static List<String> alteredProperties = new ArrayList<>();
```
keithc-ca, line 76: This should be `final`, a set, and suggest `unmodifiableProperties` is a better name:
```java
    private static final Set<String> unmodifiableProperties = new HashSet<>();
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -367,6 +373,54 @@ private static void checkFIPSCompatibility(Properties props) {
         }
     }
 
+    /**
+     * Check whether security property can be set.
```
keithc-ca, line 377: Missing "a" in "whether a security".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -367,6 +373,54 @@ private static void checkFIPSCompatibility(Properties props) {
...
+     *
+     * A security property that is set by a RestrictedSecurity profile,
+     * while FIPS security mode is enabled, cannot be reset programmatically.
+     *
+     * Every time an attempt to set a security property is made, a check is
+     * performed. If the above scenario holds true, a SecurityException is
+     * thrown.
+     *
+     * One can override this behaviour and allow the user to set any security
+     * property through the use of `-Dsemeru.fips.allowsetproperties=true'.
```
keithc-ca, line 387: Suggest using `{@code ...}` or discuss setting the system property.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -367,6 +373,54 @@ private static void checkFIPSCompatibility(Properties props) {
...
+     *
+     * @param key the security property that the user wants to set
+     * @throws SecurityException
+     *         if the security property is set by the profile and cannot
+     *         be altered.
```
keithc-ca, line 392: Please remove the trailing period here.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -367,6 +373,54 @@ private static void checkFIPSCompatibility(Properties props) {
...
+         *
+         * Allow any change, if the 'semeru.fips.allowsetproperties' flag is set to true.
+         */
+        if (userEnabledFIPS && !allowSetSecurityProps && alteredProperties.contains(key)) {
+            if (debug != null) {
+                debug.println("RestrictedSecurity: Security property '"
+                        + key + "' cannot be set.");
+                debug.println("If you want to override the check and allow all security"
+                        + "properties to be set, use `-Dsemeru.fips.allowsetproperties=true'.");
+                debug.println("BEWARE: You might not be FIPS compliant if you select to override!");
```
keithc-ca, line 412: Suggest shortening "security property" to just "property" ("security" seem redundant immediately after "RestrictedSecurity"). Please avoid mismatched quotes (back-tick and `'`).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -367,6 +373,54 @@ private static void checkFIPSCompatibility(Properties props) {
...
+            }
+            throw new SecurityException("FIPS mode: User-specified '" + key
+                    + "' cannot override profile's.");
```
keithc-ca, line 415: Suggest "profile's" -> "profile definition".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -367,6 +373,54 @@ private static void checkFIPSCompatibility(Properties props) {
...
+        }
+
+        if (debug != null) {
+            debug.println("RestrictedSecurity: Security property '"
```
keithc-ca, line 419: Redundant "Security".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -488,6 +542,11 @@ private static void setProperties(Properties props) {
                 propsOldValue = "";
             }
 
+            if (propsNewValue != null) {
+                // Add to list of properties set by the active profile.
+                alteredProperties.add(jdkPropsName);
+            }
```
keithc-ca, line 548: No need to remember the property name unless modifying it is to be disallowed.
```java
            if ((propsNewValue != null) && userEnabledFIPS && !allowSetProperties) {
                // Add to set of properties set by the active profile.
                unmodifiableProperties.add(jdkPropsName);
            }
```

src/java.base/share/classes/java/security/Security.java
```diff
@@ -794,6 +794,10 @@ public static String getProperty(String key) {
      */
     public static void setProperty(String key, String datum) {
         check("setProperty." + key);
+
+        // Check whether the change to the property is allowed, if in FIPS mode.
```
keithc-ca, line 798: This need not speak of FIPS:
```java
        // Check whether setting the property is allowed.
```

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Changed both.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Used `{@code ...}`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Shortened and fixed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

src/java.base/share/classes/java/security/Security.java
KostasTsiounis: Removed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: It should say "set" instead of "list" now.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -128,6 +133,7 @@ public String[] run() {
         isFIPSSupported = isNSSSupported;
 
         userEnabledFIPS = Boolean.parseBoolean(props[0]);
+        allowSetSecurityProps = Boolean.parseBoolean(props[4]);
```
keithc-ca, line 136: The field was renamed `allowSetProperties`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -367,6 +373,54 @@ private static void checkFIPSCompatibility(Properties props) {
...
+     * property through the use of {@code -Dsemeru.fips.allowsetproperties=true}.
+     *
+     * @param key the security property that the user wants to set
+     * @throws SecurityException
+     *         if the security property is set by the profile and cannot
+     *         be altered
+     */
+    public static void checkSetSecurityProperty(String key) {
+        if (debug != null) {
+            debug.println("RestrictedSecurity: Checking whether security property '"
```
keithc-ca, line 396: Redundant: "security".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -367,6 +373,54 @@ private static void checkFIPSCompatibility(Properties props) {
...
+                    + key + "' can be set.");
+        }
+
+        /*
+         * Only disallow setting of security properties that are set by the active profile,
+         * if FIPS has been enabled.
+         *
+         * Allow any change, if the 'semeru.fips.allowsetproperties' flag is set to true.
+         */
+        if (userEnabledFIPS && !allowSetSecurityProps && unmodifiableProperties.contains(key)) {
```
keithc-ca, line 406: Only the final part of this test is needed:
```java
        if (unmodifiableProperties.contains(key)) {
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -367,6 +373,54 @@ private static void checkFIPSCompatibility(Properties props) {
...
+            if (debug != null) {
+                debug.println("RestrictedSecurity: Property '"
+                        + key + "' cannot be set.");
```
keithc-ca, line 409: This need not be split:
```java
                debug.println("RestrictedSecurity: Property '" + key + "' cannot be set.");
```

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Changed the comment too.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Sorry about that. Thought I refactored there. Fixed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Right. Cause we checked when adding to the set. Updated.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

keithc-ca:

KostasTsiounis:
FYI @keithc-ca

PR #774, opened 2024-04-16
Remove the unneeded Hotspot features for building OpenJ9

pshipton:
OpenJ9 builds don't need to include support for the removed Hotspot
features:
`cds compiler1 compiler2 jni-check jvmci jvmti minimal services static-build vm-structs zero`
and all the gc's.

In particular this removes `libsimdsort.so` on xlinux, which is unused.

With this change configure shows:
`checking JVM features to use for variant 'server'... 'dtrace jfr management'`

### DIFF
--- a/make/autoconf/jvm-features.m4
+++ b/make/autoconf/jvm-features.m4
@@ -23,7 +23,7 @@
 # questions.
 #
 # ===========================================================================
-# (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+# (c) Copyright IBM Corp. 2023, 2024 All Rights Reserved
 # ===========================================================================
 
 ###############################################################################
@@ -47,9 +47,8 @@
 m4_define(jvm_features_valid, m4_normalize( \
     ifdef([custom_jvm_features_valid], custom_jvm_features_valid) \
     \
-    cds compiler1 compiler2 dtrace epsilongc g1gc jfr jni-check \
-    jvmci jvmti link-time-opt management minimal opt-size parallelgc \
-    serialgc services shenandoahgc static-build vm-structs zero zgc \
+    dtrace jfr \
+    link-time-opt management opt-size \
 ))
 
 # Deprecated JVM features (these are ignored, but with a warning)
@@ -235,7 +234,10 @@ AC_DEFUN_ONCE([JVM_FEATURES_CHECK_CDS],
 [
   JVM_FEATURES_CHECK_AVAILABILITY(cds, [
     AC_MSG_CHECKING([if platform is supported by CDS])
-    if test "x$OPENJDK_TARGET_OS" = xaix; then
+    if true; then
+      AC_MSG_RESULT([no, OpenJ9])
+      AVAILABLE=false
+    elif test "x$OPENJDK_TARGET_OS" = xaix; then
       AC_MSG_RESULT([no, $OPENJDK_TARGET_OS-$OPENJDK_TARGET_CPU])
       AVAILABLE=false
     else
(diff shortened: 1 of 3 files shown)


pshipton:
jenkins compile xlinux jdknext

pshipton:
@keithc-ca I notice we already have https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/closed/autoconf/custom-spec.gmk.in#L187 but seems it's not working. I'll be interested to see if my change removes `libsimdsort.so` from the xlinux build.

keithc-ca:
> but seems it's not working

The line you quote is included before the definition modified here.

pshipton:
Do you have suggestions on fixing it, or should we accept this change? I also wonder in what version it broke.

pshipton:
This change doesn't fix the problem, libsimdsort.so is still present in the PR build.

Including the shared library depends on the compiler2 feature.
https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/make/modules/java.base/Lib.gmk#L214
https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/make/autoconf/jvm-features.m4#L542

keithc-ca:
This doesn't appear to prevent that library from being built (and included). It seems only relevant to the make target `create-buildjdk-interim-image` which doesn't work with or without this change.

I suggest we should disable unwanted hotspot features, effectively making the default configuration include `--disable-jvm-feature-compiler2`, etc.

pshipton:
Updated. I've built xlinux and confirmed libsimdsort.so is no longer present.
```
checking if JVM feature 'dtrace' is available... yes
checking if platform is supported by JVMCI... no, OpenJ9
checking if JVM feature 'jvmci' is available... no
checking if platform is supported by Shenandoah... yes
checking if JVM feature 'shenandoahgc' is available... yes
checking if static-build is enabled in configure... no, use --enable-static-build to enable static build.
checking if JVM feature 'static-build' is available... no
checking if platform is supported by ZGC... yes
checking if JVM feature 'zgc' is available... yes
checking JVM features to use for variant 'server'... 'dtrace jfr management'
```

make/autoconf/jvm-features.m4
```diff
@@ -47,9 +47,9 @@
 m4_define(jvm_features_valid, m4_normalize( \
     ifdef([custom_jvm_features_valid], custom_jvm_features_valid) \
     \
-    cds compiler1 compiler2 dtrace epsilongc g1gc jfr jni-check \
-    jvmci jvmti link-time-opt management minimal opt-size parallelgc \
-    serialgc services shenandoahgc static-build vm-structs zero zgc \
+    dtrace jfr \
+    link-time-opt management opt-size \
+    static-build \
```
keithc-ca, line 52: Do we want to remove `static-build` from this list as well? It's only available on macOS, but I don't know if it works with OpenJ9.

Suggest the explicit `JVM_FEATURES_CHECK_xxx` functions be updated as well to avoid misleading output from the configuration step: for example, it should not say
```
checking if platform is supported by CDS... yes
checking if JVM feature 'cds' is available... yes
```
instead, `JVM_FEATURES_CHECK_CDS` should be updated, like `JVM_FEATURES_CHECK_JVMCI` so it says
```
checking if platform is supported by CDS... no, OpenJ9
checking if JVM feature 'cds' is available... no
```
Similar changes should be made to `JVM_FEATURES_CHECK_SHENANDOAHGC` and `JVM_FEATURES_CHECK_ZGC`.

Now that CDS is disabled here, lines 682-684 in `custom-hook.m4` are redundant.

keithc-ca (changes requested):

make/autoconf/jvm-features.m4
pshipton: Updated.

keithc-ca:
Jenkins compile aix,osx,win,zlinux jdknext

keithc-ca:
Jenkins compile aix jdknext

keithc-ca (approved):

PR #777, opened 2024-04-18
Exclude libsimdsort from copyright scans

pshipton:
The feature is not included in OpenJ9 builds.

Related to https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/774

Tested via https://openj9-jenkins.osuosl.org/view/Infrastructure/job/CopyrightCheck-OpenJDK/1063
Previous builds without this change failed.


buildenv/jenkins/jobs/infrastructure/copyrightCheckDir.sh
```diff
@@ -325,6 +325,9 @@ check () {
     */share/classes/sun/security/util/math/intpoly/FieldGen.jsh)
       trace  "$1 is not part of the built JDK"
       IN_JDK=0;;
+    src/java.base/linux/native/libsimdsort/*)
+      trace  "$1 deemed not to be in the built JDK because it matches case parameter expansion src/java.base/linux/native/libsimdsort/* and the feature is not enabled"
+      IN_JDK=0;;
```
keithc-ca, line 330: Please remove the trailing whitespace here (and while you're at, the other three lines in this file).

Please also exclude `src/jdk.incubator.vector/*/native/libjsvml/*` which depends on the JVM feature `compiler2` which was removed in #774.

keithc-ca (changes requested):

buildenv/jenkins/jobs/infrastructure/copyrightCheckDir.sh
pshipton: I fixed the whitespace. I prefer not to exclude other files which aren't causing any failures. I removed the simd files because they don't have the CPE and cause the copyright jobs to fail.

keithc-ca (approved):

PR #780, opened 2024-04-19
Support restore with -XX:CRaCRestoreFrom=PATH

singh264:
Support restore with -XX:CRaCRestoreFrom=PATH that
specifies the path to the checkpoint image.

Issue: https://github.com/eclipse-openj9/openj9/issues/19261
Signed-off-by: Amarpreet Singh <Amarpreet.A.Singh@ibm.com>

### DIFF
--- a/closed/src/java.base/share/native/libjli/criuhelpers.c
+++ b/closed/src/java.base/share/native/libjli/criuhelpers.c
@@ -1,0 +1,303 @@
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * IBM designates this particular file as subject to the "Classpath" exception
+ * as provided by IBM in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * ===========================================================================
+ */
+
+#include "criuhelpers.h"
+
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+
+#include "java.h"
+#include <ctype.h>
+#include <sys/wait.h>
+
+/**
+ * Get the value of a command line option in the array of command line arguments.
+ * @param[in] optionName The name of the command line option to search for
+ * @param[in] argc The number of command line arguments
+ * @param[in] argv The array of command line arguments
+ * @param[out] error A pointer to an integer for the error code
+ * @return A pointer to the value of the command line option if found, NULL otherwise
+ */
+static const char *
+getCommandLineOptionValue(const char *optionName, int argc, char **argv, int *error)
+{
+    const char *value = NULL;
+    int i = 0;
+    size_t optionNameLength = strlen(optionName);
+    *error = -1;
+    for (i = argc - 1; i >= 0; i--) {
+        const char *arg = argv[i];
+        if (0 == strncmp(arg, optionName, optionNameLength)) {
+            const char *equals = arg + optionNameLength;
+            if (('=' == *equals) || ('\0' == *equals)) {
+                *error = 0;
+                value = equals;
+                if ('=' == *value) {
+                    value += 1;
+                }
+                if ('\0' == *value) {
+                    value = NULL;
+                }
+                break;
+            }
+        }
+    }
+    return value;
+}
+
+/**
+ * Get the checkpoint directory from command line options.
+ * @param[in] argc The number of command line arguments
+ * @param[in] argv The array of command line arguments
+ * @param[in/out] error A pointer to an integer for the error code
+ * @return A pointer to the checkpoint directory if found, NULL otherwise
+ */
+static const char *
+getCheckpointDirectory(int argc, char **argv, int *error)
+{
+    const char *checkpointDirectory = NULL;
+    const char *checkpointDirectoryPropertyValue = getCommandLineOptionValue("-XX:CRaCRestoreFrom", argc, argv, error);
+    if (0 == *error) {
+        if (NULL != checkpointDirectoryPropertyValue) {
+            checkpointDirectory = checkpointDirectoryPropertyValue;
+        } else {
+            JLI_ReportErrorMessage("The value of the command line option -XX:CRaCRestoreFrom was not found.");
+            *error = -2;
+        }
+    }
+    return checkpointDirectory;
+}
+
+/**
+ * Get the log level specified in the command line arguments.
+ * Valid log levels are from 0 to 4, inclusive; the default is 2.
+ * @param[in] argc The number of command line arguments
+ * @param[in] argv The array of command line arguments
+ * @param[in/out] error A pointer to an integer for the error code
+ * @return The log level integer if successful, default of 2 otherwise
+ */
+static int
+getLogLevel(int argc, char **argv, int *error)
+{
+    int logLevelValue = 2; /* default */
+    const char *logLevelPropertyValue = getCommandLineOptionValue("-Dopenj9.internal.criu.logLevel", argc, argv, error);
+    if (0 == *error) {
+        const char *c = logLevelPropertyValue;
+        if (NULL == c) {
+            goto done;
+        }
+        for (; '\0' != *c; c++) {
+            if (!isdigit(*c)) {
+                goto setLogLevelOptionValueNotValidError;
+            }
+        }
+        logLevelValue = atoi(logLevelPropertyValue);
+        if ((0 <= logLevelValue) && (logLevelValue <= 4)) {
+            goto done;
+        } else {
+            goto setLogLevelOptionValueNotValidError;
+        }
+    } else if (-1 == *error) {
+        goto done;
+    }
+setLogLevelOptionValueNotValidError:
+    JLI_ReportErrorMessage(
+            "The option '-Dopenj9.internal.criu.logLevel=%s' is not valid.",
+            logLevelPropertyValue);
+    *error = -2;
+done:
+    return logLevelValue;
+}
+
+/**
+ * Check if the unprivileged mode is specified in the command line arguments.
+ * @param[in] argc The number of command line arguments
+ * @param[in] argv The array of command line arguments
+ * @param[in/out] error A pointer to an integer for the error code
+ * @return true if the unprivileged mode option is specified in the command line arguments, false otherwise
+ */
+static jboolean
+isUnprivilegedModeOn(int argc, char **argv, int *error)
+{
+    jboolean isUnprivilegedModeOn = JNI_FALSE;
+    const char *unprivilegedModePropertyValue = getCommandLineOptionValue("-Dopenj9.internal.criu.unprivilegedMode", argc, argv, error);
+    if (0 == *error) {
+        if (NULL == unprivilegedModePropertyValue) {
+            isUnprivilegedModeOn = JNI_TRUE;
+        } else {
+            JLI_ReportErrorMessage(
+                    "The option '-Dopenj9.internal.criu.unprivilegedMode=%s' does not accept a value.",
+                    unprivilegedModePropertyValue);
+            *error = -2;
+        }
+    }
+    return isUnprivilegedModeOn;
+}
+
+/**
+ * Get the log file specified in the command line arguments.
+ * @param[in] argc The number of command line arguments
+ * @param[in] argv The array of command line arguments
+ * @param[in/out] error A pointer to an integer for the error code
+ * @return A pointer to the log file string if successful, NULL otherwise
+ */
+static const char *
+getLogFile(int argc, char **argv, int *error)
+{
+    const char *logFile = NULL;
+    const char *logFilePropertyValue = getCommandLineOptionValue("-Dopenj9.internal.criu.logFile", argc, argv, error);
+    if (0 == *error) {
+        if (NULL != logFilePropertyValue) {
+            logFile = logFilePropertyValue;
+        } else {
+            JLI_ReportErrorMessage("The option -Dopenj9.internal.criu.logFile requires a value.");
+            *error = -2;
+        }
+    }
+    return logFile;
+}
+
+/**
+ * Restore the system state from a checkpoint using the CRIU tool.
+ * @param[in] checkpointDirectory The directory containing the checkpoint data
+ * @param[in] logLevel The log level for CRIU logging
+ * @param[in] unprivilegedModeOn Indicates whether the unprivileged mode option is on
+ * @param[in] logFile The log file option for CRIU
+ * @return 0 if the execution of the 'criu restore' command succeeds, -1 otherwise
+ */
+static int
+restoreFromCheckpoint(const char *checkpointDirectory, int logLevel, jboolean unprivilegedModeOn, const char *logFile)
+{
+    int length = -1;
+    char logLevelOption[4] = { 0 };
+    char *logFileOption = NULL;
+    int argc = 0;
+    const char *argv[9] = { NULL };
+    argv[argc++] = "criu";
+    argv[argc++] = "restore";
+    argv[argc++] = "-D";
+    argv[argc++] = checkpointDirectory;
+    /* The log level is a single digit. */
+    snprintf(logLevelOption, sizeof(logLevelOption), "-v%d", logLevel);
+    argv[argc++] = logLevelOption;
+    argv[argc++] = "--shell-job";
+    if (unprivilegedModeOn) {
+        argv[argc++] = "--unprivileged";
+    }
+    if (NULL != logFile) {
+        length = strlen(logFile) + sizeof("--log-file=%s") - 1;
+        logFileOption = (char *)JLI_MemAlloc(length + 1);
+        if (NULL == logFileOption) {
+            JLI_ReportErrorMessage("Failed to allocate memory for option '--log-file=%s'.", logFile);
+            goto fail;
+        }
+        if (snprintf(logFileOption, length + 1, "--log-file=%s", logFile) < 0) {
+            JLI_ReportErrorMessage("Failed to format option '--log-file=%s'.", logFile);
+            goto fail;
+        }
+        argv[argc++] = logFileOption;
+    }
+    argv[argc] = NULL;
+    execvp(argv[0], (char * const *)argv);
+    /* If execvp returns, there was an error. */
+fail:
+    if (NULL != logFileOption) {
+        JLI_MemFree((void *)logFileOption);
+    }
+    return -1;
+}
+
+/**
+ * Handle the restoration of the system state from a checkpoint.
+ * @param[in] argc The number of command line arguments
+ * @param[in] argv The array of command line arguments
+ */
+void
+handleCRaCRestore(int argc, char **argv)
+{
+    const char *checkpointDirectory = NULL;
+    int error = 0;
+    int parentProcessExitStatus = EXIT_SUCCESS;
+    pid_t childProcessPid = 0;
+    int logLevel = 0;
+    int childProcessExitStatus = EXIT_SUCCESS;
+    jboolean unprivilegedModeOn = JNI_FALSE;
+    const char *logFile = NULL;
+    int childProcessPidStatus = 0;
+    int childProcessPidExitStatus = 0;
+    checkpointDirectory = getCheckpointDirectory(argc, argv, &error);
+    if (-1 == error) {
+        /* Option -XX:CRaCRestoreFrom not specified. */
+        return;
+    }
+    if (-2 == error) {
+        JLI_ReportErrorMessage("Failed to get the CRIU checkpoint directory.");
+        parentProcessExitStatus = EXIT_FAILURE;
+        goto doneParentProcess;
+    }
+    /*
+     * The if block will be invoked by the child process,
+     * and the else block will be invoked by the parent process.
+     */
+    childProcessPid = fork();
+    if (0 == childProcessPid) {
+        logLevel = getLogLevel(argc, argv, &error);
+        if (-2 == error) {
+            JLI_ReportErrorMessage("Failed to get the CRIU log level.");
+            childProcessExitStatus = EXIT_FAILURE;
+            goto doneChildProcess;
+        }
+        unprivilegedModeOn = isUnprivilegedModeOn(argc, argv, &error);
+        if (-2 == error) {
+            JLI_ReportErrorMessage("Failed to get the CRIU unprivileged mode.");
+            childProcessExitStatus = EXIT_FAILURE;
+            goto doneChildProcess;
+        }
+        logFile = getLogFile(argc, argv, &error);
+        if (-2 == error) {
+            JLI_ReportErrorMessage("Failed to get the CRIU log file.");
+            childProcessExitStatus = EXIT_FAILURE;
+            goto doneChildProcess;
+        }
+        childProcessExitStatus = restoreFromCheckpoint(checkpointDirectory, logLevel, unprivilegedModeOn, logFile);
+doneChildProcess:
+        exit(childProcessExitStatus);
+    } else {
+        waitpid(childProcessPid, &childProcessPidStatus, 0);
+        if (WIFEXITED(childProcessPidStatus)) {
+            childProcessPidExitStatus = WEXITSTATUS(childProcessPidStatus);
+            if (EXIT_SUCCESS == childProcessPidExitStatus) {
+                JLI_ReportMessage("Completed restore with -XX:CRaCRestoreFrom=PATH.");
+            } else {
+                JLI_ReportErrorMessage("Failed to restore from checkpoint, error=%d.", childProcessPidExitStatus);
+                parentProcessExitStatus = EXIT_FAILURE;
+            }
+        } else {
+            JLI_ReportErrorMessage("The CRIU restore child process failed.");
+            parentProcessExitStatus = EXIT_FAILURE;
+        }
+    }
+doneParentProcess:
+    exit(parentProcessExitStatus);
+}
+#endif /* defined(J9VM_OPT_CRAC_SUPPORT) */
--- a/closed/src/java.base/share/native/libjli/criuhelpers.h
+++ b/closed/src/java.base/share/native/libjli/criuhelpers.h
@@ -1,0 +1,34 @@
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * IBM designates this particular file as subject to the "Classpath" exception
+ * as provided by IBM in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * ===========================================================================
+ */
+
+#ifndef CRIUHELPERS_H
+#define CRIUHELPERS_H
+
+#include "j9cfg.h"
+
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+void handleCRaCRestore(int argc, char **argv);
+#endif /* defined(J9VM_OPT_CRAC_SUPPORT) */
+
+#endif /* CRIUHELPERS_H */
--- a/src/java.base/share/native/libjli/java.c
+++ b/src/java.base/share/native/libjli/java.c
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2024 All Rights Reserved
  * ===========================================================================
  */
 
@@ -59,6 +59,7 @@
 
 #include <assert.h>
 
+#include "criuhelpers.h"
 #include "java.h"
 #include "jni.h"
 
@@ -565,6 +566,10 @@ JavaMain(void* _args)
     jboolean isStaticMain;
     jfieldID noArgMainField;
     jboolean noArgMain;
+
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+    handleCRaCRestore(argc, argv);
+#endif /* defined(J9VM_OPT_CRAC_SUPPORT) */
 
     RegisterThread();
 


singh264:
@tajila @JasonFengJ9 requesting your review.

singh264:
> In addition, the code added in this change should be decorated with #if defined(J9VM_OPT_CRAC_SUPPORT), and j9cfg.h needs to be included.

How can I enable `J9VM_OPT_CRAC_SUPPORT`? It seems like `J9VM_OPT_CRAC_SUPPORT` should be enabled as CRAC support is enabled by default during the configure step.

JasonFengJ9:
> How can I enable J9VM_OPT_CRAC_SUPPORT? It seems like J9VM_OPT_CRAC_SUPPORT should be enabled as CRAC support is enabled by default during the configure step.

`J9VM_OPT_CRAC_SUPPORT` is enabled by default at `xa64` platform, for others, `--enable-crac-support` can enable `CRAC` support. 
Note: `CRAC` support requires `CRIU` support enabled at `xa64|xl64|xr64|xz64` by default or via `--enable-criu-support`.

src/java.base/share/native/libjli/java.c
```diff
@@ -58,9 +58,16 @@
...
 #include <assert.h>
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+#include <ctype.h>
+#include <sys/wait.h>
+#endif /* defined(J9VM_OPT_CRAC_SUPPORT) */
 
 #include "java.h"
 #include "jni.h"
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+#include "j9cfg.h"
```
JasonFengJ9, line 69: This `include` can't be enclosed within `#if defined(J9VM_OPT_CRAC_SUPPORT)`, it has `#define J9VM_OPT_CRAC_SUPPORT`, otherwise the rest of the code won't be included.

singh264:
@JasonFengJ9 re-requesting your review.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
     return 1;
 }
 
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+static char *getCommandLineOptionValue(const char *optionName, int argc, char **argv, int *error) {
+    *error = 0;
+    for (int i = 0; i < argc; i++) {
+        char *arg = argv[i];
+        if (0 == strncmp(arg, optionName, strlen(optionName))) {
```
JasonFengJ9, line 553: This might return an incorrect result when the arg is something like `-XX:CRaCRestoreFromExtra`.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+        return -1;
+    }
+    return 0;
+}
+
+static void handleCRaCRestore(int argc, char **argv) {
+    int error = 0;
+    char *checkpointDir = getCheckpointDir(argc, argv, &error);
+    if (OPTION_NOT_FOUND == error) {
+        freeCommandLineOptionValue(checkpointDir);
```
JasonFengJ9, line 747: Is it possible to return a non-null `checkpointDir` required to be freed when the `error` is `OPTION_NOT_FOUND`?

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+        if (isCommandLineOptionFoundWithError(error)) {
+            JLI_ReportErrorMessage("Failed to get the CRIU unprivileged mode, error=%d.", error);
+            freeCommandLineOptionValue(logLevel);
+            exit(EXIT_FAILURE);
+        }
+        char *logFile = getLogFile(argc, argv, &error);
+        if (isCommandLineOptionFoundWithError(error)) {
+            JLI_ReportErrorMessage("Failed to get the CRIU log file, error=%d.", error);
+            freeCommandLineOptionValue(logLevel);
+            freeCommandLineOptionValue(unprivilegedMode);
```
JasonFengJ9, line 775: The duplicated `freeCommandLineOptionValue()` code might be reduced if a cleanup label is added at the end of this method.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+        }
+    }
+    *error = OPTION_NOT_FOUND;
+    return NULL;
+}
+
+static void freeCommandLineOptionValue(char *value) {
+    if (value) {
+        JLI_MemFree(value);
+        value = NULL;
```
JasonFengJ9, line 582: This won't change the returning `value`.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+                        "The value %s of the command line option %s is not valid.",
+                        logLevelPropertyValue,
+                        logLevelPropertyName);
+                    freeCommandLineOptionValue(logLevelPropertyValue);
+                    *error = INVALID_OPTION_VALUE;
+                    return NULL;
+                }
+            }
+            int logLevelValue = atoi(logLevelPropertyValue);
+            if (0 <= logLevelValue && 4 >= logLevelValue) {
```
JasonFengJ9, line 632:
```suggestion
            if ((logLevelValue >= 0) && (logLevelValue <= 4)) {
```
Put the `rvalue` on the left only for comparing two values for equality or inequality.

keithc-ca: I suggest it reads more clearly expressed in the form `(min <= x) && (x <= max)`:
```c
            if ((0 <= logLevelValue) && (logLevelValue <= 4)) {
```

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
     return 1;
 }
 
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+static char *getCommandLineOptionValue(const char *optionName, int argc, char **argv, int *error) {
+    *error = 0;
+    for (int i = 0; i < argc; i++) {
```
keithc-ca, line 551: This isn't legal C code (variables cannot be declared in a `for` statement).

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+            }
+            return NULL;
+        }
+    }
+    *error = OPTION_NOT_FOUND;
+    return NULL;
+}
+
+static void freeCommandLineOptionValue(char *value) {
+    if (value) {
```
keithc-ca, line 580: Please make null checks explicit:
```c
    if (NULL != value) {
```

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+        JLI_MemFree(value);
+        value = NULL;
+    }
+}
+
+static jboolean isCommandLineOptionFoundWithError(int error) {
+    return OPTION_NOT_FOUND != error && error;
```
keithc-ca, line 587: Missing parentheses, comparison with zero should be explicit:
```c
return (OPTION_NOT_FOUND != error) && (0 != error);
```

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+}
+
+static char *getCheckpointDir(int argc, char **argv, int *error) {
+    const char *checkpointDirPropertyName = "-XX:CRaCRestoreFrom";
+    char *checkpointDirPropertyValue = getCommandLineOptionValue(checkpointDirPropertyName, argc, argv, error);
+    if (!*error) {
+        if (checkpointDirPropertyValue) {
```
keithc-ca, line 594: Null check should be eplicit.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+            *error = OPTION_VALUE_NOT_FOUND;
+            return NULL;
+        }
+    }
+    return NULL;
+}
+
+static char *getDefaultLogLevel(int *error) {
+    *error = 0;
+    char *defaultLogLevel = strdup("-v2");
```
keithc-ca, line 607: Not legal C code: variables must be declared at the beginning of a block.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+        *error = MEMORY_ALLOCATION_ERROR;
+    }
+    return defaultLogLevel;
+}
+
+static char *getLogLevel(int argc, char **argv, int *error) {
+    const char *logLevelPropertyName = "-Dopenj9.internal.criu.logLevel";
+    char *logLevelPropertyValue = getCommandLineOptionValue(logLevelPropertyName, argc, argv, error);
+    if (!isCommandLineOptionFoundWithError(*error)) {
+        if (logLevelPropertyValue) {
```
keithc-ca, line 619: Null check should be explicit.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+            for (const char *c = logLevelPropertyValue; *c; c++) {
```
keithc-ca, line 620: Can't declare a variable here, comparison with `'\0'` should be explicit.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +545,260 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+            int logLevelValue = atoi(logLevelPropertyValue);
+            if (0 <= logLevelValue && 4 >= logLevelValue) {
+                int logLevelLength = snprintf(NULL, 0, "-v%d", logLevelValue);
+                if (0 > logLevelLength) {
+                    JLI_ReportErrorMessage("Failed to calculate the length of the CRIU log level string.");
+                    freeCommandLineOptionValue(logLevelPropertyValue);
+                    *error = MEMORY_ALLOCATION_ERROR;
+                    return NULL;
+                }
+                char *logLevel = (char *)JLI_MemAlloc(logLevelLength + 1);
```
keithc-ca, line 640: Can't declare a local here.

keithc-ca (changes requested):
Several comments apply multiple other places.

src/java.base/share/native/libjli/java.c
singh264: How can I add the appropriate guideline to the coding standard? The following line of code seems to align closely with the [omr coding standard](https://github.com/eclipse/omr/blob/master/doc/CodingStandard.md): 

```
if ((logLevelValue >= 0) && (logLevelValue <= 4)) {
```

src/java.base/share/native/libjli/java.c
JasonFengJ9: Is `freeCommandLineOptionValue(checkpointDir)` still needed if that's the case?

singh264: > It is not possible to return a non-null checkpointDir required to be freed when the error is OPTION_NOT_FOUND.

> Is freeCommandLineOptionValue(checkpointDir) still needed if that's the case?

`freeCommandLineOptionValue(checkpointDir)` is not needed if that's the case, but it seems like we could keep it for consistency or future modifications to `getCheckpointDir` that might return non-NULL values even when `OPTION_NOT_FOUND` occurs.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +556,318 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+        return;
+    }
+    if (hasGetCommandLineOptionValueFoundError(error)) {
+        JLI_ReportErrorMessage("Failed to get the CRIU checkpoint directory, error=%d.", error);
+        parentProcessExitStatus = EXIT_FAILURE;
+        goto exitParentProcessHandleCRaCRestore;
+    }
+    unsupportedOption = getUnsupportedCommandLineOption(argc, argv, &error);
+    if (OPTION_NOT_SUPPORTED == error) {
+        JLI_ReportErrorMessage("The command line option %s is not supported for CRaC restore.", unsupportedOption);
```
JasonFengJ9, line 816: This probably won't work if other non-CRIU but valid command line options exist.

singh264: How can I determine if it is appropriate to remove the `OPTION_NOT_SUPPORTED` check given that restore occurs before jvm initialization?

JasonFengJ9: The options checked here are for `criu restore`, not for JVM, I think any unrecognized options can be ignored instead.

singh264:
@JasonFengJ9 @keithc-ca how can I re-request your review?

JasonFengJ9:
> how can I re-request your review?

@singh264 you already did, I will have a look today.

src/java.base/share/native/libjli/java.c
```diff
@@ -72,6 +77,46 @@
...
+#define CRIU_RESTORE_MAX_ARGS 9
+#define CRIU_COMMAND "criu"
+#define CRIU_RESTORE_OPTION "restore"
+#define CRIU_CHECKPOINT_DIRECTORY_OPTION "-D"
+#define CRIU_DEFAULT_LOG_LEVEL_OPTION "-v2"
+#define CRIU_UNPRIVILEGED_MODE_OPTION "--unprivileged"
+#define CRIU_SHELL_JOB_OPTION "--shell-job"
+#define EXECVP_ERROR -1
+#define EXECVP_SUCCESS 0
+#define WAIT_INDEFINITELY 0
```
keithc-ca, line 118: I'd like to see more order here: the names are generic and should be chosen so their intent, classification and usage should be clear.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +579,316 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+/**
+ * Get the value of a command line option in the array of command line arguments.
+ * @param[in] optionName The name of the command line option to search for
+ * @param[in] argc The number of command line arguments
+ * @param[in] argv The array of command line arguments
+ * @param[out] error A pointer to an integer for the error code
+ * @return A pointer to the value of the command line option if found, NULL otherwise
+ */
+static char *getCommandLineOptionValue(const char *optionName, int argc, char **argv, int *error) {
```
keithc-ca, line 591: This should return `const char *` (callers should not be modifying the target of the pointer returned). I see no reason why this needs to allocate space for a copy of the option value; then we don't need the `error` parameter, nor would we need to `free()` those strings later.

The type, name and signature and opening brace should be split across three lines:
```c
static const char *
getCommandLineOptionValue(const char *optionName, int argc, char **argv, int *error)
{
```
The same formatting applies to all new functions added.

src/java.base/share/native/libjli/java.c
keithc-ca: Please only use C syntax.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +579,316 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+                    }
+                }
+                break;
+            }
+        }
+    }
+    if (!optionNameFound) {
+        *error = OPTION_NAME_NOT_FOUND_ERROR;
+    }
+    return value;
```
keithc-ca, line 621: This does not return `NULL` if not found like the documentation says.

keithc-ca (changes requested):

src/java.base/share/native/libjli/java.c
JasonFengJ9: @singh264 please address this and the following comments.

singh264: It seems like I will address this and the following comments next week.

src/java.base/share/native/libjli/java.c
singh264: @JasonFengJ9 @keithc-ca how can I know the answer to the following question: how can I add the appropriate guideline to the coding standard?

singh264:
@JasonFengJ9 @keithc-ca how can I re-request your review?

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +588,391 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+ */
+static int
+calculateFormattedStringLength(const char *format, const char *value)
+{
+    int length = -1;
+    if (0 == strcmp(format, LOG_LEVEL_OPTION_FORMAT)) {
+        length = snprintf(NULL, 0, LOG_LEVEL_OPTION_FORMAT, atoi(value));
+    } else if (0 == strcmp(format, LOG_FILE_OPTION_FORMAT)) {
+        length = snprintf(NULL, 0, LOG_FILE_OPTION_FORMAT, value);
+    }
```
keithc-ca, line 779: These uses of `snprintf()` should be inlined where this function is used.

keithc-ca (changes requested):

singh264:
@JasonFengJ9 @keithc-ca how can I re-request your review?

src/java.base/share/native/libjli/java.c
singh264: @tajila @JasonFengJ9 @keithc-ca how can I know the answer to the following question: how can I add the appropriate guideline to the coding standard?

keithc-ca: You could open a pull request https://github.com/eclipse/omr with proposed changes to https://github.com/eclipse/omr/blob/master/doc/CodingStandard.md.

singh264: I would like to provide a [pull request](https://github.com/eclipse/omr/pull/7363) with the proposed changes to https://github.com/eclipse/omr/blob/master/doc/CodingStandard.md.

singh264:
@JasonFengJ9 @keithc-ca how can I re-request your review?

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +590,372 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+
+/**
+ * Check if an error occurred while retrieving a command line option value.
+ * @param[in] error The error code of the result of obtaining the command line option value
+ * @return true if an error occurred, false otherwise
+ */
+static jboolean
+hasGetCommandLineOptionValueFoundError(int error)
+{
+    return (0 != error) ? JNI_TRUE : JNI_FALSE;
```
keithc-ca, line 643: I don't think this function is worthwhile.

keithc-ca (changes requested):

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +590,372 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+        goto returnCreateCommandOption;
+    }
+    /*
+     * We utilize the format macros to ensure a literal is passed to snprintf
+     * and avoid the error caused by the -Wformat-nonliteral warning.
+     */
+    if (0 == strcmp(format, LOG_LEVEL_OPTION_FORMAT)) {
+        length = snprintf(NULL, 0, LOG_LEVEL_OPTION_FORMAT, atoi(value));
+    } else if (0 == strcmp(format, LOG_FILE_OPTION_FORMAT)) {
+        length = snprintf(NULL, 0, LOG_FILE_OPTION_FORMAT, value);
```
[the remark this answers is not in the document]
singh264: How can I know request more details as it seems like I do not understand the objection to format decoding here?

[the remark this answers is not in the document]
keithc-ca: This function is only expected to be called where `format` is `LOG_LEVEL_OPTION_FORMAT` or `LOG_FILE_OPTION_FORMAT`. Using a string to distinguish those cases, besides being inefficient, is unusual and hard to follow. There isn't much to this function, so I think the necessary parts should be inlined into those call sites.

src/java.base/share/native/libjli/java.c
```diff
@@ -72,6 +77,49 @@
...
+#define LOG_FILE_OPTION_FORMAT "--log-file=%s"
+
+/* The CRIU command options. */
+#define CRIU_COMMAND "criu"
+#define CRIU_RESTORE_OPTION "restore"
+#define CRIU_CHECKPOINT_DIRECTORY_OPTION "-D"
+#define CRIU_DEFAULT_LOG_LEVEL_OPTION "-v2"
+#define CRIU_SHELL_JOB_OPTION "--shell-job"
+#define CRIU_UNPRIVILEGED_MODE_OPTION "--unprivileged"
+#endif /* defined(J9VM_OPT_CRAC_SUPPORT) */
```
keithc-ca, line 122: These are (or will be) unused and should simply be removed:
* `COMMAND_OPTION_LENGTH_CALCULATION_ERROR`
* `COMMAND_OPTION_MEMORY_ALLOCATION_ERROR_MESSAGE`
* `LOG_FILE_OPTION_FORMAT`
* `LOG_LEVEL_OPTION_FORMAT`
* `MEMORY_ALLOCATION_ERROR`

Many others are used only once: I suggest it's clearer, for example, to just write `"--unprivileged"` rather than hide that text in `CRIU_UNPRIVILEGED_MODE_OPTION`, define a long from its use.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +582,329 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+ */
+static const char *
+getCommandLineOptionValue(const char *optionName, int argc, char **argv, int *error)
+{
+    const char *value = NULL;
+    int i = 0;
+    size_t optionNameLength = strlen(optionName);
+    jboolean optionNameFound = JNI_FALSE;
+    *error = 0;
+    for (; i < argc; i++) {
```
keithc-ca, line 602: Even thought it is redundant here, specifying the initial value is good practice:
```c
    for (i = 0; i < argc; i++) {
```
The first value of `i` is quite clear, and you don't need to go looking elsewhere.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +582,329 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+        const char *arg = argv[i];
+        if (0 == strncmp(arg, optionName, optionNameLength)) {
+            const char *equals = arg + optionNameLength;
+            if (('=' == *equals) || ('\0' == *equals)) {
+                optionNameFound = JNI_TRUE;
+                value = equals;
+                if ('=' == *value) {
+                    value = value + 1;
```
keithc-ca, line 610: Suggest this should use `+=`:
```c
                    value += 1;
```

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +582,329 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+/**
+ * Get the log level specified in the command line arguments.
+ * @param[in] argc The number of command line arguments
+ * @param[in] argv The array of command line arguments
+ * @param[in/out] error A pointer to an integer for the error code
+ * @return A pointer to the log level string if successful, NULL otherwise
+ */
+static const char *
+getLogLevel(int argc, char **argv, int *error)
+{
```
keithc-ca, line 668: Suggest this should return `int`, avoiding the need to call `atoi()` other places.
```c
static int
getLogLevel(int argc, char **argv, int *error)
{
    int logLevelValue = 2; /* default */
    const char *logLevelPropertyValue = getCommandLineOptionValue(LOG_LEVEL_OPTION_NAME, argc, argv, error);
    if (!isCommandLineOptionFoundWithError(*error)) {

        /* [elided] */

        if ((0 <= logLevelValue) && (logLevelValue <= 4)) {
            goto done;
        }
    }
setLogLevelOptionValueNotValidError:
    JLI_ReportErrorMessage(
            "The value of the command line option %s is not valid, value=%s.",
            LOG_LEVEL_OPTION_NAME,
            (NULL != logLevelPropertyValue) ? logLevelPropertyValue : "");
    *error = OPTION_VALUE_NOT_VALID_ERROR;
done:
    return logLevelValue;
}
```

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +582,329 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+ * @param[in/out] option A pointer to a pointer to a command option
+ */
+static void
+freeCommandOption(char **option)
+{
+    if ((NULL != option) && (NULL != *option)) {
+        JLI_MemFree((void *)*option);
+        *option = NULL;
+   }
+}
```
keithc-ca, line 755: Other changes will eliminate the need for this function.

src/java.base/share/native/libjli/java.c
```diff
@@ -534,6 +582,329 @@ invokeInstanceMainWithoutArgs(JNIEnv *env, jclass mainClass) {
...
+            JLI_ReportErrorMessage(COMMAND_OPTION_MEMORY_ALLOCATION_ERROR_MESSAGE, logFile, LOG_FILE_OPTION_FORMAT);
+            restoreStatus = MEMORY_ALLOCATION_ERROR;
+            goto freeLogFileOption;
+        }
+        argv[argc++] = logFileOption;
+    }
+    argv[argc] = NULL;
+    if (EXECVP_ERROR == execvp(argv[0], (char * const *)argv)) {
+        restoreStatus = EXECVP_ERROR;
+    }
```
keithc-ca, line 827: It's not clear there's any value in checking the return value from `execvp()` - it doesn't return when successful.

keithc-ca (changes requested):

src/java.base/share/native/libjli/java.c
singh264: How can I request clarification as it seems like the suggestion is to remove all the macros by hardcoding the strings and I do not understand the following part of the [comment](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/780#discussion_r1638816318):

```
 ..., define a long from its use.
```

?

src/java.base/share/native/libjli/java.c
singh264: How can I request clarification as it seems like the good practice is to initialize `i` at the start of the function during declaration and with the for loop in the following way:

```c
    for (i = 0; i < argc; i++) {
```

even though the first value of `i` is quite clear, and we don't need to go looking elsewhere

?

singh264:
@tajila @JasonFengJ9 @keithc-ca how can I re-request your review?

src/java.base/share/native/libjli/java.c
singh264: @tajila @keithc-ca how can I know the answer to the following questions: how can I request clarification as it seems like the suggestion is to remove all the macros by hardcoding the strings and I do not understand the following part of the [comment](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/780#discussion_r1638816318):

```
 ..., define a long from its use.
```

?

src/java.base/share/native/libjli/java.c
singh264: @tajila @keithc-ca how can I know the answer to the following question: how can I request clarification as it seems like the good practice is to initialize i at the start of the function during declaration and with the for loop in the following way:

```c
    for (i = 0; i < argc; i++) {
```

even though the first value of i is quite clear, and we don't need to go looking elsewhere

?

keithc-ca: I wasn't suggesting that you remove the initial value from the declaration.
The optimizer will eliminate any actual redundant assignments, but better to include the initial value in each `for` loop. If a second loop were added later, it would need the initializer clause, but shouldn't be coded differently than this one.

src/java.base/share/native/libjli/java.c
singh264: It seems like I have addressed the comment.

singh264:
@tajila @JasonFengJ9 @keithc-ca how can I re-request your review?

JasonFengJ9:
Quite a few codes are being added in this `java.c` file, I  would suggest moving those `CRIU/CRaC` specific helper methods into a `closed` file such as  `closed/src/java.base/share/native//libjli/criuhelpers.c`.

singh264:
How can I fix the following [Copyright Check](https://openj9-jenkins.osuosl.org/job/PullRequest-CopyrightCheck-OpenJDK/1623/console) 

```
18:39:39  ###################################
[Pipeline] echo
18:39:39  The following files were modified and have incorrect copyrights
[Pipeline] echo
18:39:39  closed/src/java.base/share/native/libjli/criuhelpers.c
18:39:39  	IBM Copyright should contain the GPLv2 license
18:39:39  	IBM Copyright should contain the IBM Classpath Exception
[Pipeline] echo
18:39:39  closed/src/java.base/share/native/libjli/criuhelpers.h
18:39:39  	IBM Copyright should contain the GPLv2 license
18:39:39  	IBM Copyright should contain the IBM Classpath Exception
[Pipeline] echo
18:39:39  ###################################
```

for `closed/src/java.base/share/native/libjli/criuhelpers.h` and `closed/src/java.base/share/native/libjli/criuhelpers.c`?

JasonFengJ9:
> How can I fix the following [Copyright Check](https://openj9-jenkins.osuosl.org/job/PullRequest-CopyrightCheck-OpenJDK/1623/console)

https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/0b87eb4e968b8abe041eb9e551b3e9f7ffdba045/closed/src/java.base/share/native/libjncrypto/NativeCrypto_md.h#L1-L23
Note: change the `2019, 2022` to `2024, 2024` for a new file.

singh264:
How can I confirm that the changes need to be backported to the following repositories

- https://github.com/ibmruntimes/openj9-openjdk-jdk23
- https://github.com/ibmruntimes/openj9-openjdk-jdk22
- https://github.com/ibmruntimes/openj9-openjdk-jdk21
- https://github.com/ibmruntimes/openj9-openjdk-jdk20
- https://github.com/ibmruntimes/openj9-openjdk-jdk19
- https://github.com/ibmruntimes/openj9-openjdk-jdk18
- https://github.com/ibmruntimes/openj9-openjdk-jdk17
- https://github.com/ibmruntimes/openj9-openjdk-jdk16
- https://github.com/ibmruntimes/openj9-openjdk-jdk15
- https://github.com/ibmruntimes/openj9-openjdk-jdk11

to support restore with -XX:CRaCRestoreFrom=PATH?

keithc-ca:
It's premature to worry about back-ports until this change has converged on something we can all agree upon, but when the time comes, we'll want to back-port to all active versions that support CRaC (which currently means jdk23, jdk22, jdk21, jdk17 and jdk11).

singh264:
How can I know when the change has converged on something we can all agree upon to backport to jdk23, jdk22, jdk21, jdk17 and jdk11?

tajila:
@keithc-ca @JasonFengJ9 is there any more feedback on these changes?

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,334 @@
...
+    const char *argv[9] = { NULL };
+    argv[argc++] = "criu";
+    argv[argc++] = "restore";
+    argv[argc++] = "-D";
+    argv[argc++] = checkpointDirectory;
+    length = snprintf(NULL, 0, "-v%d", logLevel);
+    if (length < 0) {
+        char logLevelString[2] = { 0 };
+        sprintf(logLevelString, "%d", logLevel);
+        JLI_ReportErrorMessage("Failed to calculate the length of the command option, value=%s, format=%s.", logLevelString, "-v%d");
```
keithc-ca, line 200: There's no need for `sprintf()` here, the message can use `%d` with `logLevel`:
```c
        JLI_ReportErrorMessage("Failed to calculate the length of option '-v%d'.", logLevel);
```
I don't think we need to allocate memory for `-vN`, a local array is fine (given we need at most four characters).

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,334 @@
...
+    if (unprivilegedModeOn) {
+        argv[argc++] = "--unprivileged";
+    }
+    if (NULL != logFile) {
+        length = snprintf(NULL, 0, "--log-file=%s", logFile);
+        if (length < 0) {
+            JLI_ReportErrorMessage("Failed to calculate the length of the command option, value=%s, format=%s.", logFile, "--log-file=%s");
+            restoreStatus = -1;
+            goto freeLogLevelOption;
+        }
```
keithc-ca, line 230: This is better expressed using `strlen()` which won't fail:
```c
        length = strlen(logFile) + sizeof("--log-file=%s") - 1;
```

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,334 @@
...
+        logFileOption = (char *)JLI_MemAlloc(length + 1);
+        if (NULL == logFileOption) {
+            JLI_ReportErrorMessage("Failed to allocate memory for the command option, value=%s, format=%s.", logFile, "--log-file=%s");
```
keithc-ca, line 233: Suggest inlining the option name (here and elsewhere):
```c
            JLI_ReportErrorMessage("Failed to allocate memory for option '--log-file=%s'", logFile);
```

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,334 @@
...
+            restoreStatus = -1;
+            goto freeLogLevelOption;
+        }
+        if (snprintf(logFileOption, length + 1, "--log-file=%s", logFile) < 0) {
+            JLI_ReportErrorMessage("Failed to allocate memory for the command option, value=%s, format=%s.", logFile, "--log-file=%s");
```
keithc-ca, line 238: This message is wrong; it isn't an allocation that failed:
```c
            JLI_ReportErrorMessage("Failed to format option '--log-file=%s'.", logFile);
```

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,334 @@
...
+    execvp(argv[0], (char * const *)argv);
+    /* If execvp returns, there was an error. */
+    restoreStatus = -1;
+freeLogFileOption:
+    if (logFileOption != NULL) {
+        JLI_MemFree((void *)logFileOption);
+        logFileOption = NULL;
+    }
+freeLogLevelOption:
+    if (logLevelOption != NULL) {
```
keithc-ca, line 254: Constants on the left of `!=` please (lines 249 and 254).

closed/src/java.base/share/native/libjli/criuhelpers.h
```diff
@@ -0,0 +1,59 @@
...
+
+static const char *
+getLogFile(int argc, char **argv, int *error);
+
+static int
+restoreFromCheckpoint(const char *checkpointDirectory, int logLevel, jboolean unprivilegedModeOn, const char *logFile);
+
+void
+handleCRaCRestore(int argc, char **argv);
+#endif /* defined(J9VM_OPT_CRAC_SUPPORT) */
```
keithc-ca, line 57: Header files should not declare `static` features: they belong in the relevant C/C++ source file.
This should include only what is necessary for the declarations here.
The result should be just:
```c
#include "j9cfg.h"

#if defined(J9VM_OPT_CRAC_SUPPORT)
void
handleCRaCRestore(int argc, char **argv);
#endif /* defined(J9VM_OPT_CRAC_SUPPORT) */
```

src/java.base/share/native/libjli/java.c
```diff
@@ -61,6 +61,7 @@
 
 #include "java.h"
 #include "jni.h"
+#include "criuhelpers.h"
```
keithc-ca, line 64: Please keep includes sorted:
```c
#include "criuhelpers.h"
#include "java.h"
#include "jni.h"
```

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,334 @@
...
+    int logLevel = 0;
+    int childProcessExitStatus = EXIT_SUCCESS;
+    jboolean unprivilegedModeOn = JNI_FALSE;
+    const char *logFile = NULL;
+    int childProcessPidStatus = 0;
+    int childProcessPidExitStatus = 0;
+    checkpointDirectory = getCheckpointDirectory(argc, argv, &error);
+    if (-1 == error) {
+        return;
+    }
```
keithc-ca, line 283: A comment would be welcome here, explaining that if the restore option is not specified.

keithc-ca (changes requested):

closed/src/java.base/share/native/libjli/criuhelpers.h
singh264: How can I keep the includes for `java.h`, `<ctype.h>` and `<sys/wait.h>` and avoid the compilation errors that result when I keep just 

```c
#include "j9cfg.h"

#if defined(J9VM_OPT_CRAC_SUPPORT)
void
handleCRaCRestore(int argc, char **argv);
#endif /* defined(J9VM_OPT_CRAC_SUPPORT) */
```

?

closed/src/java.base/share/native/libjli/criuhelpers.h
```diff
@@ -0,0 +1,38 @@
...
+
+#ifndef CRIUHELPERS_H
+#define CRIUHELPERS_H
+
+#include "java.h"
+#include "j9cfg.h"
+
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+#include <ctype.h>
+#include <sys/wait.h>
```
keithc-ca, line 33: The only `#include` that belongs here is `j9cfg.h`. If needed, the other `#include` directives should be added to the `.c` file.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+ * @param[in] argc The number of command line arguments
+ * @param[in] argv The array of command line arguments
+ * @param[out] error A pointer to an integer for the error code
+ * @return A pointer to the value of the command line option if found, NULL otherwise
+ */
+static const char *
+getCommandLineOptionValue(const char *optionName, int argc, char **argv, int *error)
+{
+    const char *value = NULL;
+    int i = argc - 1;
```
keithc-ca, line 40: The initial value should be zero, even though it will be unused.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+ * @param[in] argv The array of command line arguments
+ * @param[in/out] error A pointer to an integer for the error code
+ * @return The log level integer if successful, default of 2 otherwise
+ */
+static int
+getLogLevel(int argc, char **argv, int *error)
+{
+    int logLevelValue = 2; /* default */
+    const char *logLevelPropertyValue = NULL;
+    logLevelPropertyValue = getCommandLineOptionValue("-Dopenj9.internal.criu.logLevel", argc, argv, error);
```
keithc-ca, line 98: Please join these:
```c
    const char *logLevelPropertyValue = getCommandLineOptionValue("-Dopenj9.internal.criu.logLevel", argc, argv, error);
```

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+        if ((0 <= logLevelValue) && (logLevelValue <= 4)) {
+            goto done;
+        } else {
+            goto setLogLevelOptionValueNotValidError;
+        }
+    } else if (-1 == *error) {
+        goto done;
+    }
+setLogLevelOptionValueNotValidError:
+    JLI_ReportErrorMessage("The value of the command line option is not valid, option=-Dopenj9.internal.criu.logLevel, value=%s.", logLevelPropertyValue);
```
keithc-ca, line 119: This line is too long, please fold. The message can also be more concise: e.g. the word "option" need appear only once, as in
```c
    JLI_ReportErrorMessage(
            "The option '-Dopenj9.internal.criu.logLevel=%s' is not valid.",
            logLevelPropertyValue);
```

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+static jboolean
+isUnprivilegedModeOn(int argc, char **argv, int *error)
+{
+    jboolean isUnprivilegedModeOn = JNI_FALSE;
+    const char *unprivilegedModePropertyValue = getCommandLineOptionValue("-Dopenj9.internal.criu.unprivilegedMode", argc, argv, error);
+    if (0 == *error) {
+        if (NULL == unprivilegedModePropertyValue) {
+            isUnprivilegedModeOn = JNI_TRUE;
+        } else {
+            JLI_ReportErrorMessage("The value of the command line option is not expected, option=-Dopenj9.internal.criu.unprivilegedMode, value=%s.", unprivilegedModePropertyValue);
```
keithc-ca, line 141: Too long, can be more concise:
```c
            JLI_ReportErrorMessage(
                    "The option '-Dopenj9.internal.criu.unprivilegedMode=%s' does not accept a value.",
                    unprivilegedModePropertyValue);
```

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+static const char *
+getLogFile(int argc, char **argv, int *error)
+{
+    const char *logFile = NULL;
+    const char *logFilePropertyValue = getCommandLineOptionValue("-Dopenj9.internal.criu.logFile", argc, argv, error);
+    if (0 == *error) {
+        if (NULL != logFilePropertyValue) {
+            logFile = logFilePropertyValue;
+        } else {
+            JLI_ReportErrorMessage("The value of the command line option -Dopenj9.internal.criu.logFile was not found.");
```
keithc-ca, line 164: Suggest (more concise and consistent):
```c
            JLI_ReportErrorMessage("The option -Dopenj9.internal.criu.logFile requires a value.");
```

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+    if (length < 0) {
+        JLI_ReportErrorMessage("Failed to calculate the length of option '-v%d'.", logLevel);
+        restoreStatus = -1;
+        goto done;
+    }
+    if (snprintf(logLevelOption, sizeof(logLevelOption), "-v%d", logLevel) < 0) {
+        JLI_ReportErrorMessage("Failed to format the log level option '-v%d'.", logLevel);
+        restoreStatus = -1;
+        goto done;
+    }
```
keithc-ca, line 202: One call to `snprintf()` is sufficient (we know it will fit).
```c
    snprintf(logLevelOption, sizeof(logLevelOption), "-v%d", logLevel);
```

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+    argv[argc++] = logLevelOption;
+    argv[argc++] = "--shell-job";
+    if (unprivilegedModeOn) {
+        argv[argc++] = "--unprivileged";
+    }
+    if (NULL != logFile) {
+        length = strlen(logFile) + sizeof("--log-file=%s") - 1;
+        logFileOption = (char *)JLI_MemAlloc(length - 1);
```
keithc-ca, line 210: The argument must be `length + 1` (to account for the NUL-terminator); here and on line 216.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+        if (NULL == logFileOption) {
+            JLI_ReportErrorMessage("Failed to allocate memory for option '--log-file=%s'.", logFile);
+            restoreStatus = -1;
+            goto done;
```
keithc-ca, line 214: Suggest `fail` is a better label name.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+ * @param[in] checkpointDirectory The directory containing the checkpoint data
+ * @param[in] logLevel The log level for CRIU logging
+ * @param[in] unprivilegedModeOn Indicates whether the unprivileged mode option is on
+ * @param[in] logFile The log file option for CRIU
+ * @return 0 if the execution of the 'criu restore' command succeeds, -1 otherwise
+ */
+static int
+restoreFromCheckpoint(const char *checkpointDirectory, int logLevel, jboolean unprivilegedModeOn, const char *logFile)
+{
+    int restoreStatus = 0;
```
keithc-ca, line 182: Suggest removing this: the only value that would ever be read from it will be -1, which can be used in the return statement.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+        argv[argc++] = logFileOption;
+    }
+    argv[argc] = NULL;
+    execvp(argv[0], (char * const *)argv);
+    /* If execvp returns, there was an error. */
+    restoreStatus = -1;
+freeLogFileOption:
+    if (NULL != logFileOption) {
+        JLI_MemFree((void *)logFileOption);
+        logFileOption = NULL;
```
keithc-ca, line 230: This assignment is not useful (there are no following reads).

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+    int logLevel = 0;
+    int childProcessExitStatus = EXIT_SUCCESS;
+    jboolean unprivilegedModeOn = JNI_FALSE;
+    const char *logFile = NULL;
+    int childProcessPidStatus = 0;
+    int childProcessPidExitStatus = 0;
+    checkpointDirectory = getCheckpointDirectory(argc, argv, &error);
+    /* We ignore restore if the restore option -XX:CRaCRestoreFrom is not specified. */
+    if (-1 == error) {
+        return;
```
keithc-ca, line 257: Suggest the comment should be placed where it applies:
```c
    if (-1 == error) {
        /* Option -XX:CRaCRestoreFrom not specified. */
        return;
```

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,309 @@
...
+    }
+    if (-2 == error) {
+        JLI_ReportErrorMessage("Failed to get the CRIU checkpoint directory, error=%d.", error);
```
keithc-ca, line 260: There's no point in including `error=%d` in messages like this where the value can only be -2.

keithc-ca (changes requested):

closed/src/java.base/share/native/libjli/criuhelpers.h
JasonFengJ9: Removed this header file.

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: The initial value is set to `0`.

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Updated

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Change it to `fail`.

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Removed it.

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Removed it.

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Moved the comment.

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Updated

closed/src/java.base/share/native/libjli/criuhelpers.h
keithc-ca: Why? I suggest keeping the header file is preferable to declaring the function in each source file where it is used (the declaration is currently missing from `src/java.base/share/native/libjli/java.c`).

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,308 @@
...
+
+#include "j9cfg.h"
+
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+
+#include "java.h"
+#include <ctype.h>
+#include <sys/wait.h>
+
+void handleCRaCRestore(int argc, char **argv);
```
keithc-ca, line 33: This should include `criuhelpers.h` instead of declaring it here.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,308 @@
...
+ * ===========================================================================
+ */
+
+#include "j9cfg.h"
+
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+
+#include "java.h"
+#include <ctype.h>
+#include <sys/wait.h>
```
keithc-ca, line 31: I think this needs to include `jli_util.h`, `jni.h` and `string.h`.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,308 @@
...
+    const char *argv[9] = { NULL };
+    argv[argc++] = "criu";
+    argv[argc++] = "restore";
+    argv[argc++] = "-D";
+    argv[argc++] = checkpointDirectory;
+    length = snprintf(NULL, 0, "-v%d", logLevel);
+    if (length < 0) {
+        JLI_ReportErrorMessage("Failed to calculate the length of option '-v%d'.", logLevel);
+        goto fail;
+    }
```
keithc-ca, line 205: Please remove this; by ensuring `logLevelOption` is large enough we can be sure it could never fail.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,308 @@
...
+        }
+        argv[argc++] = logFileOption;
+    }
+    argv[argc] = NULL;
+    execvp(argv[0], (char * const *)argv);
+    /* If execvp returns, there was an error. */
+    if (NULL != logFileOption) {
+        JLI_MemFree((void *)logFileOption);
+    }
+fail:
```
keithc-ca, line 231: This label must be before line 228 (see line 221).

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,308 @@
...
+    const char *logFile = NULL;
+    int childProcessPidStatus = 0;
+    int childProcessPidExitStatus = 0;
+    checkpointDirectory = getCheckpointDirectory(argc, argv, &error);
+    if (-1 == error) {
+        /* Option -XX:CRaCRestoreFrom not specified. */
+        return;
+    }
+    if (-2 == error) {
+        JLI_ReportErrorMessage("Failed to get the CRIU checkpoint directory, error=-2.");
```
keithc-ca, line 259: Please remove `, error=-2` from messages like this. See also lines 271, 277, and 283.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,308 @@
...
+        } else {
+            goto setLogLevelOptionValueNotValidError;
+        }
+    } else if (-1 == *error) {
+        goto done;
+    }
+setLogLevelOptionValueNotValidError:
+    JLI_ReportErrorMessage(
+        "The option '-Dopenj9.internal.criu.logLevel=%s' is not valid.",
+        logLevelPropertyValue);
```
keithc-ca, line 127: Please indent these lines four more spaces (compare with lines 150-152).

src/java.base/share/native/libjli/java.c
```diff
@@ -566,6 +567,10 @@ JavaMain(void* _args)
     jfieldID noArgMainField;
     jboolean noArgMain;
 
+#if defined(J9VM_OPT_CRAC_SUPPORT)
+    handleCRaCRestore(argc, argv);
+#endif /* defined(J9VM_OPT_CRAC_SUPPORT) */
+
```
keithc-ca, line 573: Previously, this call occurred after the call to `RegisterThread()`: Why is it now before?

keithc-ca (changes requested):

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Added `criuhelpers.h` with `handleCRaCRestore()` declaration.

closed/src/java.base/share/native/libjli/criuhelpers.h
JasonFengJ9: Added it.

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: `java.h` covers `jli_util.h`, `jni.h`, `string.h` and others.

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Removed

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Moved `fail` label up.

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Removed

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Added spaces.

src/java.base/share/native/libjli/java.c
JasonFengJ9: `handleCRaCRestore()` doesn't require `RegisterThread()`, and we would like to invoke `criu restore` as early as possible.
On a side note, I attempted to move this function into OpenJ9 respo, hoping to avoid backporting it to other extension repos. In some quick experiments on a Fyre image, it was ~30ms slower which doesn't fit the instanton perf goal.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,303 @@
...
+            JLI_ReportErrorMessage("The value of the command line option -XX:CRaCRestoreFrom was not found.");
+            *error = -2;
+        }
+    }
+    return checkpointDirectory;
+}
+
+/**
+ * Get the log level specified in the command line arguments.
+ * The valid log level is [0, 4], 2 by default.
```
keithc-ca, line 93: Suggest this comment be rewritten:
* Valid log levels are from 0 to 4, inclusive; the default is 2.

closed/src/java.base/share/native/libjli/criuhelpers.c
```diff
@@ -0,0 +1,303 @@
...
+    int length = -1;
+    char logLevelOption[4] = { 0 };
+    char *logFileOption = NULL;
+    int argc = 0;
+    const char *argv[9] = { NULL };
+    argv[argc++] = "criu";
+    argv[argc++] = "restore";
+    argv[argc++] = "-D";
+    argv[argc++] = checkpointDirectory;
+    /* The logLevel is [0, 4], or 2 by default. */
```
keithc-ca, line 200: The default is not relevant here, only the limits.
```c
    /* The log level is a single digit. */
```

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Updated

closed/src/java.base/share/native/libjli/criuhelpers.c
JasonFengJ9: Updated

keithc-ca (commented):
Please squash. The intermediate state doesn't represent something I think we need to keep.

JasonFengJ9:
> Please squash. 

Done

keithc-ca:
Jenkins test sanity amac,xlinux,zlinux jdknext

keithc-ca:
Opened https://github.com/eclipse-openj9/openj9/issues/20193 for the test failure on s390x_linux which seems unrelated to this as CRaC is not enabled on that platform.

keithc-ca (approved):

PR #784, opened 2024-04-26
Update get_source.sh

keithc-ca:
* improve feedback for invalid arguments
* improve feedback for cloning steps
* remove support for non-standard options (unused since [openj9#19387](https://github.com/eclipse-openj9/openj9#19387))
* remove unused variables

### DIFF
--- a/get_source.sh
+++ b/get_source.sh
@@ -22,9 +22,6 @@
 # exit immediately if any unexpected error occurs
 set -e
 
-openssloptions=""
-DOWNLOAD_OPENSSL=false
-
 # These maps are keyed by the prefix of option names (e.g. openj9, omr).
 declare -A source_branch    # branch or tag
 declare -A source_folder    # local working directory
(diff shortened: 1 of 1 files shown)


pshipton:
We should open a doc issue to have a release note that `--openssl-repo` and `--openssl-version` are removed, and what they are replaced with.

pshipton:
Also it would be worth announcing this on Slack, since anyone doing their own builds may be surprised by the change.

keithc-ca:
There's nothing in openj9-docs that currently mentions those (old) options; and https://github.com/eclipse-openj9/openj9/pull/19387 updated the documentation in openj9. Do you not think that's sufficient?

keithc-ca:
Slack announcement: https://openj9.slack.com/archives/CDS7QE9HB/p1714150601939569.

pshipton:
Usually when there is a change there is a release note to give notice of the change. I can amend my statement: We should open a doc issue to have a release note that --openssl-repo and --openssl-version are removed, **and refer to the updated build instructions.**

pshipton:
jenkins compile amac jdknext

keithc-ca:
> Usually when there is a change there is a release note to give notice of the change

Yes, but so far we've only been documenting changes that impact users of the SDK. This isn't such a change.

pshipton (approved):

pshipton:
I'm fine if it just goes into the OpenJ9 release notes document and not the User Guide. The doc issue can mention that.
https://github.com/eclipse-openj9/openj9/tree/master/doc/release-notes

pshipton:
I got what I needed from the build, which created the JVM, but failed to upload afterwards.

keithc-ca:
I created https://github.com/eclipse-openj9/openj9/pull/19392, but even there I couldn't find an earlier example of pointing out changes to instructions for building a JDK.

PR #789, opened 2024-05-14
Always set COMPATIBLE_CDS_ALIGNMENT_DEFAULT to false for OpenJ9

pshipton:
Otherwise there is an error.
`configure: error: Option --enable-compatible-cds-alignment is not available`

Fixes https://github.com/eclipse-openj9/openj9/issues/19488

Merging to staging.

### DIFF
--- a/make/autoconf/jdk-options.m4
+++ b/make/autoconf/jdk-options.m4
@@ -24,7 +24,7 @@
 #
 
 # ===========================================================================
-# (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
+# (c) Copyright IBM Corp. 2022, 2024 All Rights Reserved
 # ===========================================================================
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


make/autoconf/jdk-options.m4
```diff
@@ -684,7 +684,7 @@ AC_DEFUN([JDKOPT_ENABLE_DISABLE_CDS_ARCHIVE],
 #
 AC_DEFUN([JDKOPT_ENABLE_DISABLE_COMPATIBLE_CDS_ALIGNMENT],
 [
-  UTIL_ARG_ENABLE(NAME: compatible-cds-alignment, DEFAULT: $COMPATIBLE_CDS_ALIGNMENT_DEFAULT,
+  UTIL_ARG_ENABLE(NAME: compatible-cds-alignment, DEFAULT: false,
```
keithc-ca, line 687: I suggest a better solution would be to ensure that `COMPATIBLE_CDS_ALIGNMENT_DEFAULT=false` so `AC_SUBST(COMPATIBLE_CDS_ALIGNMENT_DEFAULT)` is consistent should `COMPATIBLE_CDS_ALIGNMENT_DEFAULT` be used elsewhere. See lines 202-206.

keithc-ca (changes requested):

make/autoconf/jdk-options.m4
pshipton: Updated.

keithc-ca (approved):

PR #791, opened 2024-05-16
Throw OOM if array size exceeds int maximum

theresa-m:
Port from https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/753


keithc-ca (approved):
Tested via https://github.com/ibmruntimes/openj9-openjdk-jdk22/pull/59.

PR #793, opened 2024-05-27
Allow extension of RestrictedSecurity profiles

KostasTsiounis:
`RestrictedSecurity` profiles sometimes share a lot of duplicate settings with only minor differences. With these changes the extension, similar to object-orientation, of profiles becomes possible.

More specifically, a profile `A` can extend another a profile `B`, using `RestrictedSecurity.<profile A name>.extends = RestrictedSecurity.<profile B name>`. This allows profile `A` to inherit all of profile `B`'s properties. One can add additional properties to profile `A`, or amend some of the existing ones. That includes overriding, appending or removing from a property (wherever that's applicable).

An additional property is introduced. The `RestrictedSecurity.<profile name>.desc.hash = <hash algorithm>:<hash>` is used to ensure the profile hasn't been unintentionally altered. The profile's properties are hashed using the selected `<hash algorithm>`, and the result is compared to the `<hash>` provided through the property. This property is mandatory for base profiles (i.e., profiles that are not extending anything), and optional for the rest.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -23,7 +23,10 @@
  */
 package openj9.internal.security;
 
+import java.nio.charset.StandardCharsets;
 import java.security.AccessController;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.security.PrivilegedAction;
 import java.security.Provider.Service;
 import java.time.LocalDate;
@@ -42,6 +45,8 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import sun.security.util.Debug;
 
@@ -290,7 +295,7 @@ public final class RestrictedSecurity {
     private static void getProfileID(Properties props) {
         String potentialProfileID = "RestrictedSecurity." + selectedProfile;
 
-        if (selectedProfile.indexOf(".") != -1) {
+        if (selectedProfile.indexOf('.') != -1) {
             /* The default profile is used, or the user specified the
              * full <profile.version>.
              */
@@ -358,8 +363,8 @@ public final class RestrictedSecurity {
         }
     }
 
-    private static void checkFIPSCompatibility(Properties props) {
-        boolean isFIPSProfile = Boolean.parseBoolean(props.getProperty(profileID + ".desc.fips"));
+    private static void checkFIPSCompatibility() {
+        boolean isFIPSProfile = restricts.descIsFIPS;
         if (isFIPSProfile) {
             if (debug != null) {
                 debug.println("RestrictedSecurity profile " + profileID
@@ -441,13 +446,9 @@ public final class RestrictedSecurity {
                 getProfileID(props);
                 checkIfKnownProfileSupported();
 
-                // If user enabled FIPS, check whether chosen profile is applicable.
-                if (userEnabledFIPS) {
-                    checkFIPSCompatibility(props);
-                }
-
                 // Initialize restricted security properties from java.security file.
-                restricts = new RestrictedSecurityProperties(profileID, props);
+                ProfileParser profileParser = new ProfileParser(profileID, props);
+                restricts = profileParser.getProperties();
 
                 // Restricted security properties checks.
                 restrictsCheck();
@@ -476,6 +477,9 @@ public final class RestrictedSecurity {
                 }
 
                 securityEnabled = true;
+
+                // Check whether the hash values match (i.e., profiles haven't been altered).
+                profileParser.checkHashValues();
             }
         } catch (Exception e) {
             if (debug != null) {
@@ -536,22 +540,16 @@ public final class RestrictedSecurity {
             String jdkPropsName = entry.getKey();
             String propsNewValue = entry.getValue();
 
-            String propsOldValue = props.getProperty(jdkPropsName);
-            if (isNullOrBlank(propsOldValue)) {
-                propsOldValue = "";
-            }
-
             if ((propsNewValue != null) && userEnabledFIPS && !allowSetProperties) {
                 // Add to set of properties set by the active profile.
                 unmodifiableProperties.add(jdkPropsName);
             }
 
             if (!isNullOrBlank(propsNewValue)) {
-                String values = isNullOrBlank(propsOldValue) ? propsNewValue : (propsOldValue + ", " + propsNewValue);
-                props.setProperty(jdkPropsName, values);
+                props.setProperty(jdkPropsName, propsNewValue);
                 if (debug != null) {
-                    debug.println("Added restricted security properties, with property: " + jdkPropsName + " value: "
-                            + values);
+                    debug.println("Added restricted security properties, with property: "
+                            + jdkPropsName + " value: " + propsNewValue);
                 }
             }
         }
@@ -587,6 +585,11 @@ public final class RestrictedSecurity {
                 || isNullOrBlank(restricts.jdkSecureRandomAlgorithm)) {
             printStackTraceAndExit("Restricted security mode secure random is missing.");
         }
+
+        // If user enabled FIPS, check whether chosen profile is applicable.
+        if (userEnabledFIPS) {
+            checkFIPSCompatibility();
+        }
     }
 
     /**
@@ -597,12 +600,15 @@ public final class RestrictedSecurity {
      */
     private static boolean isPolicySunset(String descSunsetDate) {
         boolean isSunset = false;
-        try {
-            isSunset = LocalDate.parse(descSunsetDate, DateTimeFormatter.ofPattern("yyyy-MM-dd"))
-                    .isBefore(LocalDate.now());
-        } catch (DateTimeParseException except) {
-            printStackTraceAndExit(
-                    "Restricted security policy sunset date is incorrect, the correct format is yyyy-MM-dd.");
+        // Only check if a sunset date is specified in the profile.
+        if (!isNullOrBlank(descSunsetDate)) {
+            try {
+                isSunset = LocalDate.parse(descSunsetDate, DateTimeFormatter.ofPattern("yyyy-MM-dd"))
+                        .isBefore(LocalDate.now());
+            } catch (DateTimeParseException except) {
+                printStackTraceAndExit(
+                        "Restricted security policy sunset date is incorrect, the correct format is yyyy-MM-dd.");
+            }
         }
 
         if (debug != null) {
@@ -631,33 +637,65 @@ public final class RestrictedSecurity {
     }
 
     /**
+     * Check if the input string is asterisk (*).
+     *
+     * @param string input string for checking
+     * @return true if the input string is asterisk
+     */
+    private static boolean isAsterisk(String string) {
+        return "*".equals(string);
+    }
+
+    /**
+     * Get the provider name defined in provider construction method.
+     *
+     * @param providerName provider name or provider with packages
+     * @return provider name defined in provider construction method
+     */
+    private static String getProvidersSimpleName(String providerName) {
+        if (providerName.equals("com.sun.security.sasl.Provider")) {
+            // The main class for the SunSASL provider is com.sun.security.sasl.Provider.
+            return "SunSASL";
+        } else {
+            // Remove the provider's class package names if present.
+            int pos = providerName.lastIndexOf('.');
+            if (pos >= 0) {
+                providerName = providerName.substring(pos + 1);
+            }
+            // Provider without package names.
+            return providerName;
+        }
+    }
+
+    /**
      * This class is used to save and operate on restricted security
      * properties which are loaded from the java.security file.
      */
     private static final class RestrictedSecurityProperties {
+        private final String profileID;
 
-        private String descName;
-        private boolean descIsDefault;
-        private boolean descIsFIPS;
-        private String descNumber;
-        private String descPolicy;
-        private String descSunsetDate;
+        private final String descName;
+        private final boolean descIsDefault;
+        private final boolean descIsFIPS;
+        private final String descNumber;
+        private final String descPolicy;
+        private final String descSunsetDate;
 
         // Security properties.
-        private String jdkTlsDisabledNamedCurves;
-        private String jdkTlsDisabledAlgorithms;
-        private String jdkTlsEphemeralDHKeySize;
-        private String jdkTlsLegacyAlgorithms;
-        private String jdkCertpathDisabledAlgorithms;
-        private String jdkSecurityLegacyAlgorithms;
-        private String keyStoreType;
-        private String keyStore;
+        private final String jdkTlsDisabledNamedCurves;
+        private final String jdkTlsDisabledAlgorithms;
+        private final String jdkTlsEphemeralDHKeySize;
+        private final String jdkTlsLegacyAlgorithms;
+        private final String jdkCertpathDisabledAlgorithms;
+        private final String jdkSecurityLegacyAlgorithms;
+        private final String keyStoreType;
+        private final String keyStore;
 
         // For SecureRandom.
-        String jdkSecureRandomProvider;
-        String jdkSecureRandomAlgorithm;
+        final String jdkSecureRandomProvider;
+        final String jdkSecureRandomAlgorithm;
 
-        String jdkFipsMode;
+        final String jdkFipsMode;
 
         // Provider with argument (provider name + optional argument).
         private final List<String> providers;
@@ -666,262 +704,46 @@ public final class RestrictedSecurity {
         // The map is keyed by provider name.
         private final Map<String, Constraint[]> providerConstraints;
 
-        private final String profileID;
+        private RestrictedSecurityProperties(String profileID, ProfileParser parser) {
+            this.profileID = profileID;
 
-        // The java.security properties.
-        private final Properties securityProps;
+            this.descName = parser.getProperty("descName");
+            this.descIsDefault = parser.descIsDefault;
+            this.descIsFIPS = parser.descIsFIPS;
+            this.descNumber = parser.getProperty("descNumber");
 
-        /**
-         *
-         * @param id    the restricted security custom profile ID
-         * @param props the java.security properties
-         * @param trace the user security trace
-         * @param audit the user security audit
-         * @param help  the user security help
-         */
-        private RestrictedSecurityProperties(String id, Properties props) {
-            Objects.requireNonNull(props);
+            this.descPolicy = parser.getProperty("descPolicy");
+            this.descSunsetDate = parser.getProperty("descSunsetDate");
 
-            profileID = id;
-            securityProps = props;
+            // Security properties.
+            this.jdkTlsDisabledNamedCurves = parser.getProperty("jdkTlsDisabledNamedCurves");
+            this.jdkTlsDisabledAlgorithms = parser.getProperty("jdkTlsDisabledAlgorithms");
+            this.jdkTlsEphemeralDHKeySize = parser.getProperty("jdkTlsEphemeralDHKeySize");
+            this.jdkTlsLegacyAlgorithms = parser.getProperty("jdkTlsLegacyAlgorithms");
+            this.jdkCertpathDisabledAlgorithms = parser.getProperty("jdkCertpathDisabledAlgorithms");
+            this.jdkSecurityLegacyAlgorithms = parser.getProperty("jdkSecurityLegacyAlgorithms");
+            this.keyStoreType = parser.getProperty("keyStoreType");
+            this.keyStore = parser.getProperty("keyStore");
 
-            providers = new ArrayList<>();
-            providersSimpleName = new ArrayList<>();
-            providerConstraints = new HashMap<>();
+            // For SecureRandom.
+            this.jdkSecureRandomProvider = parser.getProperty("jdkSecureRandomProvider");
+            this.jdkSecureRandomAlgorithm = parser.getProperty("jdkSecureRandomAlgorithm");
 
-            // Initialize the properties.
-            init();
-        }
+            this.jdkFipsMode = parser.getProperty("jdkFipsMode");
 
-        /**
-         * Initialize restricted security properties.
-         */
-        private void init() {
-            if (debug != null) {
-                debug.println("Initializing restricted security mode.");
-            }
-
-            try {
-                // Load restricted security providers from java.security properties.
-                initProviders();
-                // Load restricted security properties from java.security properties.
-                initProperties();
-                // Load restricted security provider constraints from java.security properties.
-                initConstraints();
-            } catch (Exception e) {
-                if (debug != null) {
-                    debug.println("Unable to initialize restricted security mode.");
-                }
-                printStackTraceAndExit(e);
-            }
+            this.providers = new ArrayList<>(parser.providers);
+            this.providersSimpleName = new ArrayList<>(parser.providersSimpleName);
+            this.providerConstraints = parser.providerConstraints
+                                             .entrySet()
+                                             .stream()
+                                             .collect(Collectors.toMap(
+                                                     e -> e.getKey(),
+                                                     e -> e.getValue().toArray(new Constraint[0])
+                                             ));
 
             if (debug != null) {
-                debug.println("Initialization of restricted security mode completed.");
-
-                // Print all available restricted security profiles.
-                listAvailableProfiles();
-
                 // Print information of utilized security profile.
                 listUsedProfile();
-            }
-        }
-
-        /**
-         * Load restricted security provider.
-         */
-        private void initProviders() {
-            if (debug != null) {
-                debug.println("\tLoading providers of restricted security profile.");
-            }
-
-            for (int pNum = 1;; ++pNum) {
-                String providerInfo = securityProps
-                        .getProperty(profileID + ".jce.provider." + pNum);
-
-                if ((providerInfo == null) || providerInfo.trim().isEmpty()) {
-                    break;
-                }
-
-                if (!areBracketsBalanced(providerInfo)) {
-                    printStackTraceAndExit("Provider format is incorrect: " + providerInfo);
-                }
-
-                int pos = providerInfo.indexOf('[');
-                String providerName = (pos < 0) ? providerInfo.trim() : providerInfo.substring(0, pos).trim();
-                // Provider with argument (provider name + optional argument).
-                providers.add(pNum - 1, providerName);
-
-                // Remove the provider's optional arguments if present.
-                pos = providerName.indexOf(' ');
-                if (pos >= 0) {
-                    providerName = providerName.substring(0, pos);
-                }
-                providerName = providerName.trim();
-
-                // Remove argument, e.g. -NSS-FIPS, if present.
-                pos = providerName.indexOf('-');
-                if (pos >= 0) {
-                    providerName = providerName.substring(0, pos);
-                }
-
-                // Provider name defined in provider construction method.
-                providerName = getProvidersSimpleName(providerName);
-                providersSimpleName.add(pNum - 1, providerName);
-            }
-
-            if (providers.isEmpty()) {
-                printStackTraceAndExit(
-                        "No providers are specified as part of the Restricted Security profile.");
-            }
-
-            if (debug != null) {
-                debug.println("\tProviders of restricted security profile successfully loaded.");
-            }
-        }
-
-        /**
-         * Load restricted security properties.
-         */
-        private void initProperties() {
-            if (debug != null) {
-                debug.println("\tLoading properties of restricted security profile.");
-            }
-
-            descName = parseProperty(securityProps.getProperty(profileID + ".desc.name"));
-            descIsDefault = Boolean.parseBoolean(parseProperty(securityProps.getProperty(profileID + ".desc.default")));
-            descIsFIPS = Boolean.parseBoolean(parseProperty(securityProps.getProperty(profileID + ".desc.fips")));
-            descNumber = parseProperty(securityProps.getProperty(profileID + ".desc.number"));
-            descPolicy = parseProperty(securityProps.getProperty(profileID + ".desc.policy"));
-            descSunsetDate = parseProperty(securityProps.getProperty(profileID + ".desc.sunsetDate"));
-
-            jdkTlsDisabledNamedCurves = parseProperty(
-                    securityProps.getProperty(profileID + ".tls.disabledNamedCurves"));
-            jdkTlsDisabledAlgorithms = parseProperty(
-                    securityProps.getProperty(profileID + ".tls.disabledAlgorithms"));
-            jdkTlsEphemeralDHKeySize = parseProperty(
-                    securityProps.getProperty(profileID + ".tls.ephemeralDHKeySize"));
-            jdkTlsLegacyAlgorithms = parseProperty(
-                    securityProps.getProperty(profileID + ".tls.legacyAlgorithms"));
-            jdkCertpathDisabledAlgorithms = parseProperty(
-                    securityProps.getProperty(profileID + ".jce.certpath.disabledAlgorithms"));
-            jdkSecurityLegacyAlgorithms = parseProperty(
-                    securityProps.getProperty(profileID + ".jce.legacyAlgorithms"));
-            keyStoreType = parseProperty(
-                    securityProps.getProperty(profileID + ".keystore.type"));
-            keyStore = parseProperty(
-                    securityProps.getProperty(profileID + ".javax.net.ssl.keyStore"));
-
-            jdkSecureRandomProvider = parseProperty(
-                    securityProps.getProperty(profileID + ".securerandom.provider"));
-            jdkSecureRandomAlgorithm = parseProperty(
-                    securityProps.getProperty(profileID + ".securerandom.algorithm"));
-            jdkFipsMode = parseProperty(
-                    securityProps.getProperty(profileID + ".fips.mode"));
-
-            if (debug != null) {
-                debug.println("\tProperties of restricted security profile successfully loaded.");
-            }
-        }
-
-        /**
-         * Load security constraints with type, algorithm, attributes.
-         *
-         * Example:
-         * RestrictedSecurity1.jce.provider.1 = SUN [{CertPathBuilder, PKIX, *},
-         * {Policy, JavaPolicy, *}, {CertPathValidator, *, *}].
-         */
-        private void initConstraints() {
-            if (debug != null) {
-                debug.println("\tLoading constraints of restricted security profile.");
-            }
-
-            for (int pNum = 1; pNum <= providersSimpleName.size(); pNum++) {
-                String providerName = providersSimpleName.get(pNum - 1);
-                String providerInfo = securityProps
-                        .getProperty(profileID + ".jce.provider." + pNum);
-
-                if (debug != null) {
-                    debug.println("\t\tLoading constraints for security provider: " + providerName);
-                }
-
-                // Check if the provider has constraints.
-                if (providerInfo.indexOf('[') < 0) {
-                    if (debug != null) {
-                        debug.println("\t\t\tNo constraints for security provider: " + providerName);
-                    }
-                    providerConstraints.put(providerName, new Constraint[0]);
-                    continue;
-                }
-
-                // Remove the whitespaces in the format separator if present.
-                providerInfo = providerInfo.trim()
-                        .replaceAll("\\[\\s+\\{", "[{")
-                        .replaceAll("\\}\\s+\\]", "}]")
-                        .replaceAll("\\}\\s*,\\s*\\{", "},{");
-
-                int startIndex = providerInfo.lastIndexOf("[{");
-                int endIndex = providerInfo.indexOf("}]");
-
-                // Provider with constraints.
-                if ((startIndex > 0) && (endIndex > startIndex)) {
-                    String[] constrArray = providerInfo
-                            .substring(startIndex + 2, endIndex).split("\\},\\{");
-
-                    if (constrArray.length <= 0) {
-                        printStackTraceAndExit("Constraint format is incorrect: " + providerInfo);
-                    }
-
-                    // Constraint object array.
-                    // For each constraint type, algorithm and attributes.
-                    Constraint[] constraints = new Constraint[constrArray.length];
-
-                    int cNum = 0;
-                    for (String constr : constrArray) {
-                        String[] input = constr.split(",");
-
-                        // Each constraint must includes 3 fields(type, algorithm, attributes).
-                        if (input.length != 3) {
-                            printStackTraceAndExit("Constraint format is incorrect: " + providerInfo);
-                        }
-
-                        String inType = input[0].trim();
-                        String inAlgorithm = input[1].trim();
-                        String inAttributes = input[2].trim();
-
-                        // Each attribute must includes 2 fields (key and value) or *.
-                        if (!isAsterisk(inAttributes)) {
-                            String[] attributeArray = inAttributes.split(":");
-                            for (String attribute : attributeArray) {
-                                String[] in = attribute.split("=", 2);
-                                if (in.length != 2) {
-                                    printStackTraceAndExit(
-                                            "Constraint attributes format is incorrect: " + providerInfo);
-                                }
-                            }
-                        }
-
-                        Constraint constraint = new Constraint(inType, inAlgorithm, inAttributes);
-
-                        if (debug != null) {
-                            debug.println("\t\t\tConstraint specified for security provider: " + providerName);
-                            debug.println("\t\t\t\twith type: " + inType);
-                            debug.println("\t\t\t\tfor algorithm: " + inAlgorithm);
-                            debug.println("\t\t\t\twith attributes: " + inAttributes);
-                        }
-                        constraints[cNum] = constraint;
-                        cNum++;
-                    }
-                    providerConstraints.put(providerName, constraints);
-                    if (debug != null) {
-                        debug.println("\t\tSuccessfully loaded constraints for security provider: " + providerName);
-                    }
-                } else {
-                    printStackTraceAndExit("Constraint format is incorrect: " + providerInfo);
-                }
-            }
-
-            if (debug != null) {
-                debug.println("\tAll constraints of restricted security profile successfully loaded.");
             }
         }
 
@@ -933,6 +755,10 @@ public final class RestrictedSecurity {
          */
         boolean isRestrictedServiceAllowed(Service service) {
             String providerName = service.getProvider().getName();
+
+            if (debug != null) {
+                debug.println("Checking service " + service.toString() + " offered by provider " + providerName + ".");
+            }
 
             // Provider with argument, remove argument.
             // e.g. SunPKCS11-NSS-FIPS, remove argument -NSS-FIPS.
@@ -950,6 +776,9 @@ public final class RestrictedSecurity {
                 return false;
             } else if (constraints.length == 0) {
                 // Allow this provider with no constraints.
+                if (debug != null) {
+                    debug.println("No constraints for provider " + providerName + ".");
+                }
                 return true;
             }
 
@@ -957,27 +786,42 @@ public final class RestrictedSecurity {
             String type = service.getType();
             String algorithm = service.getAlgorithm();
 
+            if (debug != null) {
+                debug.println("Security constraints check of provider.");
+            }
             for (Constraint constraint : constraints) {
                 String cType = constraint.type;
                 String cAlgorithm = constraint.algorithm;
                 String cAttribute = constraint.attributes;
+                if (debug != null) {
+                    debug.println("Checking provider constraint:"
+                                + "\n\tService type: " + cType
+                                + "\n\tAlgorithm: " + cAlgorithm
+                                + "\n\tAttributes: " + cAttribute);
+                }
 
                 if (!isAsterisk(cType) && !type.equals(cType)) {
                     // The constraint doesn't apply to the service type.
+                    if (debug != null) {
+                        debug.println("The constraint doesn't apply to the service type.");
+                    }
                     continue;
                 }
                 if (!isAsterisk(cAlgorithm) && !algorithm.equalsIgnoreCase(cAlgorithm)) {
                     // The constraint doesn't apply to the service algorithm.
+                    if (debug != null) {
+                        debug.println("The constraint doesn't apply to the service algorithm.");
+                    }
                     continue;
                 }
 
                 // For type and algorithm match, and attribute is *.
                 if (isAsterisk(cAttribute)) {
                     if (debug != null) {
-                        debug.println("Security constraints check."
-                                + " Service type: " + type
-                                + " Algorithm: " + algorithm
-                                + " is allowed in provider: " + providerName);
+                        debug.println("The following service:"
+                                + "\n\tService type: " + type
+                                + "\n\tAlgorithm: " + algorithm
+                                + "\nis allowed in provider: " + providerName);
                     }
                     return true;
                 }
@@ -993,36 +837,47 @@ public final class RestrictedSecurity {
                     String cName = input[0].trim();
                     String cValue = input[1].trim();
                     String sValue = service.getAttribute(cName);
+                    if (debug != null) {
+                        debug.println("Checking specific attribute with:"
+                                + "\n\tName: " + cName
+                                + "\n\tValue: " + cValue
+                                + "\nagainst the service attribute value: " + sValue);
+                    }
                     if ((sValue == null) || !cValue.equalsIgnoreCase(sValue)) {
                         // If any attribute doesn't match, return service is not allowed.
                         if (debug != null) {
-                            debug.println(
-                                    "Security constraints check."
-                                            + " Service type: " + type
-                                            + " Algorithm: " + algorithm
-                                            + " Attribute: " + cAttribute
-                                            + " is NOT allowed in provider: " + providerName);
+                            debug.println("Attributes don't match!");
+                            debug.println("The following service:"
+                                        + "\n\tService type: " + type
+                                        + "\n\tAlgorithm: " + algorithm
+                                        + "\n\tAttribute: " + cAttribute
+                                        + "\nis NOT allowed in provider: " + providerName);
                         }
                         return false;
                     }
+                    if (debug != null) {
+                        debug.println("Attributes match!");
+                    }
                 }
                 if (debug != null) {
-                    debug.println(
-                            "Security constraints check."
-                                    + " Service type: " + type
-                                    + " Algorithm: " + algorithm
-                                    + " Attribute: " + cAttribute
-                                    + " is allowed in provider: " + providerName);
+                    debug.println("All attributes matched!");
+                    debug.println("The following service:"
+                                + "\n\tService type: " + type
+                                + "\n\tAlgorithm: " + algorithm
+                                + "\n\tAttribute: " + cAttribute
+                                + "\nis allowed in provider: " + providerName);
                 }
                 return true;
             }
-            if (debug != null) {
-                debug.println("Security constraints check."
-                        + " Service type: " + type
-                        + " Algorithm: " + algorithm
-                        + " is NOT allowed in provider: " + providerName);
-            }
+
             // No match for any constraint, return NOT allowed.
+            if (debug != null) {
+                debug.println("Could not find a constraint to match.");
+                debug.println("The following service:"
+                            + "\n\tService type: " + type
+                            + "\n\tAlgorithm: " + algorithm
+                            + "\nis NOT allowed in provider: " + providerName);
+            }
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


keithc-ca:
It would be helpful to have a fuller description here of how users will interact with this change, including what the restrictions are and perhaps some examples.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+            System.out.println();
+
+            // List providers.
+            System.out.println(profileID + " Profile Providers:");
+            System.out.println("===============================================");
+            for (int providerPosition = 0; providerPosition < providers.size(); providerPosition++) {
+                printProperty(profileID + ".jce.provider." + (providerPosition + 1) + ": ",
+                        providers.get(providerPosition));
+                String providerSimpleName = providersSimpleName.get(providerPosition);
+                for (Constraint providerConstraint: providerConstraints.get(providerSimpleName)) {
```
keithc-ca, line 916: Space both sides of `:`, please.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+        private Map<String, List<String>> profilesInfo;
+
+        // The java.security properties.
+        private final Properties securityProps;
+
+        private final List<String> appendableProps = Arrays.asList("jdkTlsDisabledNamedCurves",
+                                                                   "jdkTlsDisabledAlgorithms",
+                                                                   "jdkTlsLegacyAlgorithms",
+                                                                   "jdkCertpathDisabledAlgorithms",
+                                                                   "jdkSecurityLegacyAlgorithms");
```
keithc-ca, line 975: Suggest removing this, and updating `isPropertyAppendable()` to:
```java
        private static boolean isPropertyAppendable(String property) {
            switch (property) {
            case "jdkCertpathDisabledAlgorithms":
            case "jdkSecurityLegacyAlgorithms":
            case "jdkTlsDisabledAlgorithms":
            case "jdkTlsDisabledNamedCurves":
            case "jdkTlsLegacyAlgorithms":
                return true;
            default:
                return false;
            }
        }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+            // Initialize the properties.
+            init(profileID);
+        }
+
+        private RestrictedSecurityProperties getProperties() {
+            RestrictedSecurityProperties rsProperties = new RestrictedSecurityProperties();
+
+            rsProperties.profileID = profileID;
+
+            rsProperties.descName = getProperty("descName");
```
keithc-ca, line 1006: This should be done within the constructor for `RestrictedSecurityProperties` rather than from the outside (it will also allow fields like `profileID` to be `final`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+
+            String potentialExtendsProfileID = parseProperty(securityProps.getProperty(profileID + ".extends"));
+            if (potentialExtendsProfileID != null) { // If profile extends another profile.
+                if (debug != null) {
+                    debug.println("\t'" + profileID + "' extends '" + potentialExtendsProfileID + "'.");
+                }
+
+                // Check if extended profile exists.
+                String extendsProfileID = null;
+                if (potentialExtendsProfileID.indexOf(".") != potentialExtendsProfileID.lastIndexOf(".")) {
```
keithc-ca, line 1062: Prefer `indexof(char)` over `indexOf(String)` when the argument is one character.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                        printStackTraceAndExit(potentialExtendsProfileID + " that is supposed to extend '"
+                                + profileID + "' is not present in the java.security file or any appended files.");
+                    }
+                } else {
+                    printStackTraceAndExit(potentialExtendsProfileID + " that is supposed to extend '"
+                            + profileID + "' is not a full profile name.");
+                }
+
+                // Recursively call init() on extended profile.
+                init(extendsProfileID);
```
keithc-ca, line 1089: This needs to avoid infinite recursion.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                List<String> allInfo = new ArrayList<>();
+                // Load restricted security providers from java.security properties.
+                updateProviders(profileExtensionId, allInfo);
+                // Load restricted security properties from java.security properties.
+                loadProperties(profileExtensionId, allInfo);
+
+                String hashProperty = profileExtensionId + ".desc.hash";
+                String hashValue = securityProps.getProperty(hashProperty);
+
+                // Hash value is optional in not base profiles.
```
keithc-ca, line 1137: Suggest "extension" is better than "not base".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                if (debug != null) {
+                    debug.println("\t\t\tNo constraints for security provider: " + providerName);
+                }
+                providerConstraints.put(providerName, constraints);
+                return;
+            }
+
+            // Check whether constraints are properly specified.
+            p = Pattern.compile(
+"\\[([\\+\\-]?)(\\{\\w+,[A-Za-z0-9\\.\\-/]+,[A-Za-z0-9\\=\\*\\|\\.:]+\\})(,\\{\\w+,[A-Za-z0-9\\.\\-/]+,[A-Za-z0-9\\=\\*\\|\\.:]+\\})*\\]");
```
keithc-ca, line 1400: Please split this string into logical parts; it will help with clarity.
I will have a closer look to verify that there is not unintended change in behavior.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+            m = p.matcher(providerInfo);
+
+            if (!m.find()) {
+                printStackTraceAndExit("Incorrect constraint definition for provider " + providerName);
+            }
+
+            String action = m.group(1);
+
+            // Parse all provided constraints.
+            p = Pattern.compile("\\{(\\w+),([A-Za-z0-9\\.\\-/]+),([A-Za-z0-9\\=\\*\\|\\.:]+)\\}");
```
keithc-ca, line 1410: Is this too restrictive (only ASCII letters, numbers, `.`, `-` and `/`)?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                                    "Constraint attributes format is incorrect: " + providerInfo);
+                        }
+                    }
+                }
+                Constraint constraint = new Constraint(inType, inAlgorithm, inAttributes);
+                constraints.add(constraint);
+            }
+
+            // Differeriante between add, remove and override.
+            if (!isNullOrBlank(action)) {
```
keithc-ca, line 1434: Please invert this `if` statement.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                }
+                List<Constraint> existingConstraints = providerConstraints.get(providerName);
+                if (existingConstraints == null) {
+                    existingConstraints = new ArrayList<>();
+                    providerConstraints.put(providerName, existingConstraints);
+                }
+                if (action.equals("+")) { // Appending constraints.
+                    existingConstraints.addAll(constraints);
+                } else { // Removing constraints.
+                    for (Constraint toRemove: constraints) {
```
keithc-ca, line 1448: Space both sides of `:`, please.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+            } else {
+                providerConstraints.put(providerName, constraints);
+            }
+
+            if (debug != null) {
+                debug.println("\t\t\tSuccessfully loaded constraints for security provider: " + providerName);
+            }
+        }
+
+        private void checkHexValues() {
```
keithc-ca, line 1464: Suggest `checkHashValues()` is a better name.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+            for (String profileID: profilesHashes.keySet()) {
+                String hashValue = profilesHashes.get(profileID);
```
keithc-ca, line 1466: Suggest iterating the entries:
```java
            for (Map.Entry<String, String> entry : profilesHashes.entrySet()) {
                String profileID = entry.getKey();
                String hashValue = entry.getValue();
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                List<String> allInfo = profilesInfo.get(profileID);
+
+                if (debug != null) {
+                    debug.println("Calculating hash for '" + profileID + "'.");
+                }
+                String[] hashInfo = hashValue.split(":");
```
keithc-ca, line 1472: Where is the validation that `hashValue` has exactly one colon?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                String digestAlgo = hashInfo[0].trim();
+                String expectedHash = hashInfo[1].trim();
+                try {
+                    MessageDigest md = MessageDigest.getInstance(digestAlgo);
+                    byte[] allInfoArray = allInfo.stream().sorted().collect(Collectors.joining("\n")).getBytes();
+                    byte[] resultHashArray = md.digest(allInfoArray);
```
keithc-ca, line 1478: This should use a specific charset, say, `StandardCharsets.UTF_8`.
How does this avoid including the `.hash` property in the computation?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                    StringBuffer hexString = new StringBuffer();
```
keithc-ca, line 1479: Please use a `StringBuilder` instead.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +887,615 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                    for (int i = 0; i < resultHashArray.length; i++) {
+                        hexString.append(String.format("%02x", resultHashArray[i] & 0xff));
+                    }
```
keithc-ca, line 1482: Suggest an enhanced `for` loop:
```java
                    for (byte hashByte : resultHashArray) {
                        hexString.append(String.format("%02x", hashByte & 0xff));
                    }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1115,82 +1524,130 @@ private void listAvailableProfiles() {
...
+            List<String> restrictions = new ArrayList<>();
+            for (String propertyName : propertyNames) {
+                if (propertyName.startsWith(profileToPrint + ".desc.") || propertyName.startsWith(profileToPrint + ".fips.")) {
+                    descKeys.add(propertyName + securityProps.getProperty(propertyName));
+                } else if (propertyName.startsWith(profileToPrint + ".jce.provider.")) {
+                    providers.add(propertyName + securityProps.getProperty(propertyName));
+                } else if (propertyName.startsWith(profileToPrint)) {
+                    restrictions.add(propertyName + securityProps.getProperty(propertyName));
+                }
+
```
keithc-ca, line 1540: Please remove this blank line.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1115,82 +1524,130 @@ private void listAvailableProfiles() {
...
+            String value = securityProps.getProperty(propertyKey);
+            value = parseProperty(value);
+            String newValue = null;
+            if (value != null) {
+                // Add to info to create hash.
+                allInfo.add(propertyKey + "=" + value);
+
+                // Check if property overrides, adds to or removes from previous value.
+                String existingValue = profileProperties.get(property);
+                if (!value.isBlank() && value.startsWith("+")) {
```
keithc-ca, line 1587: The `!value.isBlank()` part is unnecessary (same on line 1603).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1115,82 +1524,130 @@ private void listAvailableProfiles() {
...
+                    if (isPropertyAppendable(property)) {
```
keithc-ca, line 1588: Suggest inverting this `if` (and the one on line 1604).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1115,82 +1524,130 @@ private void listAvailableProfiles() {
...
+                        // Append additional values to property.
+                        value = value.substring(1, value.length()).trim();
```
keithc-ca, line 1590: The length isn't needed here (nor on line 1606):
```java
                        value = value.substring(1).trim();
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1115,82 +1524,130 @@ private void listAvailableProfiles() {
...
+                    }
+                } else if (!value.isBlank() && value.startsWith("-")) {
+                    if (isPropertyAppendable(property)) {
+                        // Remove values from property.
+                        value = value.substring(1, value.length()).trim();
+                        if (!value.isBlank()) {
+                            List<String> existingValues = Arrays.asList(existingValue.split(","));
+                            existingValues = existingValues.stream().map(v -> v.trim()).collect(Collectors.toList());
+                            List<String> valuesToRemove = Arrays.asList(value.split(","));
+                            for (String valueToRemove : valuesToRemove) {
```
keithc-ca, line 1611: We don't need to create so many `List` objects:
```java
                            List<String> existingValues = Stream.of(existingValue.split(","))
                                                .map(v -> v.trim())
                                                .collect(Collectors.toList());
                            String[] valuesToRemove = value.split(",");
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1115,82 +1524,130 @@ private void listAvailableProfiles() {
...
+                                if (existingValues.contains(valueToRemove.trim())) {
+                                    existingValues.remove(valueToRemove.trim());
```
keithc-ca, line 1613: `remove()` will answer whether the list changed:
```java
                                if (!existingValues.remove(valueToRemove.trim())) {
                                    printStackTraceAndExit("Value '" + valueToRemove + "' is not in existing values.");
                                }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1207,64 +1664,67 @@ private static String parseProperty(String string) {
...
+                if (other.algorithm != null)
+                    return false;
+            } else if (!algorithm.equals(other.algorithm))
+                return false;
+            if (attributes == null) {
+                if (other.attributes != null)
+                    return false;
+            } else if (!attributes.equals(other.attributes))
+                return false;
+            return true;
```
keithc-ca, line 1727: This is missing many braces (`{` and `}`) and can be dramatically simplified by using `Objects.equals()`:
```java
            if (this == obj) {
                return true;
            }
            if (obj instanceof Constraint other) {
                return Objects.equals(type, other.type)
                    && Objects.equals(algorithm, other.algorithm)
                    && Objects.equals(attributes, other.attributes);
            }
            return false;
```
Also, when overriding `equals()` you should also override `hashCode()`:
```java
        @Override
        public int hashCode() {
            return Objects.hash(type, algorithm, attributes);
        }
```

keithc-ca (changes requested):

src/java.base/share/conf/security/java.security
```diff
@@ -163,6 +163,7 @@ RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.fips = true
 RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.number = Certificate #XXX
 RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.policy = https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/
 RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.sunsetDate = 2026-09-21
+RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.hash = SHA256:f5b04d07f6fd5d11374b23dd96110b231d0241096563ae55cf3ea6fb1788d97c
```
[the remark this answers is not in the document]
KostasTsiounis: Moved.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -630,11 +635,43 @@ private static void printStackTraceAndExit(String message) {
...
+    /**
+     * Get the provider name defined in provider construction method.
+     *
+     * @param providerName provider name or provider with packages
+     * @return provider name defined in provider construction method
+     */
+    private static String getProvidersSimpleName(String providerName) {
+        if (providerName.equals("com.sun.security.sasl.Provider")) {
+            // The main class for the SunSASL provider is com.sun.security.sasl.Provider.
+            return "SunSASL";
```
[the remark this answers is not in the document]
KostasTsiounis: Yeah, I guess one of us would have to rebase depending on which one is merged first.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Done.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Moved most fields to the constructors. I'm not sure about the lists and map though. Think I should move those too?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: We only get here if the profile defines the `.extends` property. So, I'm not sure how this could lead to infinite recursion.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Split into parts and added some comments for them.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: We only allow these on purpose. The algorithm part of the constraint can only have those options, so we don't want wrong definitions.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Added a check to make sure the split produces exactly two String.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Not really sure what you mean about the `StandardCharsets.UTF_8`.

The `.hash` property is not added to the `StringBuilder`, so it's not included in the computation.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Removed in both places.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Inverted.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Removed in both places.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: I just figured that you mean something like a cyclical extension, where 2 profiles extend each other. So, I added a check to make sure the profile hasn't already been parsed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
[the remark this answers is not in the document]
keithc-ca: Actually, I expect the whole method to go away in #758.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -666,263 +719,49 @@ private static final class RestrictedSecurityProperties {
...
+                                             String jdkTlsDisabledAlgorithms,
+                                             String jdkTlsEphemeralDHKeySize,
+                                             String jdkTlsLegacyAlgorithms,
+                                             String jdkCertpathDisabledAlgorithms,
+                                             String jdkSecurityLegacyAlgorithms,
+                                             String keyStoreType,
+                                             String keyStore,
+                                             String jdkSecureRandomProvider,
+                                             String jdkSecureRandomAlgorithm,
+                                             String jdkFipsMode) {
```
keithc-ca, line 736: Rather than extract all those values at the call site (with the potential to get the order wrong), suggest passing the `ProfileParser` and allowing this constructor to query the values here. `ProfileParser` would simply
```java
        private RestrictedSecurityProperties getProperties() {
            return new RestrictedSecurityProperties(profileID, this);
        }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -536,22 +555,16 @@ private static void setProperties(Properties props) {
             String jdkPropsName = entry.getKey();
             String propsNewValue = entry.getValue();
 
-            String propsOldValue = props.getProperty(jdkPropsName);
-            if (isNullOrBlank(propsOldValue)) {
-                propsOldValue = "";
-            }
```
keithc-ca, line 542: Please explain why we don't want to combine `propsOldValue` and `propsNewValue` (below) as before.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -630,11 +651,43 @@ private static void printStackTraceAndExit(String message) {
...
+
     /**
      * This class is used to save and operate on restricted security
      * properties which are loaded from the java.security file.
      */
     private static final class RestrictedSecurityProperties {
+        private final String profileID;
 
         private String descName;
         private boolean descIsDefault;
```
keithc-ca, line 693: All fields of `RestrictedSecurityProperties` can now be `final`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +950,864 @@ private void listUsedProfile() {
...
+        private final List<String> providersSimpleName;
+        // The map is keyed by provider name.
+        private final Map<String, List<Constraint>> providerConstraints;
+
+        private final String profileID;
+
+        private Map<String, String> profilesHashes;
+        private Map<String, List<String>> profilesInfo;
+
+        private Set<String> parsedProfiles;
```
keithc-ca, line 1023: These three fields can all be `final`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +950,864 @@ private void listUsedProfile() {
...
+            profileProperties = new HashMap<>();
+
+            providers = new ArrayList<>();
+            providersSimpleName = new ArrayList<>();
+            providerConstraints = new HashMap<>();
+
+            profilesHashes = new HashMap<>();
+            profilesInfo = new HashMap<>();
+
+            hashedProfiles = new HashSet<>();
```
keithc-ca, line 1048: I think you meant `parsedProfiles` instead of `hashedProfiles`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +950,864 @@ private void listUsedProfile() {
...
+                    // Extended profile id has at least 2 dots (meaning it's a full profile id).
+                    for (Object keyObject : securityProps.keySet()) {
+                        if (keyObject instanceof String key) {
+                            if (key.startsWith(potentialExtendsProfileID + ".desc") ||
+                                key.startsWith(potentialExtendsProfileID + ".fips") ||
+                                key.startsWith(potentialExtendsProfileID + ".tls") ||
+                                key.startsWith(potentialExtendsProfileID + ".jce") ||
+                                key.startsWith(potentialExtendsProfileID + ".javax") ||
+                                key.startsWith(potentialExtendsProfileID + ".securerandom")
+                            ) {
```
keithc-ca, line 1123: Please put `||` at the beginning of each continuation line:
```java
                            if (key.startsWith(potentialExtendsProfileID + ".desc")
                            ||  key.startsWith(potentialExtendsProfileID + ".fips")
                            ||  key.startsWith(potentialExtendsProfileID + ".tls")
                            ||  key.startsWith(potentialExtendsProfileID + ".jce")
                            ||  key.startsWith(potentialExtendsProfileID + ".javax")
                            ||  key.startsWith(potentialExtendsProfileID + ".securerandom")
                            ) {
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +950,864 @@ private void listUsedProfile() {
...
+
+                if (providerInfo.trim().isEmpty()) {
+                    printStackTraceAndExit(
+                        "Cannot specify an empty provider in position "
+                                + pNum + ". Nothing specified before.");
+                }
+
+                allInfo.add(property + "=" + providerInfo);
+
+                boolean isFIPS140_2 = "140-2".equals(securityProps.getProperty(profileID + ".fips.mode"));
```
keithc-ca, line 1321: I suggest we should have a helper method to check if a given profile is FIPS-140-2, rather than have several copies.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +950,864 @@ private void listUsedProfile() {
...
+            // Check whether constraints are properly specified.
+            p = Pattern.compile(
+                "\\["
+                + "([\\+\\-]?)"                     // Option to append or remove
+                + "(\\{"                            // At least one constraint
+                    + "\\w+,"
+                    + "[A-Za-z0-9\\.\\-/_]+,"
+                    + "[A-Za-z0-9\\=\\*\\|\\.:]+"
+                + "\\})"
+                + "(,\\{"                           // More constraints [OPTIONAL]
```
keithc-ca, line 1462: Please avoid capitals in these comments (for sentence fragments).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +950,864 @@ private void listUsedProfile() {
...
+                    debug.println("\t\t\tNo constraints for security provider: " + providerName);
+                }
+                providerConstraints.put(providerName, constraints);
+                return;
+            }
+
+            // Check whether constraints are properly specified.
+            p = Pattern.compile(
+                "\\["
+                + "([\\+\\-]?)"                     // Option to append or remove
```
keithc-ca, line 1456: This pattern does not accept the same set of constraints as before. For example, whitespace between `[` and `{` was allowed before.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: We need a consistent and predictable encoding of the string (using `getByte()` involves `Charset.defaultCharset()` which is affected by the `file.encoding` system property):
```java
                    byte[] allInfoArray = allInfo.stream().sorted()
                            .collect(Collectors.joining("\n"))
                            .getBytes(StandardCharsets.UTF_8);
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +950,864 @@ private void listUsedProfile() {
...
+                    StringBuilder hexString = new StringBuilder();
+                    for (byte hashByte : resultHashArray) {
+                        hexString.append(String.format("%02x", hashByte & 0xff));
+                    }
+                    String resultHashHex = hexString.toString();
+                    if (debug != null) {
+                        debug.println("\tCalculated hash for '" + profileID + "': " + resultHashHex);
+                        debug.println("\tExpected hash for '" + profileID + "': " + expectedHash);
+                    }
+                    if (!resultHashHex.equals(expectedHash)) {
```
keithc-ca, line 1560: This should tolerate case differences, using `equalsIgnoreCase()`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +950,864 @@ private void listUsedProfile() {
...
+        private static boolean isPropertyAppendable(String property) {
+            switch (property) {
+                case "jdkCertpathDisabledAlgorithms":
+                case "jdkSecurityLegacyAlgorithms":
+                case "jdkTlsDisabledAlgorithms":
+                case "jdkTlsDisabledNamedCurves":
+                case "jdkTlsLegacyAlgorithms":
+                    return true;
+                default:
+                    return false;
```
keithc-ca, line 1728: Please don't indent case labels:
```java
            switch (property) {
            case "jdkCertpathDisabledAlgorithms":
            case "jdkSecurityLegacyAlgorithms":
            case "jdkTlsDisabledAlgorithms":
            case "jdkTlsDisabledNamedCurves":
            case "jdkTlsLegacyAlgorithms":
                return true;
            default:
                return false;
            }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +950,864 @@ private void listUsedProfile() {
...
+                + "\\s*(\\s\\[\\s*"                                 // Constraints [OPTIONAL]
+                    + "([\\+\\-])?\\s*"
+                    + "([A-Za-z0-9\\{\\}\\-\\.\\=\\*\\|:,/_\\s]+)"
+                + "\\])?\\s*$");
+            Matcher m = p.matcher(providerInfo);
+            if (m.find()) {
+                if (!isFIPS140_2) {
+                    String providerName = m.group(1);
+                    if (providerName.indexOf('.') == -1) {
+                        printStackTraceAndExit("Provider can only be specified using"
```
keithc-ca, line 1759: Suggest "must" instead of "can only" and make it clear a class name is required:
```java
                        printStackTraceAndExit("Provider must be specified using"
                                + " the fully-qualified class name: " + providerName);
```

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: We didn't want the property to append to what was there before, but rather completely override it.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: That's because I did `providerInfo = providerInfo.replaceAll("\\s+", "");` a few lines above.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Yep, mistake. Fixed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Added a helper method.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -23,13 +23,17 @@
...
 import java.security.AccessController;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.security.PrivilegedAction;
 import java.security.Provider.Service;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
+import java.util.Arrays;
```
keithc-ca, line 36: Please remove this unused import.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -225,6 +231,25 @@ public static boolean isFIPSEnabled() {
...
+    public static boolean isFIPS1402() {
+        if (securityEnabled) {
+            return "140-2".equals(restricts.jdkFipsMode);
+        }
+        return false;
+    }
+
+    private static boolean isFIPS1402Profile(String profileID, Properties securityProps) {
+        return "140-2".equals(securityProps.getProperty(profileID + ".fips.mode"));
+    }
```
keithc-ca, line 251: This can be an instance method in `ProfileParser` so the `Properties` object is implied.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -666,263 +724,41 @@ private static final class RestrictedSecurityProperties {
...
+            this.jdkFipsMode = parser.getProperty("jdkFipsMode");
+
+            this.providers = new ArrayList<>(parser.providers);
+            this.providersSimpleName = new ArrayList<>(parser.providersSimpleName);
+            this.providerConstraints = new HashMap<>(parser.providerConstraints.entrySet()
+                                                                               .stream()
+                                                                               .collect(Collectors.toMap(
+                                                                                    e -> e.getKey(),
+                                                                                    e -> e.getValue().toArray(new Constraint[0])
+                                                                               )));
```
keithc-ca, line 761: The map returned by `collect()` can be used directly:
```c
            this.providerConstraints = parser.providerConstraints
                                             .entrySet()
                                             .stream()
                                             .collect(Collectors.toMap(
                                                     e -> e.getKey(),
                                                     e -> e.getValue().toArray(new Constraint[0])
                                                     ));
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +947,840 @@ private void listUsedProfile() {
...
+        private RestrictedSecurityProperties getProperties() {
+            RestrictedSecurityProperties rsProperties = 
+                    new RestrictedSecurityProperties(this.profileID, this);
+
+            if (debug != null) {
+                // Print information of utilized security profile.
+                rsProperties.listUsedProfile();
+            }
+
+            return rsProperties;
```
keithc-ca, line 1060: Suggest moving the call to `` inside the constructor to simplify this:
```java
            return new RestrictedSecurityProperties(profileID, this);
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: A cycle might involve one or more profiles; I think the code will handle all those cases.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1124,147 +947,840 @@ private void listUsedProfile() {
...
+            // Check whether constraints are properly specified.
+            p = Pattern.compile(
+                "\\["
+                + "([\\+\\-]?)"                     // option to append or remove
+                + "(\\{"                            // at least one constraint
+                    + "\\w+,"
+                    + "[A-Za-z0-9\\.\\-/_]+,"
+                    + "[A-Za-z0-9\\=\\*\\|\\.:]+"
+                + "\\})"
+                + "(,\\{"                           // more constraints [OPTIONAL]
```
keithc-ca, line 1432: nit: Please use lowercase "optional".

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Moved.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Moved the debug to constructor.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: That's a different difference in the language accepted. It's not clear that's appropriate.

KostasTsiounis: We only get here after passing the original check in `checkProviderFormat()`, which means the original properties have already been verified allowing for spaces. I just remove the spaces at this point to make things more readable and easily parsable.

KostasTsiounis:
I removed the `isFIPS1402()` method, as it's no longer needed by https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/758

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +924,604 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+            }
+
+            // Check whether constraints are properly specified.
+            p = Pattern.compile(
+                "\\["
+                + "([\\+\\-]?)"                     // option to append or remove
+                + "(\\{"                            // at least one constraint
+                    + "\\w+,"
+                    + "[A-Za-z0-9\\.\\-/_]+,"
+                    + "[A-Za-z0-9\\=\\*\\|\\.:]+"
```
keithc-ca, line 1411: Please avoid escapes where they're not needed (within `[]`, `.`, `=`, `*`, `|`, `:` are treated literally, and `-` (if the first or last character) need not be escaped). In fact, the javadoc for `Pattern` says:
> It is an error to use a backslash prior to any alphabetic character that does not denote an escaped construct

When I originally suggested decomposing patterns, I was thinking the parts could be named, like this:
```java
            final String typeRE = "\\w+";
            final String algoRE = "[A-Za-z0-9./_-]+";
            final String attrRE = "[A-Za-z0-9=*|.:]+";
            final String consRE = "\\{(" + typeRE + "),(" + algoRE + "),(" + attrRE + ")\\}";

            p = Pattern.compile(
                "\\["
                + "([+-]?)"             // option to append or remove
                + consRE                // at least one constraint
                + "(," + consRE + ")*"  // more constraints [optional]
                + "\\]");
```
With the parentheses included in `consRE` above, that pattern can directly be reused to parse constraints:
```java
            // Parse all provided constraints.
            p = Pattern.compile(consRE);
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +924,604 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+            for (int i = numOfExistingProviders; i >= providerPos; i--) {
+                if (debug != null) {
+                    debug.println("\t\t\tNot the last provider. More to be removed.");
+                }
+
+                String providerInfo = securityProps.getProperty(profileExtensionId + ".jce.provider." + i);
+                if ((providerInfo == null) || !providerInfo.trim().isEmpty()) {
+                    printStackTraceAndExit(
+                        "Cannot specify an empty provider in position "
+                                + providerPos + " without empty the ones after it.");
```
keithc-ca, line 1234: Suggest "without empty" should be "with non-empty".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1207,64 +1697,71 @@ private static String parseProperty(String string) {
...
-                    if (deque.isEmpty() || (deque.removeFirst().charValue() != ch)) {
-                        return false;
+        private static void checkProviderFormat(String providerInfo, boolean update, boolean isFIPS140_2) {
+            Pattern p = Pattern.compile(
+                "^([A-Za-z0-9\\-\\.]+)"                             // provider name
+                + "\\s*(\\s[A-Za-z0-9\\$\\{\\}\\./]+)?"             // configuration file [optional]
+                + "\\s*(\\s\\[\\s*"                                 // constraints [optional]
+                    + "([\\+\\-])?\\s*"
+                    + "([A-Za-z0-9\\{\\}\\-\\.\\=\\*\\|:,/_\\s]+)"
+                + "\\])?\\s*$");
```
keithc-ca, line 1707: Unnecessary escapes, and parentheses; naming parts adds clarity:
```java
            final String nameRE = "[A-Za-z0-9.-]+";
            final String fileRE = "[A-Za-z0-9./${}]+";
            Pattern p = Pattern.compile(
                  "^(" + nameRE + ")"                   // provider name
                + "\\s*"
                + "(" + fileRE + ")?"                   // configuration file [optional]
                + "\\s*"
                + "(\\["                                // constraints [optional]
                + "\\s*"
                    + "([+-])?"                         // action [optional]
                    + "[A-Za-z0-9{}.=*|:,/_\\s-]+"
                    + "\\])?"
                + "\\s*"
                + "$");
```
Should `fileRE` also accept `\`?
The string of constraints could be simplified to `[^\\]]+` (a sequence of one or more characters other than `]`).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1207,64 +1697,71 @@ private static String parseProperty(String string) {
...
+                        printStackTraceAndExit("Provider must be specified using"
+                                + " the fully-qualified class name: " + providerName);
                     }
-                    break;
-                default:
-                    break;
                 }
+
+                String symbol = m.group(4);
+                if (!update && (symbol != null)) {
```
keithc-ca, line 1719: Suggest `symbol` should be `action` and and `isNullOrBlank()` should be used (like in `setConstraints()`):
```java
                String action = m.group(4);
                if (!update && !isNullOrBlank(action)) {
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1207,64 +1697,71 @@ private static String parseProperty(String string) {
...
+                + "\\s*(\\s[A-Za-z0-9\\$\\{\\}\\./]+)?"             // configuration file [optional]
+                + "\\s*(\\s\\[\\s*"                                 // constraints [optional]
+                    + "([\\+\\-])?\\s*"
+                    + "([A-Za-z0-9\\{\\}\\-\\.\\=\\*\\|:,/_\\s]+)"
+                + "\\])?\\s*$");
+            Matcher m = p.matcher(providerInfo);
+            if (m.find()) {
+                if (!isFIPS140_2) {
+                    String providerName = m.group(1);
+                    if (providerName.indexOf('.') == -1) {
```
keithc-ca, line 1712: Suggest the name should not be permitted to start with `.`
```java
                    if (providerName.indexOf('.') <= 0) {
```

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -666,262 +704,45 @@ private static final class RestrictedSecurityProperties {
...
+            this.jdkFipsMode = parser.getProperty("jdkFipsMode");
+
+            this.providers = new ArrayList<>(parser.providers);
+            this.providersSimpleName = new ArrayList<>(parser.providersSimpleName);
+            this.providerConstraints = parser.providerConstraints.entrySet()
+                                                                 .stream()
+                                                                 .collect(Collectors.toMap(
+                                                                    e -> e.getKey(),
+                                                                    e -> e.getValue().toArray(new Constraint[0])
+                                                                 ));
```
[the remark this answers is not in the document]
KostasTsiounis: Moved to separate line.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -666,262 +704,45 @@ private static final class RestrictedSecurityProperties {
...
-                        debug.println("\t\tSuccessfully loaded constraints for security provider: " + providerName);
-                    }
-                } else {
-                    printStackTraceAndExit("Constraint format is incorrect: " + providerInfo);
-                }
-            }
-
-            if (debug != null) {
-                debug.println("\tAll constraints of restricted security profile successfully loaded.");
+                rsProperties.listUsedProfile();
```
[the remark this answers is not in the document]
KostasTsiounis: Sorry, left from where it was before. Changed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +924,604 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+
+            // Initialize the properties.
+            init(profileID);
+        }
+
+        private RestrictedSecurityProperties getProperties() {
+            return new RestrictedSecurityProperties(this.profileID, this);
+        }
+
+        private static boolean isFIPS1402Profile(String profileID) {
```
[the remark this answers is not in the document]
KostasTsiounis: Forgot to remove it. Done.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +924,604 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                    // Extended profile id has at least 2 dots (meaning it's a full profile id).
+                    for (Object keyObject : securityProps.keySet()) {
+                        if (keyObject instanceof String key) {
+                            if (key.startsWith(potentialExtendsProfileID + ".desc")
+                            ||  key.startsWith(potentialExtendsProfileID + ".fips")
+                            ||  key.startsWith(potentialExtendsProfileID + ".tls")
+                            ||  key.startsWith(potentialExtendsProfileID + ".jce")
+                            ||  key.startsWith(potentialExtendsProfileID + ".javax")
+                            ||  key.startsWith(potentialExtendsProfileID + ".securerandom")
+                            ) {
```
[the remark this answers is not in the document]

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +924,604 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+                                // If even one security property is found for this profile id, move on.
```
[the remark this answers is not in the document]
KostasTsiounis: Changed the wording to explain better.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +924,604 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+
+            for (int pNum = 1;; ++pNum) {
+                String property = profileID + ".jce.provider." + pNum;
+                String providerInfo = securityProps.getProperty(property);
+
+                if (providerInfo == null) {
+                    break;
+                }
+
+                if (providerInfo.trim().isEmpty()) {
```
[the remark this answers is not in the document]
KostasTsiounis: Replaced everywhere.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Makes sense. I didn't know we don't need to escape special characters when in `[]`. Changed to suggestion.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: I changed to `without emptying`, which is what I wanted to say actually.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: I added the backslash to the `fileRE`. I don't think we want to allow all characters other than `]`. We want to avoid as many potential user typos as possible by only allowing what is allowed to be used.

Changed the pattern to suggestion too.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +924,604 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+        }
+
+        private void setConstraints(String providerName, String providerInfo, boolean providerChanged) {
+            if (debug != null) {
+                debug.println("\t\tLoading constraints for security provider: " + providerName);
+            }
+
+            List<Constraint> constraints = new ArrayList<>();
+
+            providerInfo = providerInfo.replaceAll("\\s+", "");
```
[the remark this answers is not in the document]
keithc-ca: We haven't provided an example `java.security` that has significant whitespace within constraints, but how do we know that we will never need them? For example, suppose a constraint needs to specify an attribute value which must include a space?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: Most characters between `[` and `]` don't need to be escaped, but some do - see https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: The backslash must be escaped in the Java literal and also in the regular expression. In Java you have to say `"\\\\"` to get a string with the two backslashes needed in the pattern.

KostasTsiounis: Fixed that.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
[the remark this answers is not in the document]
KostasTsiounis: All of the attributes provided in existing providers are `ImplementedIn=Software`, or single values or lists of things like `ValidationAlgorithm`, `ThreadSafe`, `SupportedKeyClasses`, `KeySize` etc. I haven't found anything else and I cannot think of anything that would need a space. Anything that comes to your mind?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -666,265 +704,49 @@ private static final class RestrictedSecurityProperties {
...
+
+            this.providers = new ArrayList<>(parser.providers);
+            this.providersSimpleName = new ArrayList<>(parser.providersSimpleName);
+            this.providerConstraints = parser.providerConstraints
+                                             .entrySet()
+                                             .stream()
+                                             .collect(Collectors.toMap(
+                                                e -> e.getKey(),
+                                                e -> e.getValue().toArray(new Constraint[0])
+                                             ));
```
keithc-ca, line 742: These three lines should be indented eight spaces more than line 739.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1207,64 +1699,78 @@ private static String parseProperty(String string) {
...
+            final String fileRE = "[A-Za-z0-9./\\\\${}]+";
+            Pattern p = Pattern.compile(
+                  "^(" + nameRE + ")"                   // provider name
+                + "\\s*"
+                + "(" + fileRE + ")?"                   // configuration file [optional]
+                + "\\s*"
+                + "(\\["                                // constraints [optional]
+                    + "\\s*"
+                    + "([+-])?"                         // action [optional]
+                    + "[A-Za-z0-9{}.=*|:,/_\\s-]+"      // constraint definition [optional]
```
keithc-ca, line 1713: I don't think the comment belongs here (at least the "optional" part anyway).

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: I don't think that says what we want. The properties file is broken in that it has a non-empty provider after an empty one; "emptying" is what someone might need to do to fix it, but it doesn't describe the situation discovered here.

KostasTsiounis: Changed to `when non-empty ones are specified after it.`. Does that make sense?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Fixed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Removed the "optional" part.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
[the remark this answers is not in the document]
keithc-ca: I don't have a specific example to offer, but I was thinking along the lines of a "transformation" (the first argument for `Cipher.getInstance(String, String)`) that might need to contain one or more spaces.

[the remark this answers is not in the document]
KostasTsiounis: I haven't encountered a transformation that requires a space anywhere so far. I think we should remove the spaces to be sure that we avoid typos at this point. We can always make another change in the future if such a scenario comes up, even though I don't think that's a likely one.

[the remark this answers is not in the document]
keithc-ca: I don't know enough to comment on the likelihood of such a problem, but I couldn't find anything in this repository that would require such spaces, so that seems a reasonable posture.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1070,23 +925,605 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+            init(profileID);
+        }
+
+        private RestrictedSecurityProperties getProperties() {
+            return new RestrictedSecurityProperties(this.profileID, this);
+        }
+
+        private boolean isFIPS1402Profile(String profileID) {
+            return "140-2".equals(securityProps.getProperty(profileID + ".fips.mode"));
+        }
```
keithc-ca, line 1043: I suggest it doesn't make sense to add this or the uses of the method given the plan is to require fully-qualified class names for all profiles (FIPS140-2 or not).

KostasTsiounis: I also use it to determine whether to check the hash value in older profiles. Should I just add the check there directly and remove the method?

keithc-ca: Sorry. Yes, to decide whether to require a hash, we should keep this method and its use in that context.
We can remove the `isFIPS140_2` parameter from `checkProviderFormat()` and `parseProvider()`.

KostasTsiounis: Done.

keithc-ca:
I think the changes are good. Please squash and ensure that the commit message accurately and adequately describes these changes.

KostasTsiounis:
Squashed and updated commit message to match issue description.

keithc-ca:
Jenkins copyright check

keithc-ca:
Jenkins line endings check

keithc-ca:

keithc-ca:
Testing in https://openj9-jenkins.osuosl.org/job/Grinder/3673 failed because a newer version of jtreg is required. See https://github.com/adoptium/TKG/pull/573 and https://github.com/adoptium/ci-jenkins-pipelines/pull/1051 that aim to address that.

In the meantime, you could start back-porting this to jdk23, jdk22, etc.

keithc-ca:
Merging on the strength of testing results in https://github.com/ibmruntimes/openj9-openjdk-jdk23/pull/2.

PR #795, opened 2024-05-29
Remove unnecessary changes to EC key implementations

KostasTsiounis:
There is no need for additional code in the `ECPrivateKeyImpl` and `ECPublicKeyImpl` classes anymore, compared to the upstream, so the IBM copyright is also no longer needed.

Signed-off by: Kostas Tsiounis <kostas.tsiounis@ibm.com>

### DIFF
--- a/src/java.base/share/classes/sun/security/ec/ECPrivateKeyImpl.java
+++ b/src/java.base/share/classes/sun/security/ec/ECPrivateKeyImpl.java
@@ -23,12 +23,6 @@
  * questions.
  */
 
-/*
- * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2024 All Rights Reserved
- * ===========================================================================
- */
-
 package sun.security.ec;
 
 import java.io.IOException;
@@ -40,8 +34,6 @@ import java.security.*;
 import java.security.interfaces.*;
 import java.security.spec.*;
 import java.util.Arrays;
-
-import jdk.crypto.jniprovider.NativeCrypto;
 
 import sun.security.ec.point.AffinePoint;
 import sun.security.ec.point.MutablePoint;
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/sun/security/ec/ECPrivateKeyImpl.java
```diff
@@ -23,12 +23,6 @@
  * questions.
  */
 
-/*
- * ===========================================================================
```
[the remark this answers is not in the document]
KostasTsiounis: Removed all changes compared to upstream.

src/java.base/share/classes/sun/security/ec/ECPublicKeyImpl.java
```diff
@@ -40,8 +34,6 @@
 import java.security.interfaces.*;
 import java.security.spec.*;
 
-import jdk.crypto.jniprovider.NativeCrypto;
```
keithc-ca, line 43: The import of `BigInteger` is not present upstream, nor is the final newline; compare with c86650f47c04e5d349dc22017fa14b94f93ad6df.

keithc-ca (changes requested):

src/java.base/share/classes/sun/security/ec/ECPublicKeyImpl.java
KostasTsiounis: Sorry about that. I was looking at the wrong version, butI think I got everything now.

src/java.base/share/classes/sun/security/ec/ECPrivateKeyImpl.java
```diff
@@ -246,4 +237,4 @@ private void readObject(ObjectInputStream stream)
         throw new InvalidObjectException(
                 "ECPrivateKeyImpl keys are not directly deserializable");
     }
-}
+}
```
keithc-ca, line 240: Please restore this newline from upstream (but that's not present upstream in jdk22).

keithc-ca (changes requested):

src/java.base/share/classes/sun/security/ec/ECPrivateKeyImpl.java
KostasTsiounis: Done.

keithc-ca (approved):

PR #800, opened 2024-06-04
Improve interaction between CRIU and CRaC configuration options

keithc-ca:
CRaC support requires CRIU support. If CRIU support is disabled explicitly and CRaC is enabled explicitly, configuration will now fail rather than incur a compile error later.

Otherwise, if CRaC support is requested explictly, CRIU support will implicitly be enabled if that is not the default for CRIU.

Also, if CRIU support is explicitly disabled, CRaC will be implicitly disabled if that is not the default for CRaC.

### DIFF
--- a/closed/autoconf/custom-hook.m4
+++ b/closed/autoconf/custom-hook.m4
@@ -41,8 +41,7 @@ AC_DEFUN_ONCE([CUSTOM_EARLY_HOOK],
   OPENJ9_CONFIGURE_CMAKE
   OPENJ9_CONFIGURE_COMPILERS
   OPENJ9_CONFIGURE_CONTINUATION_PROFILE
-  OPENJ9_CONFIGURE_CRAC_SUPPORT
-  OPENJ9_CONFIGURE_CRIU_SUPPORT
+  OPENJ9_CONFIGURE_CRAC_AND_CRIU_SUPPORT
   OPENJ9_CONFIGURE_CUDA
   OPENJ9_CONFIGURE_DDR
   OPENJ9_CONFIGURE_DEMOS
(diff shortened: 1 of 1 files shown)


JasonFengJ9:
`CRaC` always requires `CRIU`.
Can we assume `--enable-crac-support` implies `--enable-criu-support` instead of checking `CRIU` and throwing an error otherwise? 
* `--enable-criu-support` enables `CRIU`
* `--enable-crac-support` enabled `CRIU` and `CRaC`

@tajila thoughts?

keithc-ca:
I suppose we could check the options the other way around, so explicitly saying `--enable-crac-support` would provide a default for CRIU, but I'm not sure that's much better or substantially different. My motivation for this change was due to compile errors that result if you just say `--disable-criu-support` on a platform where CRaC is enabled by default. This will prevent those compile errors by disabling CRaC if CRIU is disabled.

There are three valid combinations of the two configuration options; this does not change that, but it does point out when the fourth, invalid combination is requested.

JasonFengJ9:
How about the following?
* `--enable-criu-support` enables `CRIU`
* `--enable-crac-support` enabled `CRIU` and `CRaC`
* `--disable-criu-support` disables `CRIU` and `CRaC`
* `--disable-crac-support` disables `CRaC`

keithc-ca:
> How about the following?

That seems reasonable to me. @tajila Do you agree?

Using `--enable-crac-support` in combination with `--disable-criu-support` should still be disallowed.

JasonFengJ9:
> Using --enable-crac-support in combination with --disable-criu-support should still be disallowed.

Right, I think a configuration error can tell the user this is an invalid combination.

tajila:
> That seems reasonable to me. @tajila Do you agree?

> Using --enable-crac-support in combination with --disable-criu-support should still be disallowed.


Looks good to me.

keithc-ca:
Updated as suggested.

closed/autoconf/custom-hook.m4
```diff
@@ -340,57 +339,86 @@ AC_DEFUN([OPENJ9_PLATFORM_EXTRACT_VARS_FROM_CPU],
...
+      origin_crac="implicitly disabled"
+      enable_crac_support=no
+    else
+      origin_crac=default
+      enable_crac_support=$default_crac
+    fi
   else
     AC_MSG_ERROR([--enable-crac-support accepts no argument])
   fi
-  AC_SUBST(OPENJ9_ENABLE_CRAC_SUPPORT)
```
JasonFengJ9, line 367: > AC_MSG_ERROR([--enable-crac-support accepts no argument])

I think this existing message is not correct, maybe something like `AC_MSG_ERROR([Unrecognised --enable-crac-support argument])`?

Similarly for `AC_MSG_ERROR([--enable-criu-support accepts no argument])`.

closed/autoconf/custom-hook.m4
```diff
@@ -340,57 +339,86 @@ AC_DEFUN([OPENJ9_PLATFORM_EXTRACT_VARS_FROM_CPU],
...
-      *)
-        AC_MSG_RESULT([no (default)])
-        ;;
-    esac
+    # Adjust if CRaC is explicitly enabled.
+    if test "x$enable_crac_support" = xyes && test "x$default_criu" = xno ; then
+      origin_criu="implicitly enabled"
+      enable_criu_support=yes
+    else
+      origin_criu=default
```
JasonFengJ9, line 396:
```suggestion
      origin_criu="default"
```

closed/autoconf/custom-hook.m4
```diff
@@ -340,57 +339,86 @@ AC_DEFUN([OPENJ9_PLATFORM_EXTRACT_VARS_FROM_CPU],
...
-      *)
-        AC_MSG_RESULT([no (default)])
-        ;;
-    esac
+    # Adjust if CRUI is explicitly disabled.
+    if test "x$enable_criu_support" = xno && test "x$default_crac" = xyes ; then
+      origin_crac="implicitly disabled"
+      enable_crac_support=no
+    else
+      origin_crac=default
```
JasonFengJ9, line 379:
```suggestion
      origin_crac="default"
```

closed/autoconf/custom-hook.m4
```diff
@@ -340,57 +339,86 @@ AC_DEFUN([OPENJ9_PLATFORM_EXTRACT_VARS_FROM_CPU],
...
-        OPENJ9_ENABLE_CRIU_SUPPORT=true
-        ;;
-      *)
-        AC_MSG_RESULT([no (default)])
-        ;;
-    esac
+    # Adjust if CRaC is explicitly enabled.
+    if test "x$enable_crac_support" = xyes && test "x$default_criu" = xno ; then
+      origin_criu="implicitly enabled"
+      enable_criu_support=yes
```
JasonFengJ9, line 394: If `$default_criu` is `no`, this means `CRIU` is not supported in this platform, and can't be enabled.
An error should be thrown here.

closed/autoconf/custom-hook.m4
keithc-ca: `configure` allows `--enable-crac-support=xxx`, but the only values for `xxx` that are acceptable are `yes`, `no` and the empty string. This is the same feedback offered for other `--enable`/`--disable` options and the same as before this change.

closed/autoconf/custom-hook.m4
keithc-ca: Just because the default is `no`, doesn't mean it's not supported. I think this will simplify things for someone working towards adding support on a platform where it is not enabled by default.

keithc-ca: It is my understanding that this is what you expected; see https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/800#issuecomment-2150302245.

closed/autoconf/custom-hook.m4
```diff
@@ -340,57 +339,86 @@ AC_DEFUN([OPENJ9_PLATFORM_EXTRACT_VARS_FROM_CPU],
...
-  OPENJ9_ENABLE_CRAC_SUPPORT=false
+  AC_ARG_ENABLE([crac-support], [AS_HELP_STRING([--enable-crac-support], [enable CRaC support @<:@platform dependent@:>@])])
+  AC_ARG_ENABLE([criu-support], [AS_HELP_STRING([--enable-criu-support], [enable CRIU support @<:@platform dependent@:>@])])
+
+  # Complain about explicitly requested, but illegal combinations.
+  if test "x$enable_crac_support" = xyes && test "x$enable_criu_support" = xno ; then
+    AC_MSG_ERROR([--enable-crac-support requires CRIU support])
+  fi
+
+  # Compute platform-specic defaults.
```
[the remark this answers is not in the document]
keithc-ca: Fixed in 226278ea6a3a4d2c6f27b3148c7043d7eb9a6eb1.

closed/autoconf/custom-hook.m4
JasonFengJ9: I thought it is consistent with earlier `origin_criu` assignments.

closed/autoconf/custom-hook.m4
JasonFengJ9: The script in question will enable `CRIU` implicitly when a user attempts to enable `CRaC` at a platform where `CRIU` is disabled by default.
Apparently, the user is working on `CRaC`, if the intention is to add `CRIU` support on that platform at the same time, I feel `CRIU` is better to be enabled explicitly, otherwise the compilation is going to fail.
In addition, I don't see any immediate possibility to add CRIU support in those default `no` platforms such as Windows, Mac, zOS, and AIX.

> It is my understanding that this is what you expected; see https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/800#issuecomment-2150302245.

Sorry I have to change it to:
* `--enable-crac-support` enabled `CRaC` where `CRIU` is enabled by default, otherwise an error will be thrown

keithc-ca: Again, a _default_ should not be equated with _support_. Before this change, you could use the options an all combinations, including one that makes no sense (given CRaC depends upon CRIU), or on platforms where it was unlikely to work. The aim of this change was to detect that one non-sensical choice early, saving time for a developer. I (and Tobi) agreed to your suggestions to allow `--enable-crac-support` to imply `--enable-criu-support`, and to have `--disable-criu-support` to imply `--disable-crac-support` making those outcomes easier to obtain.

I do not think this is the time (I'm not yet convinced it will ever be) to start introducing higher-level checking of configuration choices like you're now proposing. If you disagree, I suggest that is well outside the scope of this change.

JasonFengJ9: The configuration scenario in question is in the development stage, `origin_criu="implicitly enabled"` will show a hint of why a compilation occurs. 
I won't block the current change.

JasonFengJ9:
jenkins compile amac,alinux64 jdknext

JasonFengJ9 (approved):

JasonFengJ9:
@pshipton pls review/merge

pshipton (approved):

pshipton:
jenkins compile amac jdknext

pshipton:
Pls create a PR for jdk23.

PR #807, opened 2024-06-21
Postpone hash check of security profiles

KostasTsiounis:
The check for the hash value of loaded security profiles is postponed until the provider list has been populated.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -76,6 +76,8 @@ public final class RestrictedSecurity {
 
     private static String userSecurityID;
 
+    private static ProfileParser profileParser;
+
     private static RestrictedSecurityProperties restricts;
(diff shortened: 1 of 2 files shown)


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -169,6 +171,13 @@ private RestrictedSecurity() {
         super();
     }
 
+    public static void checkHashValues() {
```
keithc-ca, line 174: Please add javadoc for this method.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Done.

keithc-ca:

keithc-ca (approved):

PR #813, opened 2024-07-11
Add fully-qualified class names of jar verification providers

KostasTsiounis:
Additional entries, pertaining to the fully-qualified class names of the providers, are added to the list utilized to populate the `ProviderList` that is in turn used for jar verification.

This PR is related to https://github.com/eclipse-openj9/openj9/issues/19825.

### DIFF
--- a/src/java.base/share/classes/sun/security/jca/Providers.java
+++ b/src/java.base/share/classes/sun/security/jca/Providers.java
@@ -100,6 +100,20 @@ public class Providers {
         "SunJCE",
     };
 
+    // Hardcoded fully-qualified class names of providers to use for JAR
+    // verification when RestrictedSecurity is enabled (similar to
+    // jarVerificationProviders array).
+    //
+    // MUST NOT be on the bootclasspath and not in signed JAR files.
+    private static final String[] restrictedJarVerificationProviders = {
+        "sun.security.provider.Sun",
+        "sun.security.rsa.SunRsaSign",
+        // Note: when SunEC is in a signed JAR file, it's not signed
+        // by EC algorithms. So it's still safe to be listed here.
+        "sun.security.ec.SunEC",
+        "com.sun.crypto.provider.SunJCE",
+    };
+
     // Return Sun provider.
     // This method should only be called by
     // sun.security.util.ManifestEntryVerifier and java.security.SecureRandom.
@@ -115,7 +129,10 @@ public class Providers {
      */
     public static Object startJarVerification() {
         ProviderList currentList = getProviderList();
-        ProviderList jarList = currentList.getJarList(jarVerificationProviders);
+        ProviderList jarList = currentList.getJarList(
+                RestrictedSecurity.isEnabled()
+                        ? restrictedJarVerificationProviders
+                        : jarVerificationProviders);
         if (jarList.getProvider("SUN") == null) {
             // add backup provider
             Provider p;


src/java.base/share/classes/sun/security/jca/Providers.java
```diff
@@ -90,14 +90,18 @@ private Providers() {
...
-        "SunRsaSign",
+        "SUN", "sun.security.provider.Sun",
+        "SunRsaSign", "sun.security.rsa.SunRsaSign",
         // Note: when SunEC is in a signed JAR file, it's not signed
         // by EC algorithms. So it's still safe to be listed here.
-        "SunEC",
-        "SunJCE",
+        "SunEC", "sun.security.ec.SunEC",
+        "SunJCE", "com.sun.crypto.provider.SunJCE",
     };
```
keithc-ca, line 105: To avoid potentially loading providers twice, perhaps we should check `RestrictedSecurity.isEnabled()` to decide whether to use a list of fully-qualified names instead of the normal list?

KostasTsiounis: Changed to have two separate arrays and a check to decide which one to use.

src/java.base/share/classes/sun/security/jca/Providers.java
```diff
@@ -100,6 +100,20 @@ private Providers() {
         "SunJCE",
     };
 
+    // Hardcoded fully qualified names of providers to use for JAR
```
keithc-ca, line 103: This should be hyphenated: "fully-qualified class names".

src/java.base/share/classes/sun/security/jca/Providers.java
```diff
@@ -100,6 +100,20 @@ private Providers() {
...
+    // verification when RestrictedSecurity is enabled.
+    // (Similar to jarVerificationProviders array)
```
keithc-ca, line 105: Missing period, suggest:
```
    // verification when RestrictedSecurity is enabled (similar to
    // jarVerificationProviders array).
```

src/java.base/share/classes/sun/security/jca/Providers.java
```diff
@@ -115,7 +129,13 @@ public static Provider getSunProvider() {
...
     public static Object startJarVerification() {
         ProviderList currentList = getProviderList();
-        ProviderList jarList = currentList.getJarList(jarVerificationProviders);
+        ProviderList jarList;
+        if (RestrictedSecurity.isEnabled()) {
+            jarList = currentList.getJarList(jarVerificationProvidersRS);
+        } else {
+            jarList = currentList.getJarList(jarVerificationProviders);
+        }
+
```
keithc-ca, line 138: Suggest either capturing the argument to `getJarList()` in a local, or simply making it conditional
```java
        ProviderList jarList = currentList.getJarList(
                RestrictedSecurity.isEnabled()
                        ? restrictedJarVerificationProviders
                        : jarVerificationProviders);
```
We don't need blank line 138.

src/java.base/share/classes/sun/security/jca/Providers.java
```diff
@@ -100,6 +100,20 @@ private Providers() {
         "SunJCE",
     };
 
+    // Hardcoded fully qualified names of providers to use for JAR
+    // verification when RestrictedSecurity is enabled.
+    // (Similar to jarVerificationProviders array)
+    //
+    // MUST NOT be on the bootclasspath and not in signed JAR files.
+    private static final String[] jarVerificationProvidersRS = {
```
keithc-ca, line 108: Suggest a better name is `restrictedJarVerificationProviders` (rather than abbreviate in `RS`).

keithc-ca (changes requested):

keithc-ca:
The message in the first commit is missing "is" in "that _is_ utilized for Jar verification".

src/java.base/share/classes/sun/security/jca/Providers.java
KostasTsiounis: Changed.

src/java.base/share/classes/sun/security/jca/Providers.java
KostasTsiounis: Changed to suggestion.

src/java.base/share/classes/sun/security/jca/Providers.java
KostasTsiounis: Changed to suggestion.

src/java.base/share/classes/sun/security/jca/Providers.java
KostasTsiounis: Changed to suggestion.

KostasTsiounis:
> The message in the first commit is missing "is" in "that _is_ utilized for Jar verification".

I fixed the message of the first commit and squashed.

src/java.base/share/classes/sun/security/jca/Providers.java
```diff
@@ -100,6 +100,20 @@ private Providers() {
         "SunJCE",
     };
 
+    // Hardcoded fully-qualified names of providers to use for JAR
```
keithc-ca, line 103: These are "class" names: The comment should say so.

KostasTsiounis: Updated to indicate that.

keithc-ca:

keithc-ca (approved):

PR #824, opened 2024-07-31
Avoid rebuilding native code headers for OpenJCEPlus

KostasTsiounis:
Some `OpenJCEPlus` headers are already compiled as part of a previous target, so there is no reason to recompile them. Instead an additional variable is passed as part of the make command to indicate that this step can be skipped.

### DIFF
--- a/closed/make/modules/openjceplus/Lib.gmk
+++ b/closed/make/modules/openjceplus/Lib.gmk
@@ -23,13 +23,14 @@ include LibCommon.gmk
 ifeq (true,$(BUILD_OPENJCEPLUS))
 
 # Identify the desired JGSKIT target platform.
-OPENJCEPLUS_BOOT_JDK := $(BOOT_JDK)
+EXPORT_COMPILER_ENV_VARS :=
+OPENJCEPLUS_JDK := $(JDK_OUTPUTDIR)
 OPENJCEPLUS_GSKIT_HOME := $(OPENJCEPLUS_TOPDIR)/ock/jgsk_sdk
+OPENJCEPLUS_HEADER_FILES := $(SUPPORT_OUTPUTDIR)/headers/openjceplus
 OPENJCEPLUS_JCE_CLASSPATH := $(JDK_OUTPUTDIR)/modules/openjceplus:$(JDK_OUTPUTDIR)/modules/java.base
 OPENJCEPLUS_JGSKIT_MAKE := jgskit.mak
 OPENJCEPLUS_JGSKIT_MAKE_PATH := $(OPENJCEPLUS_TOPDIR)/src/main/native
 OPENJCEPLUS_JGSKIT_PLATFORM :=
-OPENJCEPLUS_VS_LIB :=
 
 ifeq ($(call isTargetOs, aix), true)
   OPENJCEPLUS_JGSKIT_PLATFORM := ppc-aix64
@@ -43,12 +44,13 @@ else ifeq ($(call isTargetOs, linux), true)
   endif
 else ifeq ($(call isTargetOs, windows), true)
   ifeq ($(call isTargetCpu, x86_64), true)
-    OPENJCEPLUS_BOOT_JDK := $(call MixedPath,$(OPENJCEPLUS_BOOT_JDK))
+    EXPORT_COMPILER_ENV_VARS := LIB='$(OPENJ9_VS_LIB)'
+    OPENJCEPLUS_JDK := $(call MixedPath,$(OPENJCEPLUS_JDK))
     OPENJCEPLUS_GSKIT_HOME := $(call MixedPath,$(OPENJCEPLUS_GSKIT_HOME))
-    OPENJCEPLUS_JCE_CLASSPATH := "$(call MixedPath,$(JDK_OUTPUTDIR)/modules/openjceplus)\;$(call MixedPath,$(JDK_OUTPUTDIR)/modules/java.base)"
+    OPENJCEPLUS_HEADER_FILES := $(call MixedPath,$(OPENJCEPLUS_HEADER_FILES))
+    OPENJCEPLUS_JCE_CLASSPATH := "$(call MixedPath,$(JDK_OUTPUTDIR)/modules/openjceplus);$(call MixedPath,$(JDK_OUTPUTDIR)/modules/java.base)"
     OPENJCEPLUS_JGSKIT_MAKE := jgskit.win64.mak
     OPENJCEPLUS_JGSKIT_PLATFORM := win64
-    OPENJCEPLUS_VS_LIB := LIB='$(OPENJ9_VS_LIB)'
   endif
 endif
 
@@ -60,13 +62,17 @@ endif # OPENJCEPLUS_JGSKIT_PLATFORM
 
 compile-libs :
 	@$(ECHO) Compiling OpenJCEPlus native code
-	export \
-			$(OPENJCEPLUS_VS_LIB) \
+	$(EXPORT_COMPILER_ENV_VARS) \
+		$(MAKE) \
+			-C $(OPENJCEPLUS_JGSKIT_MAKE_PATH) \
+			-f $(OPENJCEPLUS_JGSKIT_MAKE) \
+			EXTERNAL_HEADERS=true \
 			GSKIT_HOME=$(OPENJCEPLUS_GSKIT_HOME) \
-			JAVA_HOME=$(OPENJCEPLUS_BOOT_JDK) \
+			JAVA_HOME=$(OPENJCEPLUS_JDK) \
 			JCE_CLASSPATH=$(OPENJCEPLUS_JCE_CLASSPATH) \
+			OPENJCEPLUS_HEADER_FILES=$(OPENJCEPLUS_HEADER_FILES) \
 			PLATFORM=$(OPENJCEPLUS_JGSKIT_PLATFORM) \
-		&& $(MAKE) -j1 -C $(OPENJCEPLUS_JGSKIT_MAKE_PATH) -f $(OPENJCEPLUS_JGSKIT_MAKE) all
+			all
 	@$(ECHO) OpenJCEplus compile complete
 
 TARGETS += compile-libs


closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -23,7 +23,7 @@ include LibCommon.gmk
 ifeq (true,$(BUILD_OPENJCEPLUS))
 
 # Identify the desired JGSKIT target platform.
-OPENJCEPLUS_BOOT_JDK := $(BOOT_JDK)
+OPENJCEPLUS_BOOT_JDK := $(JDK_OUTPUTDIR)
```
keithc-ca, line 26: This can't be expected to work. The JDK in `$(JDK_OUTPUTDIR)` isn't necessarily usable yet.

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -64,6 +63,7 @@ compile-libs :
 			JAVA_HOME=$(OPENJCEPLUS_BOOT_JDK) \
 			JCE_CLASSPATH=$(OPENJCEPLUS_JCE_CLASSPATH) \
 			PLATFORM=$(OPENJCEPLUS_JGSKIT_PLATFORM) \
+      OPENJCEPLUS_HEADER_FILES=$(JDK_OUTPUTDIR)/support/headers \
```
keithc-ca, line 66: I expect you want `$(JDK_OUTPUTDIR)/support/headers/openjceplus`.

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -43,8 +44,9 @@ else ifeq ($(call isTargetOs, linux), true)
   endif
 else ifeq ($(call isTargetOs, windows), true)
   ifeq ($(call isTargetCpu, x86_64), true)
-    OPENJCEPLUS_BOOT_JDK := $(call MixedPath,$(OPENJCEPLUS_BOOT_JDK))
+    OPENJCEPLUS_JDK := $(call MixedPath,$(OPENJCEPLUS_JDK))
     OPENJCEPLUS_GSKIT_HOME := $(call MixedPath,$(OPENJCEPLUS_GSKIT_HOME))
+    OPENJCEPLUS_HEADER_FILES := $(call MixedPath,$(SUPPORT_OUTPUTDIR)/headers/openjceplus)
```
keithc-ca, line 49: This should use `OPENJCEPLUS_HEADER_FILES` defined above:
```
    OPENJCEPLUS_HEADER_FILES := $(call MixedPath,$(OPENJCEPLUS_HEADER_FILES))
```

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -43,8 +44,9 @@ else ifeq ($(call isTargetOs, linux), true)
...
     OPENJCEPLUS_JCE_CLASSPATH := "$(call MixedPath,$(JDK_OUTPUTDIR)/modules/openjceplus)\;$(call MixedPath,$(JDK_OUTPUTDIR)/modules/java.base)"
```
keithc-ca, line 50: I don't think `;` needs to be escaped here (it's within quotes).

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -63,10 +65,14 @@ compile-libs :
 	export \
 			$(OPENJCEPLUS_VS_LIB) \
 			GSKIT_HOME=$(OPENJCEPLUS_GSKIT_HOME) \
-			JAVA_HOME=$(OPENJCEPLUS_BOOT_JDK) \
+			JAVA_HOME=$(OPENJCEPLUS_JDK) \
 			JCE_CLASSPATH=$(OPENJCEPLUS_JCE_CLASSPATH) \
 			PLATFORM=$(OPENJCEPLUS_JGSKIT_PLATFORM) \
```
keithc-ca, line 70: Suggest you list these alphabetically:
```
			GSKIT_HOME=$(OPENJCEPLUS_GSKIT_HOME) \
			JAVA_HOME=$(OPENJCEPLUS_JDK) \
			JCE_CLASSPATH=$(OPENJCEPLUS_JCE_CLASSPATH) \
			$(OPENJCEPLUS_VS_LIB) \
			PLATFORM=$(OPENJCEPLUS_JGSKIT_PLATFORM) \
```
At this stage, the name `JAVA_HOME` is misleading (it isn't yet a usable JDK).

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -63,10 +65,14 @@ compile-libs :
...
-		&& $(MAKE) -j1 -C $(OPENJCEPLUS_JGSKIT_MAKE_PATH) -f $(OPENJCEPLUS_JGSKIT_MAKE) all
+		&& $(MAKE) \
```
keithc-ca, line 71: Please confirm the intentional removal of `-j1` (the OpenJCEPlus makefiles are ready for parallel execution)?

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -63,10 +65,14 @@ compile-libs :
...
+			-C $(OPENJCEPLUS_JGSKIT_MAKE_PATH) \
+			-f $(OPENJCEPLUS_JGSKIT_MAKE) \
+			OPENJCEPLUS_HEADER_FILES=$(OPENJCEPLUS_HEADER_FILES) \
+			noheaders
```
keithc-ca, line 75: While `noheaders` looks like it will work, I don't think the `all` target is reliable. Instead of a separate target, I suggest the OpenJCEPlus makefiles should conditionally handle generation of header files: this command would end with something like `OPENJCEPLUS_BUILD_HEADERS=false all` (the default would assume `OPENJCEPLUS_BUILD_HEADERS=true` which would use `javac` to ensure the header files exist).

keithc-ca (changes requested):

closed/make/modules/openjceplus/Lib.gmk
KostasTsiounis: I need to name it `JAVA_HOME` because that's what the makefile expects, given that in standalone builds we would use the java present in the system.

closed/make/modules/openjceplus/Lib.gmk
KostasTsiounis: Yes, I updated the prerequisites for the targets to allow for parallel execution.

closed/make/modules/openjceplus/Lib.gmk
KostasTsiounis: The `all` target would only fail in this particular scenario where someone is trying to bundle. I'm not sure a use case where someone wouldn't have to compile the headers, or try to compile them with a different java version compared to the rest of the classes, exists. I'm worried that we will be adding conditionals for a very particular scenario that we only encounter when bundling with `Semeru`. I think separate targets is more straightforward and doesn't unnecessarily complicate the makefile.

closed/make/modules/openjceplus/Lib.gmk
KostasTsiounis: Fixed.

closed/make/modules/openjceplus/Lib.gmk
KostasTsiounis: Removed backslash.

closed/make/modules/openjceplus/Lib.gmk
KostasTsiounis: Sorted alphabetically.

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -61,12 +63,17 @@ endif # OPENJCEPLUS_JGSKIT_PLATFORM
 compile-libs :
 	@$(ECHO) Compiling OpenJCEPlus native code
 	export \
-			$(OPENJCEPLUS_VS_LIB) \
 			GSKIT_HOME=$(OPENJCEPLUS_GSKIT_HOME) \
-			JAVA_HOME=$(OPENJCEPLUS_BOOT_JDK) \
+			JAVA_HOME=$(OPENJCEPLUS_JDK) \
 			JCE_CLASSPATH=$(OPENJCEPLUS_JCE_CLASSPATH) \
+			$(OPENJCEPLUS_VS_LIB) \
 			PLATFORM=$(OPENJCEPLUS_JGSKIT_PLATFORM) \
```
keithc-ca, line 70: Suggest rather than exporting these variables, they should be appear on the `make` command line (like `OPENJCEPLUS_HEADER_FILES`).

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -61,12 +63,17 @@ endif # OPENJCEPLUS_JGSKIT_PLATFORM
...
-		&& $(MAKE) -j1 -C $(OPENJCEPLUS_JGSKIT_MAKE_PATH) -f $(OPENJCEPLUS_JGSKIT_MAKE) all
+		&& $(MAKE) \
+			-C $(OPENJCEPLUS_JGSKIT_MAKE_PATH) \
+			-f $(OPENJCEPLUS_JGSKIT_MAKE) \
+			OPENJCEPLUS_HEADER_FILES=$(OPENJCEPLUS_HEADER_FILES) \
+      EXTERNAL_HEADERS=true \
```
keithc-ca, line 75: This should be indented the same as line 74 (with tabs instead of spaces).

keithc-ca (changes requested):

closed/make/modules/openjceplus/Lib.gmk
KostasTsiounis: Moved required variables to `make` command instead of exporting.

closed/make/modules/openjceplus/Lib.gmk
KostasTsiounis: Fixed.

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -60,13 +62,18 @@ endif # OPENJCEPLUS_JGSKIT_PLATFORM
...
 compile-libs :
 	@$(ECHO) Compiling OpenJCEPlus native code
-	export \
-			$(OPENJCEPLUS_VS_LIB) \
-			GSKIT_HOME=$(OPENJCEPLUS_GSKIT_HOME) \
-			JAVA_HOME=$(OPENJCEPLUS_BOOT_JDK) \
-			JCE_CLASSPATH=$(OPENJCEPLUS_JCE_CLASSPATH) \
-			PLATFORM=$(OPENJCEPLUS_JGSKIT_PLATFORM) \
-		&& $(MAKE) -j1 -C $(OPENJCEPLUS_JGSKIT_MAKE_PATH) -f $(OPENJCEPLUS_JGSKIT_MAKE) all
+  $(EXPORT_COMPILER_ENV_VARS) \
```
keithc-ca, line 65: This line must begin with a tab (else it isn't part of the script for `compile-libs`).

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -60,13 +62,18 @@ endif # OPENJCEPLUS_JGSKIT_PLATFORM
...
+	$(MAKE) \
+		-C $(OPENJCEPLUS_JGSKIT_MAKE_PATH) \
+		-f $(OPENJCEPLUS_JGSKIT_MAKE) \
+		EXTERNAL_HEADERS=true \
+		GSKIT_HOME=$(OPENJCEPLUS_GSKIT_HOME) \
+		JAVA_HOME=$(OPENJCEPLUS_JDK) \
+		JCE_CLASSPATH=$(OPENJCEPLUS_JCE_CLASSPATH) \
+		OPENJCEPLUS_HEADER_FILES=$(OPENJCEPLUS_HEADER_FILES) \
+		$(OPENJCEPLUS_VS_LIB) \
```
keithc-ca, line 74: `OPENJCEPLUS_VS_LIB` is not defined so this line isn't helpful.

keithc-ca (changes requested):

closed/make/modules/openjceplus/Lib.gmk
KostasTsiounis: Sorry. The editor added spaces instead of tab and I missed it. Fixed now.

closed/make/modules/openjceplus/Lib.gmk
KostasTsiounis: Removed.

closed/make/modules/openjceplus/Lib.gmk
```diff
@@ -60,13 +62,17 @@ endif # OPENJCEPLUS_JGSKIT_PLATFORM
...
+	$(MAKE) \
+		-C $(OPENJCEPLUS_JGSKIT_MAKE_PATH) \
+		-f $(OPENJCEPLUS_JGSKIT_MAKE) \
+		EXTERNAL_HEADERS=true \
+		GSKIT_HOME=$(OPENJCEPLUS_GSKIT_HOME) \
+		JAVA_HOME=$(OPENJCEPLUS_JDK) \
+		JCE_CLASSPATH=$(OPENJCEPLUS_JCE_CLASSPATH) \
+		OPENJCEPLUS_HEADER_FILES=$(OPENJCEPLUS_HEADER_FILES) \
+		PLATFORM=$(OPENJCEPLUS_JGSKIT_PLATFORM) \
+		all
```
keithc-ca, line 75: Suggest lines 66-75 be indented further to clarify that they are all part of the command that begins on line 65.

KostasTsiounis: Done.

KostasTsiounis:
Squashed.

keithc-ca:
Jenkins compile alinux,win jdknext

PR #826, opened 2024-08-12
Restore OpenJCEPlus certified binaries in JRE

jasonkatonica:
When creating a JRE using the target `legacy-jre-image` the certified binaries in the openjceplus module are changed due to the use of the `--strip-debug` argument used with by the jlink command to create the JRE.

This update restores the certified files to their original values that are present in the `JDK` image. This restoration only occurs if the `openjceplus` certified binaries are already present after the `jlink` operation is executed. This update is required since any changes to the FIPS libraries will cause a failure when loading the dll due to a failure during the self verification process done by the library.


make/Images.gmk
```diff
@@ -117,6 +121,7 @@ $(eval $(call SetupExecute, jlink_jre, \
     OUTPUT_DIR := $(JRE_IMAGE_DIR), \
     SUPPORT_DIR := $(JRE_IMAGE_SUPPORT_DIR), \
     PRE_COMMAND := $(RM) -r $(JRE_IMAGE_DIR), \
+    POST_COMMAND := [ -d $(JRE_IMAGE_DIR)/lib/C ] && $(CP) -rfv $(IMAGES_OUTPUTDIR)/../support/modules_libs/openjceplus/C $(JRE_IMAGE_DIR)/lib, \
```
keithc-ca, line 124: First, I don't think it's reasonable that openjceplus content is sensitive to the effects of `--strip-debug`, and this doesn't address the problem that will continue to exist with users using `jlink --strip-debug` to create their own runtimes. Having said that, even as an _interim fix_, this `POST_COMMAND` needs to change:
* in builds without openjceplus it will just fail (it should do nothing)
* in builds with openjceplus, the action should not be predicated on the existence of the `lib/C` directory (it should fail if it does not exist)
* the command for Windows will probably need to be different
* it should use `$(SUPPORT_OUTPUTDIR)` instead of `$(IMAGES_OUTPUTDIR)/../support`
* the `-v` option should be omitted from the copy command

I suggest that before the call to `SetupExecute` we want something like
```
ifeq ($(BUILD_OPENJCEPLUS), false)
  LEGACY_JRE_POST_COMMAND :=
else ifeq ($(call isTargetOs, windows), true)
  LEGACY_JRE_POST_COMMAND := $(CP) -rf $(SUPPORT_OUTPUTDIR)/modules_cmds/openjceplus/C $(JRE_IMAGE_DIR)/bin
else
  LEGACY_JRE_POST_COMMAND := $(CP) -rf $(SUPPORT_OUTPUTDIR)/modules_libs/openjceplus/C $(JRE_IMAGE_DIR)/lib
endif
```
then the `POST_COMMAND` parameter becomes
```
    POST_COMMAND := $(LEGACY_JRE_POST_COMMAND), \
```

keithc-ca (changes requested):

make/Images.gmk
jasonkatonica: Yes agreed ideally `openjceplus` FIPS binaries are not sensitive to the effects of `--strip-debug` or the `strip` command. Users performing `jlink` themselves could run into this same issue. 

This being said we are unsure how to make any recommendations to the OCKC team such that the binary is not sensitive to `strip`. Currently only a single byte is changed by this command so there are no obvious causes here to me ( such as compiling with debug symbols ) that are causing the file to be sensitive to `strip`. Additionally at this point in time the binary has been approved by NIST and is not subject to change for the forseeable future. As an interim fix we certainly need to document in the FIPS related documenation that this argument sent to the jlink command is not supported if somebody were to run this on their own.

I did make updates as recommended in the latest version of the code.
- The suggested variable `LEGACY_JRE_POST_COMMAND` is set only when openjceplus bundling is set. 
- The directory for windows indeed is the bin directory instead of lib. This was modified as suggested.
- Variable `$(SUPPORT_OUTPUTDIR)` was used.
- The verbose `-v` flag was removed from the copy.

jasonkatonica: Hi @johnpeck-us-ibm can you please review the above issue we are trying to fix? We would like to have the ability for the FIPS module to be unaffected by the `jlink` command using the `--strip-debug` argument. This seems to be observed to behaving similar to running a `strip` command within Linux. Currently running this command causes a few bytes to change, which then causes the self verification process of the module to fail upon loading the FIPS certified library.

johnpeck-us-ibm: The FIPS code is tamper resistant. 

Any modification to the library will result in the Selftests that FIPS 140 requires to fail. This in turn causes the initialization of the Java Provider to fail. This is done so that the code can not be modified in any way. The ICC libraries already have all the symbols removed. So, the use of --strip-debug is not needed. Also, since the jlink or strip commands rewrites the ICC library the tamper resistance of the library detects this and therefore causes the initialization of library to fail.

Any modification to the library SHOULD cause this failure. Since, this tool is trying to modify the library the tamper resistance is working as designed.

We  CANNOT try and allow ANY kind of modification of the FIPS libraries. This would break module security and mean that the module is not secure.

make/Images.gmk
```diff
@@ -110,13 +114,22 @@ $(eval $(call SetupExecute, jlink_jdk, \
 
 JLINK_JDK_TARGETS := $(jlink_jdk)
 
+ifeq ($(BUILD_OPENJCEPLUS), false)
+  LEGACY_JRE_POST_COMMAND :=
+else ifeq ($(call isTargetOs, windows), true)
+  LEGACY_JRE_POST_COMMAND := $(CP) -rf $(SUPPORT_OUTPUTDIR)/modules_cmds/openjceplus/C $(JRE_IMAGE_DIR)/bin
+else
+  LEGACY_JRE_POST_COMMAND := $(CP) -rf $(SUPPORT_OUTPUTDIR)/modules_libs/openjceplus/C $(JRE_IMAGE_DIR)/lib
+endif
```
keithc-ca, line 123: After discovering that `StripNativeDebugSymbolsPlugin` only applies on Linux, this can be simplified:
```
ifeq ($(call And, $(BUILD_OPENJCEPLUS) $(call isTargetOs, linux)), true)
  LEGACY_JRE_POST_COMMAND := $(CP) -rf $(call FindLibDirForModule, openjceplus)/C $(JRE_IMAGE_DIR)/lib
else
  LEGACY_JRE_POST_COMMAND :=
endif
```
but perhaps tweaking `StripNativeDebugSymbolsPlugin` to avoid a need for this is a better path forward.

jasonkatonica: OK will pursue a fix by changing the code in `StripNativeDebugSymbolsPlugin` instead of the makefiles themselves. Advantage certainly being that all users of `jlink` will avoid the stripping of the library.

jasonkatonica: Updated the code according to this conversation.

src/jdk.jlink/linux/classes/jdk/tools/jlink/internal/plugins/StripNativeDebugSymbolsPlugin.java
```diff
@@ -131,6 +136,33 @@ public ResourcePool transform(ResourcePool in, ResourcePoolBuilder out) {
...
+            if (resource.moduleName().equals("openjceplus") &&
+                resource.path().endsWith(SHARED_LIBS_EXT) &&
+                resource.path().contains("C/icc")) {
+                return false;
+            } else {
+                return true;
+            }
+        } else {
+            return false;
+        }
```
keithc-ca, line 163: This method should be `static`. Suggest a `switch` statement makes this clearer and avoids redundant tests:
```java
        switch (resource.type()) {
        case NATIVE_CMD:
            return true;
        case NATIVE_LIB:
            String path = resource.path();
            if (path.endsWith(SHARED_LIBS_EXT)) {
                if (!(resource.moduleName().equals("openjceplus") && path.contains("/C/icc/"))) {
                    return true;
                }
            }
            break;
        default:
            break;
        }
        return false;
```
Note:
* this has exactly one call to `type()` and at most one call to `path()`
* the path "contains" test is more specific

src/jdk.jlink/linux/classes/jdk/tools/jlink/internal/plugins/StripNativeDebugSymbolsPlugin.java
```diff
@@ -131,6 +136,33 @@ public ResourcePool transform(ResourcePool in, ResourcePoolBuilder out) {
         return out.build();
     }
 
+    /**
+     * Method to determine if a particular resource should be stripped.
```
keithc-ca, line 140: This javadoc suggests what I think is a better name for the method: `shouldStrip()`.

src/jdk.jlink/linux/classes/jdk/tools/jlink/internal/plugins/StripNativeDebugSymbolsPlugin.java
```diff
@@ -131,6 +136,33 @@ public ResourcePool transform(ResourcePool in, ResourcePoolBuilder out) {
...
+     *
+     * Particular paths are added here to handle libraries within the openjceplus module.
+     * The FIPS certified library located in the C/icc directory is sensitive to
+     * any modifications to the native library. Performing any modifications to the library
+     * in any way, causes the FIPS library to fail to load due to a self verification check made.
+     *
+     * @param resource the resource to examine for stripping eligibility.
+     * @return return true if stripping should be done on a particular resource, false otherwise.
```
keithc-ca, line 148: Please omit the trailing periods from these tags.

keithc-ca (changes requested):

src/jdk.jlink/linux/classes/jdk/tools/jlink/internal/plugins/StripNativeDebugSymbolsPlugin.java
jasonkatonica: Updated / tested as suggested above.

src/jdk.jlink/linux/classes/jdk/tools/jlink/internal/plugins/StripNativeDebugSymbolsPlugin.java
jasonkatonica: Renamed the method to `shouldStrip()`

src/jdk.jlink/linux/classes/jdk/tools/jlink/internal/plugins/StripNativeDebugSymbolsPlugin.java
jasonkatonica: Trailing periods removed from @param and @return phrases.

src/jdk.jlink/linux/classes/jdk/tools/jlink/internal/plugins/StripNativeDebugSymbolsPlugin.java
```diff
@@ -22,6 +22,13 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ */
+
```
keithc-ca, line 31: My preference is that you not add blank lines before the `package` declaration as there were none present before this change. I wasn't going to mention that, but since this should be squashed anyway, please remove lines 25 and 31.

jasonkatonica: I removed line `25` and `31` and then performed a `squash`.

keithc-ca (approved):

keithc-ca:

PR #832, opened 2024-08-26
Add support for EC aliases in SunPKCS11

BobMDu:
In Java 8 with IBMPKCS11Impl, we allowed the usage of `1.2.840.10045.2.1` as an alias for `EC` when using ECKeyFactory. This is support we want to continue to offer with SunPKCS11 and Semeru.

Changing the function from `d()` to [`dA()`](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java#L726) allows [`getAliases()`](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/12ee8ca858929b069c1635152a1fefdf778e9af1/src/java.base/share/classes/sun/security/util/SecurityProviderConstants.java#L81) to be called. which checks the [list of known aliases](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/12ee8ca858929b069c1635152a1fefdf778e9af1/src/java.base/share/classes/sun/security/util/KnownOIDs.java#L281) for EC that contains `1.2.840.10045.2.1`.

Signed-off-by: Bob Du <bob.du@ibm.com>

### DIFF
--- a/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
+++ b/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
@@ -874,7 +874,7 @@ public final class SunPKCS11 extends AuthProvider {
         d(KPG, "DH",            P11KeyPairGenerator,
                 dhAlias,
                 m(CKM_DH_PKCS_KEY_PAIR_GEN));
-        d(KPG, "EC",            P11KeyPairGenerator,
+        dA(KPG, "EC",            P11KeyPairGenerator,
                 m(CKM_EC_KEY_PAIR_GEN));
 
         dA(KG,  "ARCFOUR",       P11KeyGenerator,
@@ -924,7 +924,7 @@ public final class SunPKCS11 extends AuthProvider {
         d(KF, "DH",             P11DHKeyFactory,
                 dhAlias,
                 m(CKM_DH_PKCS_KEY_PAIR_GEN, CKM_DH_PKCS_DERIVE));
-        d(KF, "EC",             P11ECKeyFactory,
+        dA(KF, "EC",             P11ECKeyFactory,
                 m(CKM_EC_KEY_PAIR_GEN, CKM_ECDH1_DERIVE,
                     CKM_ECDSA, CKM_ECDSA_SHA1));
 


src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -924,7 +924,7 @@ private static void register(Descriptor d) {
         d(KF, "DH",             P11DHKeyFactory,
                 dhAlias,
                 m(CKM_DH_PKCS_KEY_PAIR_GEN, CKM_DH_PKCS_DERIVE));
-        d(KF, "EC",             P11ECKeyFactory,
+        dA(KF, "EC",             P11ECKeyFactory,
```
keithc-ca, line 927: This seems like it might just be an oversight from upstream. Are we planning to suggest this change to openjdk?

BobMDu: I wasn't planning on suggesting this. Do you know who I can contact to confirm if this is an oversight?

keithc-ca: I offered a suggestion via a private message.

keithc-ca: I looked at this again more closely: `KnownOIDs.EC` doesn't define any aliases, so this should have no effect.
Do you have a small test program that demonstrates use of the OID `1.2.840.10045.2.1` that works in an older Java version but fails here?

BobMDu: Thanks for the contact suggestion. 

I see the alias listed in [KnownOIDs.java on line 281](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/61b445b6e999dffc6ea425c6d87570a4e259ab17/src/java.base/share/classes/sun/security/util/KnownOIDs.java#L281). 

The alias works with the IBMPKCS11Impl in Java 8 but not with SunPKCS11 in any Semeru release. You can quickly test this by getting an instance of `KeyFactory` with that alias and a configured PKCS11 provider. 

```
import java.security.*;

public class AliasTest {
    public static void main (String[] args) {

        try {

            KeyFactory kf = KeyFactory.getInstance("1.2.840.10045.2.1", "IBMPKCS11Impl-jazzsv1");
//            KeyFactory kf = KeyFactory.getInstance("1.2.840.10045.2.1", "SunPKCS11-jazzsv1");

            System.out.println("Success!");
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```

keithc-ca: This doesn't work for me. A standard jdk has the "SunPKCS11" provider but neither "IBMPKCS11Impl-jazzsv1" nor "SunPKCS11-jazzsv1" providers. I tested with similar code:
```java
public class AliasTest {

    public static void main(String[] args) {
        String[] algorithms = { "EC", "1.2.840.10045.2.1", "OID.1.2.840.10045.2.1" };
        String providerName = (args.length == 0) ? "SunPKCS11" : args[0];
        Provider provider = Security.getProvider(providerName);

        if (provider == null) {
            System.out.format("Can't find '%s' provider.%n", providerName);
            System.exit(1);
        }

        for (String algorithm : algorithms) {
            try {
                KeyFactory factory = KeyFactory.getInstance(algorithm, provider);

                System.out.format("KeyFactory(%s) algorithm: %s from %s%n",
                        algorithm, factory.getAlgorithm(), factory.getProvider().getName());
            } catch (NoSuchAlgorithmException e) {
                System.out.format("No KeyFactory for %s from %s%n", algorithm, provider.getName());
            }
        }
    }
}
```
Run with no arguments it reports:
```
No KeyFactory for EC from SunPKCS11
No KeyFactory for 1.2.840.10045.2.1 from SunPKCS11
No KeyFactory for OID.1.2.840.10045.2.1 from SunPKCS11
```
When run with the argument "SunEC" it reports:
```
KeyFactory(EC) algorithm: EC from SunEC
KeyFactory(1.2.840.10045.2.1) algorithm: 1.2.840.10045.2.1 from SunEC
KeyFactory(OID.1.2.840.10045.2.1) algorithm: OID.1.2.840.10045.2.1 from SunEC
```
Do we need (or want) an implementation other than the one from SunEC?

What about `AlgorithmParameters.getInstance()` and `KeyPairGenerator.getInstance()`? If we support aliases for "EC", shouldn't we support those aliases for all related services provided by SunPKCS11?

BobMDu: > This doesn't work for me. A standard jdk has the "SunPKCS11" provider but neither "IBMPKCS11Impl-jazzsv1" nor "SunPKCS11-jazzsv1" providers. 

SunPKCS11 must be first configured to point to a PKCS#11 library and a token with a config file specified in the `java.security` file. The `-jazzv1` being appended to the provider results from it being configured correctly. I'm not sure about the configuration of your system but we have crypto cards and opencryptoki installed on ours to enable testing. 

> Do we need (or want) an implementation other than the one from SunEC?

SunEC is a different provider and we want to continue supporting this capability specifically with a PKCS#11 provider, since it provides native interfaces allow the use of smart cards and hardware accelerators on Z for cryptography. 

> What about AlgorithmParameters.getInstance() and KeyPairGenerator.getInstance()? If we support aliases for "EC", shouldn't we support those aliases for all related services provided by SunPKCS11?

`AlgorithmParameters` [already checks](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/cbc1dde68e13e2c7eae6169fa075dab253f5dba9/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java#L933) for aliases using `dA()`. 

IBMPKCS11Impl also doesn't support the `1.2.840.10045.2.1` alias for `KeyPairGenerator`. It only does so for `KeyFactory` and `AlgorithmParameters`.

keithc-ca:
I think the OIDs should work for `KeyPairGenerator` regardless of what `IBMPKCS11Impl` does.

BobMDu:
Updated to allow the OID to work for `KeyPairGenerator` as well. Following up on previous comment to suggest this change to openjdk.

keithc-ca:
Ping @BobMDu.

BobMDu:
@keithc-ca Sorry I wasn't sure if we were going to merge this right away. Is it ok to merge first while waiting for possible openjdk integration? If so I will rebase.

keithc-ca:
> Is it ok to merge first while waiting for possible openjdk integration? If so I will rebase.

Yes, the plan was to merge this ahead of an upstream fix. Please rebase.

keithc-ca (approved):
Test failures in sanity.openjdk are known.

PR #833, opened 2024-08-26
Add metadata files needed to build jdk.jfr module

thallium:


src/hotspot/share/jfr/metadata/metadata.xsd
```diff
@@ -0,0 +1,132 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+===========================================================================
+(c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+===========================================================================
+-->
+
```
keithc-ca, line 8: If this file is otherwise identical to upstream, then no IBM copyright should be added. If it is different, please explain those differences and put the IBM copyright notice after the Oracle notice.

keithc-ca (changes requested):
Same comment for both files.

src/hotspot/share/jfr/metadata/metadata.xsd
thallium: Yes these two files are identical to upstream, except the original `metadata.xsd` contains a trailing whitespace. I removed the whitespace. Does that make it "different"?

keithc-ca: Please don't make any changes, not even to trailing whitespace.

tajila:
@thallium update the copywrite ignore list https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/.copyrightignore to include the two new files

tajila:
https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/buildenv/jenkins/jobs/infrastructure/copyrightCheckDir.sh may also need to be updated

thallium:
@tajila Looks like `buildenv/jenkins/jobs/infrastructure/copyrightCheck` is being run and it respects `.copyrightignore` file.

.copyrightignore
```diff
@@ -36,6 +36,8 @@
 /src/linux/doc/
 /src/solaris/doc/
 /test/
+/src/hotspot/share/jfr/metadata/metadata.xml
+/src/hotspot/share/jfr/metadata/metadata.xsd
```
keithc-ca, line 40: Please insert these so the list remains sorted.

keithc-ca (changes requested):
`copyrightCheckDir.sh` is also used (when checking a repository as a whole instead of just files changed in a pull request). Please accommodate the new metadata files in that script as well.

thallium:
@keithc-ca sorry I didn't notice the comment on `copyrightCheckDir.sh` yesterday, can you check if I have made the intended changes?

buildenv/jenkins/jobs/infrastructure/copyrightCheckDir.sh
```diff
@@ -108,6 +108,13 @@ check () {
     *.ser)
       trace "Ignoring $1 because it appears to match case parameter expansion *.ser"
       CHECK=0;;
+    # Ignore JFR metadata file
```
keithc-ca, line 111: I'm not sure this comment provides much value, but if you think it should stay, it should be corrected: there are two metadata files (plural), and the sentence should end with a period.

tajila:
> @pshipton and @tajila: please confirm your agreement that the metadata files can be included here even though their copyright notices don't contain the classpath exception.

Yes, confirming that the metadata files can be included.

pshipton:
Confirmed.

keithc-ca (approved):

tajila:
@thallium Please add backports for JDK23, 21, 17 and 11

pshipton:
Pls fix copyrightCheckDir.sh, the change added here isn't working, the copyright check is failing in every build.
@thallium

https://openj9-jenkins.osuosl.org/job/CopyrightCheck-OpenJDK/1215
```
20:35:21  E012: src/hotspot/share/jfr/metadata/metadata.xml: GPLv2 is present but Classpath exception is missing
20:35:21  Did not find src/hotspot/share/jfr/metadata/metadata.xml in known failures file /home/jenkins/jenkins-agent/workspace/CopyrightCheck-OpenJDK/workdir/ibmruntimes/openj9-openjdk-jdk.git/copyrightCheck.known.failures, treating as error
20:35:21  E012: src/hotspot/share/jfr/metadata/metadata.xsd: GPLv2 is present but Classpath exception is missing
20:35:21  Did not find src/hotspot/share/jfr/metadata/metadata.xsd in known failures file /home/jenkins/jenkins-agent/workspace/CopyrightCheck-OpenJDK/workdir/ibmruntimes/openj9-openjdk-jdk.git/copyrightCheck.known.failures, treating as error
```

thallium:
I think this should fix the issue https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/843

PR #843, opened 2024-09-16
Fix copyrightCheckDir.sh on ignoring JFR metadata files

thallium:


pshipton (approved):

pshipton:
Thanks. I'll merge it and we'll see in the next builds.

PR #844, opened 2024-09-18
Ensure static fields set during signed jar process

jasonkatonica:
When loading a signed jar file that is on the classpath, such as the
bouncy castle signed JCE jar file, it has been observed that the value
of `restrictedJarVerificationProviders` and `jarVerificationProviders`
are set to `null`. This causes a NullPointerException during the loading
process.

This update moves the static declarations of both
`restrictedJarVerificationProviders` and `jarVerificationProviders` to
be prior to the method call `RestrictedSecurity.checkHashValues()` since
this method call needs both of these fields to be initialized to work
correctly.

### DIFF
--- a/src/java.base/share/classes/sun/security/jca/Providers.java
+++ b/src/java.base/share/classes/sun/security/jca/Providers.java
@@ -56,14 +56,6 @@ public class Providers {
     // Note volatile immutable object, so no synchronization needed.
     private static volatile ProviderList providerList;
 
-    static {
-        // set providerList to empty list first in case initialization somehow
-        // triggers a getInstance() call (although that should not happen)
-        providerList = ProviderList.EMPTY;
-        providerList = ProviderList.fromSecurityProperties();
-        RestrictedSecurity.checkHashValues();
-    }
-
     private Providers() {
         // empty
     }
@@ -113,6 +105,14 @@ public class Providers {
         "sun.security.ec.SunEC",
         "com.sun.crypto.provider.SunJCE",
     };
+
+    static {
+        // set providerList to empty list first in case initialization somehow
+        // triggers a getInstance() call (although that should not happen)
+        providerList = ProviderList.EMPTY;
+        providerList = ProviderList.fromSecurityProperties();
+        RestrictedSecurity.checkHashValues();
+    }
 
     // Return Sun provider.
     // This method should only be called by


src/java.base/share/classes/sun/security/jca/Providers.java
```diff
@@ -56,18 +56,6 @@ public class Providers {
...
-        // set providerList to empty list first in case initialization somehow
-        // triggers a getInstance() call (although that should not happen)
-        providerList = ProviderList.EMPTY;
-        providerList = ProviderList.fromSecurityProperties();
-        RestrictedSecurity.checkHashValues();
-    }
-
-    private Providers() {
-        // empty
-    }
```
keithc-ca, line 69: Please leave the constructor where it was.

keithc-ca (changes requested):

src/java.base/share/classes/sun/security/jca/Providers.java
jasonkatonica: Hi Keith. I don't believe that I moved the constructor here, instead i just moved `restrictedJarVerificationProviders` and `jarVerificationProviders` above the static block. The constructor `Providers()` is right after the `static {}` both before and after the update. Perhaps its the way that github is lining up the dif here and showing movement of the static block and constructor instead of showing movement of the `restrictedJarVerificationProviders` and `jarVerificationProviders` values? In any case I can move the `Providers()` constructor above both
`restrictedJarVerificationProviders` and `jarVerificationProviders` also if that is what we want to do?

keithc-ca: In the [latest version of this file from upstream](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/5dd840a87ed8d325e0a690e901bc2c385438b8eb/src/java.base/share/classes/sun/security/jca/Providers.java), the declaration of `providerList` is followed by the `static` block (that needs to move), then the constructor. After this change, the constructor should immediately follow the declaration of `providerList`.

jasonkatonica: OK, moved constructor to after `providerList` declaration using commit [de10fad](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/844/commits/de10fad3038f8f28995a263e9be8ccfe97cfc7df). The dif now represents moving the static block down below the `restrictedJarVerificationProviders` and `jarVerificationProviders` values instead of the reverse.

keithc-ca (commented):
Thank you: This more clearly shows the movement of the static initializer block. Please squash.

jasonkatonica:
Code is now squashed into a single commit.

keithc-ca:

keithc-ca (approved):

keithc-ca:
Back-ports:
* https://github.com/ibmruntimes/openj9-openjdk-jdk23/pull/45
* https://github.com/ibmruntimes/openj9-openjdk-jdk21/pull/198
* https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/383
* https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/809
* https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/763

PR #846, opened 2024-09-23
Check RestrictedSecurity profile for errors

taoliult:
This commit checks the misspelled properties and provider order numbers in RestrictedSecurity mode. If there is a misspelled property or the provider order numbers are not consecutive. The RestrictedSecurity mode initializion will stop and throw errors.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -35,10 +35,10 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Deque;
+import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -994,6 +994,9 @@ public final class RestrictedSecurity {
         // The java.security properties.
         private final Properties securityProps;
 
+        private final Set<String> profileCheckPropertyNames;
+        private final Set<String> profileCheckProviderNames;
+
         /**
          *
          * @param id    the restricted security custom profile ID
@@ -1016,8 +1019,13 @@ public final class RestrictedSecurity {
 
             parsedProfiles = new HashSet<>();
 
+            profileCheckPropertyNames = new HashSet<>();
+            profileCheckProviderNames = new HashSet<>();
+
             // Initialize the properties.
             init(profileID);
+
+            checkProfileCheck(profileID);
         }
 
         private RestrictedSecurityProperties getProperties() {
@@ -1040,11 +1048,16 @@ public final class RestrictedSecurity {
                 printStackTraceAndExit(profileID + " has already been parsed. Potential infinite recursion.");
             }
 
-            String potentialExtendsProfileID = parseProperty(securityProps.getProperty(profileID + ".extends"));
+            loadProfileCheck(profileID);
+
+            String profileExtends = profileID + ".extends";
+            String potentialExtendsProfileID = parseProperty(securityProps.getProperty(profileExtends));
             if (potentialExtendsProfileID != null) { // If profile extends another profile.
                 if (debug != null) {
                     debug.println("\t'" + profileID + "' extends '" + potentialExtendsProfileID + "'.");
                 }
+
+                profileCheckPropertyNames.remove(profileExtends);
 
                 // Check if extended profile exists.
                 String extendsProfileID = null;
@@ -1098,6 +1111,7 @@ public final class RestrictedSecurity {
                         // Save info to be hashed and expected result to be checked later.
                         profilesHashes.put(profileID, hashValue);
                         profilesInfo.put(profileID, allInfo);
+                        profileCheckPropertyNames.remove(hashProperty);
                     } else if (!isFIPS1402Profile(profileID)) {
                         // A hash is mandatory, but not for older 140-2 profiles.
                         printStackTraceAndExit(profileID + " is a base profile, so a hash value is mandatory.");
@@ -1134,6 +1148,7 @@ public final class RestrictedSecurity {
                     // Save info to be hashed and expected result to be checked later.
                     profilesHashes.put(profileID, hashValue);
                     profilesInfo.put(profileID, allInfo);
+                    profileCheckPropertyNames.remove(hashProperty);
                 }
             } catch (Exception e) {
                 if (debug != null) {
@@ -1254,6 +1269,7 @@ public final class RestrictedSecurity {
                 allInfo.add(property + "=" + providerInfo);
 
                 parseProvider(providerInfo, pNum, false);
+                profileCheckProviderNames.remove(property);
             }
 
             if (providers.isEmpty()) {
@@ -1284,6 +1300,7 @@ public final class RestrictedSecurity {
                         removedProvider = true;
                         break;
                     }
+                    profileCheckProviderNames.remove(property);
                 }
             }
 
@@ -1311,6 +1328,7 @@ public final class RestrictedSecurity {
                 allInfo.add(property + "=" + providerInfo);
 
                 parseProvider(providerInfo, i, false);
+                profileCheckProviderNames.remove(property);
             }
         }
 
@@ -1634,6 +1652,7 @@ public final class RestrictedSecurity {
                     newValue = value;
                 }
                 profileProperties.put(property, newValue);
+                profileCheckPropertyNames.remove(propertyKey);
                 return true;
             }
             if (debug != null) {
@@ -1704,6 +1723,39 @@ public final class RestrictedSecurity {
                 }
             } else {
                 printStackTraceAndExit("Provider format is incorrect: " + providerInfo);
+            }
+        }
+
+        private void loadProfileCheck(String profileID) {
+            Enumeration<?> pNames = securityProps.propertyNames();
+            String profileDot = profileID + '.';
+            while (pNames.hasMoreElements()) {
+                String name = (String) pNames.nextElement();
+                if (name.startsWith(profileDot)) {
+                    if (name.contains(".jce.provider.")) {
+                        profileCheckProviderNames.add(name);
+                    } else {
+                        profileCheckPropertyNames.add(name);
+                    }
+                }
+            }
+        }
+
+        private void checkProfileCheck(String profileID) {
+            if (!profileCheckProviderNames.isEmpty()) {
+                printStackTraceAndExit(
+                        "The order numbers of providers in profile " + profileID
+                                + " (or a base profile) are not consecutive.");
+            }
+            if (!profileCheckPropertyNames.isEmpty()) {
+                printStackTraceAndExit(
+                        "The property names: "
+                                + profileCheckPropertyNames
+                                        .stream()
+                                        .sorted()
+                                        .collect(Collectors.joining(", "))
+                                + " in profile " + profileID
+                                + " (or a base profile) are not recognized.");
             }
         }
     }


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -994,6 +995,9 @@ private static final class ProfileParser {
         // The java.security properties.
         private final Properties securityProps;
 
+        private LinkedList<String> profileCheckPropertyNames;
+        private LinkedList<String> profileCheckProviderNames;
```
keithc-ca, line 999: Suggest these use `Set<String>` and should be `final`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1016,8 +1020,13 @@ private ProfileParser(String id, Properties props) {
 
             parsedProfiles = new HashSet<>();
 
+            profileCheckPropertyNames = new LinkedList<>();
+            profileCheckProviderNames = new LinkedList<>();
```
keithc-ca, line 1024: Suggest `HashSet`; remove unused import for `LinkedList`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1040,12 +1049,16 @@ private void init(String profileID) {
                 printStackTraceAndExit(profileID + " has already been parsed. Potential infinite recursion.");
             }
 
+            loadProfileCheck(profileID);
+
             String potentialExtendsProfileID = parseProperty(securityProps.getProperty(profileID + ".extends"));
```
keithc-ca, line 1054: Suggest a local to avoid computing `profileID + ".extends"` repeatedly:
```java
            String profileExtends = profileID + ".extends";
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1706,6 +1725,32 @@ private static void checkProviderFormat(String providerInfo, boolean update) {
                 printStackTraceAndExit("Provider format is incorrect: " + providerInfo);
             }
         }
+
+        private void loadProfileCheck(String profileID) {
+            Enumeration<?> pNames = securityProps.propertyNames();
```
keithc-ca, line 1730: Again, suggest a local to avoid repeated concatenation:
```java
            String profileDot = profileID + '.';
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1706,6 +1725,32 @@ private static void checkProviderFormat(String providerInfo, boolean update) {
...
+            while (pNames.hasMoreElements()) {
+                String name = (String) pNames.nextElement();
+                if (name.startsWith(profileID + '.') && name.contains(".jce.provider.")) {
+                    profileCheckProviderNames.add(name);
+                } else if (name.startsWith(profileID + '.')) {
+                    profileCheckPropertyNames.add(name);
+                }
```
keithc-ca, line 1737: Only one call to `startsWith()` is needed:
```java
                if (name.startsWith(profileDot)) {
                    if (name.contains(".jce.provider.")) {
                        profileCheckProviderNames.add(name);
                    } else {
                        profileCheckPropertyNames.add(name);
                    }
                }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1706,6 +1725,32 @@ private static void checkProviderFormat(String providerInfo, boolean update) {
...
+            }
+        }
+
+        private void checkProfileCheck(String profileID) {
+            if (!profileCheckProviderNames.isEmpty()) {
+                printStackTraceAndExit(
+                        "The order of providers in profile " + profileID
```
keithc-ca, line 1744: This message may be misleading; the _order_ may be correct, but there may be a problem with the keys (bad starting key, missing or repeated keys).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1706,6 +1725,32 @@ private static void checkProviderFormat(String providerInfo, boolean update) {
...
+                                + " (or its base profile) is incorrect.");
+            }
+            if (!profileCheckPropertyNames.isEmpty()) {
+                printStackTraceAndExit(
+                        "The property name: " + profileCheckPropertyNames.toString()
```
keithc-ca, line 1749: This message should be reworded to reflect the fact that one or more property names are mentioned.

I'm not sure I agree that it signals misspelling: do we really want to complain and _exit_ if extra, unused properties are specified?

keithc-ca (changes requested):
This only appears to check the profile explicitly requested, but not the profile(s) being extended. Shouldn't the rules apply to all those profiles?

taoliult:
> This only appears to check the profile explicitly requested, but not the profile(s) being extended. Shouldn't the rules apply to all those profiles?

@keithc-ca 

The rules apply to both the base profile and extended profile, details below:

In the initialize restricted security properties method “`private void init(String profileID)`”, it will first invoke the “`loadProfileCheck(profileID)`” to load the provider keyword and property keywords for the current profile (either the base profile or the extended profile). And then, at line 1055, if the current profile extends from a profile (means, the current profile is an extended profile). It will recursively call "`init()`", at line 1096, on the base profile (the profile which the current profile extended from). And in this recursively "`init()`" call, the “`loadProfileCheck(profileID)`” will be invoked again to load the provider keyword and property keywords for the base profile.

keithc-ca:
> The rules apply to both the base profile and extended profile

I don't see how that can be true. There's only one call to `checkProfileCheck()` for the profile requested by the user. How would a badly specified base profile be reported?

taoliult:
> I don't see how that can be true. There's only one call to `checkProfileCheck()` for the profile requested by the user. How would a badly specified base profile be reported?

`checkProfileCheck()` happens after the `init(profileID)` which already initialized the profiles (both base and extended profile)

```
            // Initialize the properties.
            init(profileID);

            checkProfileCheck(profileID);
```

The check only happens for the profile requested by the user. If the requested profile is an extended profile, then it checks both the extended profile and its base profile. If the requested profile is a base profile, then it checks the base profile. It won't check the profile which the user did not requested.

keithc-ca:
Perhaps an (incomplete) example will clarify my point. Consider this trio of (underspecified) profiles:
```
RestrictedSecurity.bottom.jce.provider.2 = sun.security.provider.Sun

RestrictedSecurity.middle.extends = bottom
RestrictedSecurity.middle.jce.provider.4 = sun.security.ec.SunEC

RestrictedSecurity.selected.extends = middle
RestrictedSecurity.selected.jce.provider.1 = com.ibm.crypto.plus.provider.OpenJCEPlusFIPS
RestrictedSecurity.selected.jce.provider.3 = sun.security.rsa.SunRsaSign
RestrictedSecurity.selected.jce.provider.5 = sun.security.ssl.SunJSSE
```
Please ignore the omitted hash and other details irrelevant to the example. Neither "bottom" nor "middle" specify the necessary list of providers (bottom is missing ".1" and middle is missing ".3"), but this code will not complain if the user chooses the "selected" profile.

taoliult:
@keithc-ca 

1. For the example, the bottom (base profile) provider setting as below, which missing “.1”. 

`RestrictedSecurity.bottom.jce.provider.2 = sun.security.provider.Sun`

It will report the below error when initializing. Because the base profile provider MUST start from “.1”.

```
java.lang.RuntimeException: No providers are specified as part of the Restricted Security profile.
	at java.base/openj9.internal.security.RestrictedSecurity.printStackTraceAndExit(RestrictedSecurity.java:657)
	at java.base/openj9.internal.security.RestrictedSecurity$ProfileParser.initProviders(RestrictedSecurity.java:1276)
	at java.base/openj9.internal.security.RestrictedSecurity$ProfileParser.init(RestrictedSecurity.java:1104)
	at java.base/openj9.internal.security.RestrictedSecurity$ProfileParser.init(RestrictedSecurity.java:1096)
	at java.base/openj9.internal.security.RestrictedSecurity$ProfileParser.<init>(RestrictedSecurity.java:1027)
	at java.base/openj9.internal.security.RestrictedSecurity.configure(RestrictedSecurity.java:474)
```

This check happens not in the `checkProfileCheck()` method, it happens in the `initProviders()` method.

2. For the example,  the bottom (base profile) provider setting is good, but the middle is missing “.3”.

```
RestrictedSecurity.bottom.jce.provider.1 = com.ibm.crypto.plus.provider.OpenJCEPlusFIPS
RestrictedSecurity.bottom.jce.provider.2 = sun.security.provider.Sun

RestrictedSecurity.middle.extends = bottom
RestrictedSecurity.middle.jce.provider.4 = sun.security.ec.SunEC
```

It will report the below error from `checkProfileCheck()`. Because in the `updateProviders()` method, after dealing with update of existing providers, it will then move to deal with additional providers added if there are. 

When dealing with additional providers added, the order number MUST start from the “`int i = numOfExistingProviders + 1"`. In above example, that means, it MUST start from “`.3`”. If there is no “`.3`” from the profile “middle” provider list, then `updateProviders()` method won’t keep dealing with any additional providers. That will cause the additional providers in `profileCheckProviderNames` won’t be removed. So the `checkProfileCheck()` will report the error.

```
java.lang.RuntimeException: The order of providers in profile middle (or its base profile) is incorrect.
	at java.base/openj9.internal.security.RestrictedSecurity.printStackTraceAndExit(RestrictedSecurity.java:657)
	at java.base/openj9.internal.security.RestrictedSecurity$ProfileParser.checkProfileCheck(RestrictedSecurity.java:1748)
	at java.base/openj9.internal.security.RestrictedSecurity$ProfileParser.<init>(RestrictedSecurity.java:1029)
	at java.base/openj9.internal.security.RestrictedSecurity.configure(RestrictedSecurity.java:474)
```

3. When initializing the providers, we will do that recursively. So, no matter there are base and extended profiles, or bottom, middle and selected profiles, they will be initialized in the same way.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: When change to below:

                if (name.startsWith(profileID + '.')) {
                    if (name.contains(".jce.provider.")) {
                        profileCheckProviderNames.add(name);
                    } else {
                        profileCheckPropertyNames.add(name);
                    }
                }

There is no need to use a local variable for `profileID + '.'`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: For the provider name check, it only check if the order numbers of providers are consecutive or not. For others, like bad starting key, missing key, etc. For example, 

`RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.jce.provider.1 `wrong typo to `RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.jce.prOOvider.1,` the “`if (!profileCheckPropertyNames.isEmpty()`)” will detect this, since this key can not be recognized and parsed in restricted security mode.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: For which property names can be recognized, they are actually all hard coded in the `loadProperties()` method. So, if there are property name in the profile which can not be recognized during the initialization, that means, the property name either be misspelled, or the property name is not a restricted security profile property at all and shouldn’t be in the profile. So, we exit and print out the error, and then the client can update their java.security file to correct it.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: The benefit of the local would be that it will be evaluated once (total), instead of once per iteration of the loop.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1706,6 +1725,35 @@ private static void checkProviderFormat(String providerInfo, boolean update) {
...
+            if (!profileCheckProviderNames.isEmpty()) {
+                printStackTraceAndExit(
+                        "The order numbers of providers in profile " + profileID
+                                + " (or its base profile) are not consecutive.");
+            }
+            if (!profileCheckPropertyNames.isEmpty()) {
+                printStackTraceAndExit(
+                        "One or more property names: " + profileCheckPropertyNames.toString()
+                                + " in profile " + profileID
+                                + " are misspelled.");
```
keithc-ca, line 1754: This should not use `Set.toString()` (which includes `[]` in the result). Perhaps instead of assuming the property name is "misspelled", the message should say "not recognized".
```java
                printStackTraceAndExit(
                        "The property names: "
                                + profileCheckPropertyNames
                                        .stream()
                                        .sorted()
                                        .collect(Collectors.joining(", "))
                                + " in profile " + profileID
                                + " are not recognized.");
```

keithc-ca (changes requested):
Please correct the spelling of "initialization" in the commit message.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1706,6 +1725,40 @@ private static void checkProviderFormat(String providerInfo, boolean update) {
...
+            if (!profileCheckPropertyNames.isEmpty()) {
+                printStackTraceAndExit(
+                    "The property names: "
+                            + profileCheckPropertyNames
+                                    .stream()
+                                    .sorted()
+                                    .collect(Collectors.joining(", "))
+                            + " in profile " + profileID
+                            + " (or its base profile)"
+                            + " are not recognized.");
```
keithc-ca, line 1759: These lines should be indented four more spaces for consistency with lines 1746-1748.

keithc-ca: Profiles may be extended any number of levels, so "its" should be "a" as in " (or a base profile)".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1706,6 +1725,40 @@ private static void checkProviderFormat(String providerInfo, boolean update) {
...
+            if (!profileCheckPropertyNames.isEmpty()) {
+                printStackTraceAndExit(
+                        "The property names: "
+                                + profileCheckPropertyNames
+                                        .stream()
+                                        .sorted()
+                                        .collect(Collectors.joining(", "))
+                                + " in profile " + profileID
+                                + " (or a base profile)"
+                                + " are not recognized.");
```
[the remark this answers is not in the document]

keithc-ca:

keithc-ca:
@pshipton I don't think the test failures are related to this change, but could you please offer your opinion? Thanks.

pshipton:
I suggest testing without the change for comparison, or testing another version that has stable tests.

keithc-ca:
Testing another version isn't an option here, but https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/848 only adjusts an ignore list, so it should tell us whether those failures are to be expected.

keithc-ca (approved):

keithc-ca:
The test results in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/848 appear to be the same.

PR #850, opened 2024-09-30
Remove redundant export of PATH on Windows

keithc-ca:
See [spec.gmk.template](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/make/autoconf/spec.gmk.template#L165).


pshipton:
jenkins compile win jdknext

pshipton (approved):

PR #851, opened 2024-10-01
Hold the testLoader reference to prevent GC

JasonFengJ9:
Hold the `testLoader` reference to prevent GC

The GC might occur before the second native library loading by another class loader, hold the first `testLoader` reference to prevent GC, and ensure `UnsatisfiedLinkError`.

Related to
* https://github.com/eclipse-openj9/openj9/pull/20265

### DIFF
--- a/test/jdk/java/lang/ClassLoader/nativeLibrary/NativeLibraryTest.java
+++ b/test/jdk/java/lang/ClassLoader/nativeLibrary/NativeLibraryTest.java
@@ -97,6 +97,8 @@ public class NativeLibraryTest {
             throw new RuntimeException("should fail to load the native library" +
                     " by another class loader");
         } catch (UnsatisfiedLinkError e) {}
+        // keep Runnable r strongly reachable so that it is not reclaimable by GC
+        java.lang.ref.Reference.reachabilityFence(r);
     }
 
     /*


test/jdk/java/lang/ClassLoader/nativeLibrary/NativeLibraryTest.java
```diff
@@ -84,29 +84,32 @@ public static void main(String... args) throws Exception {
      */
     static void runTest() throws Exception {
         // invoke p.Test.run() that loads the native library
-        Runnable r = newTestRunnable();
-        r.run();
-
-        // reload the native library by the same class loader
-        r.run();
+        TestLoader testLoader = newTestRunnable();
```
pshipton, line 87: I don't think the test needs to be changed so drastically. It should be sufficient to keep `Runnable r` alive until after `r1.run`. In Java 11+ you can call `java.lang.ref.Reference.reachabilityFence(r1)`.

JasonFengJ9: Replaced it with `Reference.reachabilityFence()`.

pshipton: Tested 5x in a grinder, which passed.

pshipton (approved):

PR #854, opened 2024-10-04
Use native encoded bytes when making a downcall to standard printf 

r30shah:
When making a downcall to standard printf function, use a memorysegment allocated with native charset for strings.

### DIFF
--- a/test/jdk/java/foreign/StdLibTest.java
+++ b/test/jdk/java/foreign/StdLibTest.java
@@ -22,6 +22,12 @@
  */
 
 /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ */
+
+/*
  * @test
  * @run testng/othervm --enable-native-access=ALL-UNNAMED StdLibTest
  */
@@ -29,6 +35,7 @@
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
+import java.nio.charset.Charset;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
@@ -52,6 +59,8 @@ import static org.testng.Assert.*;
 public class StdLibTest extends NativeTestHelper {
 
     final static Linker abi = Linker.nativeLinker();
+
+    final static Charset nativeCharset = Charset.forName(System.getProperty("native.encoding"));
 
     private StdLibHelper stdLibHelper = new StdLibHelper();
 
@@ -306,7 +315,7 @@ public class StdLibTest extends NativeTestHelper {
 
         int printf(String format, List<PrintfArg> args) throws Throwable {
             try (var arena = Arena.ofConfined()) {
-                MemorySegment formatStr = arena.allocateFrom(format);
+                MemorySegment formatStr = arena.allocateFrom(format, nativeCharset);
                 return (int)specializedPrintf(args).invokeExact(formatStr,
                         args.stream().map(a -> a.nativeValue(arena)).toArray());
             }
@@ -387,7 +396,7 @@ public class StdLibTest extends NativeTestHelper {
         INT(int.class, C_INT, "%d", "%d", arena -> 42, 42),
         LONG(long.class, C_LONG_LONG, "%lld", "%d", arena -> 84L, 84L),
         DOUBLE(double.class, C_DOUBLE, "%.4f", "%.4f", arena -> 1.2345d, 1.2345d),
-        STRING(MemorySegment.class, C_POINTER, "%s", "%s", arena -> arena.allocateFrom("str"), "str");
+        STRING(MemorySegment.class, C_POINTER, "%s", "%s", arena -> arena.allocateFrom("str", nativeCharset), "str");
 
         final Class<?> carrier;
         final ValueLayout layout;


r30shah:
@keithc-ca  I checked the code, and also the original commit Can I get review for the changes in this PR ? Checking the history of the file, I believe the change was made to ensure the test passes on 32-bit platform, it does not have anything to do with native encoding. So we would need this for z/OS.

test/jdk/java/foreign/StdLibTest.java
```diff
@@ -305,8 +305,10 @@ int rand() throws Throwable {
         }
 
         int printf(String format, List<PrintfArg> args) throws Throwable {
+            byte[] nativeEncodingBytes = format.getBytes(System.getProperty("native.encoding"));
```
keithc-ca, line 308: Native encoding should be applied not just to the format argument, but all string arguments to `printf()` (i.e. `PrintfArg.STRING` must do the same). Both should probably use `MemorySegment.allocateFrom(String, Charset)`.

This file should also get an IBM copyright notice.

keithc-ca (changes requested):

test/jdk/java/foreign/StdLibTest.java
r30shah: Yes, I did think of it and trying to make the change but the issue is that PrintfArg.STRING  is using enum  and both MemorySegment.allocateFrom(String, Charset) or getting Bytes from String with native encoding throws UnsupportedEncodingException, which I am trying to figure out how to handle when enum constructor is called for `STRING`.  Simple try catch  for lambda is not working. 

I have made the copyright changes.

r30shah: Checking if the static method can be called that handles the exception, and seems like it is working, Let me make the change.

test/jdk/java/foreign/StdLibTest.java
```diff
@@ -387,7 +395,7 @@ enum PrintfArg {
         INT(int.class, C_INT, "%d", "%d", arena -> 42, 42),
         LONG(long.class, C_LONG_LONG, "%lld", "%d", arena -> 84L, 84L),
         DOUBLE(double.class, C_DOUBLE, "%.4f", "%.4f", arena -> 1.2345d, 1.2345d),
-        STRING(MemorySegment.class, C_POINTER, "%s", "%s", arena -> arena.allocateFrom("str"), "str");
+        STRING(MemorySegment.class, C_POINTER, "%s", "%s", arena -> arena.allocateFrom("str", Charset.forName(System.getProperty("native.encoding"))), "str");
```
keithc-ca, line 398: The native charset should be captured in a static field and `allocateFrom()` should be used both here and for the format string.

r30shah: Yes, was working on verifying and making changes, made changes in https://github.com/ibmruntimes/openj9-openjdk-jdk/compare/3a1849e72a67997533198aa87ef88cff848560ec..e23794e3f1100f1efcaee135d6a1c7304cf0ea9e to address this. I did verify changes with JDK23 (https://github.com/r30shah/openj9-openjdk-jdk23/commit/56a31ebee44e6897d7972dee45cef5c5cfc94b04) with grinder : `/view/Test_grinder/job/Grinder/43866/`

test/jdk/java/foreign/StdLibTest.java
```diff
@@ -53,6 +60,8 @@ public class StdLibTest extends NativeTestHelper {
 
     final static Linker abi = Linker.nativeLinker();
 
+    final static Charset charSetNative = Charset.forName(System.getProperty("native.encoding"));
```
keithc-ca, line 63: Normally modifiers should appear in canonical order (`static final`), but it's reasonable to use the same order as line 61.

The name feels awkward to me - `S` should not be capitalized (it's not in "Charset"), and "native" is an adjective here which normally precedes the thing it modifies: how about `nativeCharset` or `nativeEncoding`?

keithc-ca (changes requested):

test/jdk/java/foreign/StdLibTest.java
r30shah: That sounds more clear - changed in https://github.com/ibmruntimes/openj9-openjdk-jdk/compare/3c83057c855478512324c41571b616ee813d3234..6dcafe07f14b47747bef44a718ed060073f7d53e

keithc-ca:
Jenkins test sanity.openjdk amac,win jdknext

keithc-ca:
None of the test failures appear to be related to this change.

keithc-ca (approved):

r30shah:
Yes, I was seeming some test-setup issue when I was trying to run JDKNext internally, but I did test the jdk_foreign1 and jdk_foreign0 target on JDK21 on s390x linux (`view/Test_grinder/job/Grinder/43866`), and in the failing build, I believe that it is not this test that fails.

keithc-ca:
I think the previous title was more to the point; the primary issue is the encoding of strings (a memory segment was allocated before this change and will be after).

r30shah:
Yes, I changed to highlight that the memory segment is now allocated with native charset for printf. I do not have strong opinion about this, so changing it back.

r30shah:
I realized the commit description change causes us to lost the information about launched build : Here is the link of the PR build : https://openj9-jenkins.osuosl.org/job/PullRequest-OpenJDKnext/306/

keithc-ca:
As I said, I think the primary point is the _encoding_ of the string in the memory segment (a memory segment was and will continue to be allocated, so talking about that isn't helpful). The summary here "use native encoded bytes..." more clearly conveys what this change is about.

PR #862, opened 2024-10-15
Add more granular digest flags

KostasTsiounis:
Instead of just having a single flag to enable/disable native support for all digest algorithms, separate flags are created for each one of them.

### DIFF
--- a/src/java.base/share/classes/sun/security/provider/SunEntries.java
+++ b/src/java.base/share/classes/sun/security/provider/SunEntries.java
@@ -99,10 +99,49 @@ import static sun.security.util.SecurityProviderConstants.getAliases;
 
 public final class SunEntries {
 
-    /* The property 'jdk.nativeDigest' is used to control enablement of the native
-     * digest implementation.
-     */
-    private static final boolean useNativeDigest = NativeCrypto.isAlgorithmEnabled("jdk.nativeDigest", "MessageDigest");
+    private static final boolean useNativeMD5;
+    private static final boolean useNativeSHA;
+    private static final boolean useNativeSHA224;
+    private static final boolean useNativeSHA256;
+    private static final boolean useNativeSHA384;
+    private static final boolean useNativeSHA512;
+
+    static {
+        /* The property 'jdk.nativeDigest' is used to control enablement of all native
+         * digest implementations.
+         */
+        boolean useNativeDigest = NativeCrypto.isAlgorithmEnabled("jdk.nativeDigest", "MessageDigest");
+
+        /* The property 'jdk.nativeMD5' is used to control enablement of the native
+         * MD5 implementation.
+         */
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -100,10 +100,40 @@
 public final class SunEntries {
 
     /* The property 'jdk.nativeDigest' is used to control enablement of the native
-     * digest implementation.
+     * digest implementations.
      */
     private static final boolean useNativeDigest = NativeCrypto.isAlgorithmEnabled("jdk.nativeDigest", "MessageDigest");
 
+    /* The property 'jdk.nativeMD5' is used to control enablement of the native
+     * MD5 implementations.
```
keithc-ca, line 108: This (and perhaps all the new fields) should say "implementation" instead of "implementations".

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -100,10 +100,40 @@
...
+     */
+    private static final boolean useNativeMD5 = NativeCrypto.isAlgorithmEnabled("jdk.nativeMD5", "MD5");
```
keithc-ca, line 110: Suggest each of these new flags should use `useNativeDigest` in their initialization. For example:
```java
    private static final boolean useNativeMD5 = useNativeDigest && NativeCrypto.isAlgorithmEnabled("jdk.nativeMD5", "MD5");
```
Later the code need only ask `useNativeMD5` instead of `useNativeDigest && useNativeMD5`. It also avoids most calls to `NativeCrypto.isAlgorithmEnabled()` if `useNativeDigest` is false.

The initializations could be captured in a single `static` block where `useNativeDigest` is only a local variable (there would be no references to `useNativeDigest` afterward).

keithc-ca (changes requested):

src/java.base/share/classes/sun/security/provider/SunEntries.java
KostasTsiounis: Fixed.

src/java.base/share/classes/sun/security/provider/SunEntries.java
KostasTsiounis: Changed to suggestion.

keithc-ca:

KostasTsiounis:
I checked the failing tests and I noticed that I had the wrong class for `SHA384` and `SHA512`. I fixed that with a new commit. Could we rerun the tests?

keithc-ca:
> Could we rerun the tests?

Yes, but please squash first.

keithc-ca:

keithc-ca:
The test failures are not new.

keithc-ca (approved):

PR #864, opened 2024-10-15
Fix StdLibTest for platforms with non standard native charset

r30shah:
If native encoding charset is not one of the standard charset, allocate
segment using native encoded bytes for the string for printf down call
in StdLibTest.

### DIFF
--- a/test/jdk/java/foreign/StdLibTest.java
+++ b/test/jdk/java/foreign/StdLibTest.java
@@ -62,6 +62,12 @@ public class StdLibTest extends NativeTestHelper {
 
     final static Charset nativeCharset = Charset.forName(System.getProperty("native.encoding"));
 
+    static MemorySegment allocateSegmentForString(Arena arena, String str) {
+        // Allocate a memory segment using the native-encoded bytes.
+        byte[] nativeBytes = str.getBytes(nativeCharset);
+        return arena.allocateFrom(ValueLayout.JAVA_BYTE, Arrays.copyOf(nativeBytes, nativeBytes.length + 1));
+    }
+
     private StdLibHelper stdLibHelper = new StdLibHelper();
 
     @Test(dataProvider = "stringPairs")
@@ -315,7 +321,7 @@ public class StdLibTest extends NativeTestHelper {
 
         int printf(String format, List<PrintfArg> args) throws Throwable {
             try (var arena = Arena.ofConfined()) {
-                MemorySegment formatStr = arena.allocateFrom(format, nativeCharset);
+                MemorySegment formatStr = allocateSegmentForString(arena, format);
                 return (int)specializedPrintf(args).invokeExact(formatStr,
                         args.stream().map(a -> a.nativeValue(arena)).toArray());
             }
@@ -396,7 +402,7 @@ public class StdLibTest extends NativeTestHelper {
         INT(int.class, C_INT, "%d", "%d", arena -> 42, 42),
         LONG(long.class, C_LONG_LONG, "%lld", "%d", arena -> 84L, 84L),
         DOUBLE(double.class, C_DOUBLE, "%.4f", "%.4f", arena -> 1.2345d, 1.2345d),
-        STRING(MemorySegment.class, C_POINTER, "%s", "%s", arena -> arena.allocateFrom("str", nativeCharset), "str");
+        STRING(MemorySegment.class, C_POINTER, "%s", "%s", arena -> allocateSegmentForString(arena, "str"), "str");
 
         final Class<?> carrier;
         final ValueLayout layout;


r30shah:
@keithc-ca  Can I get review for this PR. I did test this change on Windows where test failure showed up with JDK23 :  `Test_openjdk23_j9_sanity.openjdk_x86-64_windows_Personal_testList_2/11/console`.

test/jdk/java/foreign/StdLibTest.java
```diff
@@ -62,6 +62,14 @@ public class StdLibTest extends NativeTestHelper {
 
     final static Charset nativeCharset = Charset.forName(System.getProperty("native.encoding"));
 
+    static MemorySegment allocateSegmentForString(Arena arena, String str) {
+        try {
+            return arena.allocateFrom(str, nativeCharset);
+        } catch (IllegalArgumentException ex) {
+            return arena.allocateFrom(str);
+        }
```
keithc-ca, line 70: Why should this be restricted to standard charsets? Why not:
```java
        byte[] data = str.getBytes(nativeCharset);
        data = Arrays.copyOf(data, data.length + 1); // NUL-terminate.
        return arena.allocateFrom(ValueLayout.JAVA_BYTE, data);
```

keithc-ca:
Please also update the commit message to avoid lines longer than 72 characters in the body.

test/jdk/java/foreign/StdLibTest.java
r30shah: You are right, Looking at the API it should be actually right thing to do rather than going through the default charset. 
Code changed in https://github.com/ibmruntimes/openj9-openjdk-jdk/compare/d028c2086dd022dc7efe806a9a2d53f7d23c4540..6c2a1a01939d74c924fa89f0bcf96628c9c0b444

Pushing updated commit message as well.

r30shah:
Updated commit message in https://github.com/ibmruntimes/openj9-openjdk-jdk/compare/6c2a1a01939d74c924fa89f0bcf96628c9c0b444..839d97322d6d32906483aef112ec5657eba9639e

test/jdk/java/foreign/StdLibTest.java
```diff
@@ -62,6 +62,16 @@ public class StdLibTest extends NativeTestHelper {
...
     final static Charset nativeCharset = Charset.forName(System.getProperty("native.encoding"));
 
+    static MemorySegment allocateSegmentForString(Arena arena, String str) {
+        try {
+            return arena.allocateFrom(str, nativeCharset);
+        } catch (IllegalArgumentException ex) {
+            // Allocate a memory segment using the native encoded bytes
+            byte[] nativeBytes = str.getBytes(nativeCharset);
+            return arena.allocateFrom(ValueLayout.JAVA_BYTE, Arrays.copyOf(nativeBytes, nativeBytes.length + 1));
+        }
```
keithc-ca, line 72: I don't think using `try` is appropriate here; just use the `catch` block (with a hyphen in "native-encoded" and a period at the end of the sentence):
```java
        // Allocate a memory segment using the native-encoded bytes.
        byte[] nativeBytes = str.getBytes(nativeCharset);
        return arena.allocateFrom(ValueLayout.JAVA_BYTE, Arrays.copyOf(nativeBytes, nativeBytes.length + 1));
```

keithc-ca (changes requested):

test/jdk/java/foreign/StdLibTest.java
r30shah: Changed in https://github.com/ibmruntimes/openj9-openjdk-jdk/compare/839d97322d6d32906483aef112ec5657eba9639e..748da917b346468236b8ee444c7581e06ffce316

test/jdk/java/foreign/StdLibTest.java
```diff
@@ -62,6 +62,13 @@ public class StdLibTest extends NativeTestHelper {
 
     final static Charset nativeCharset = Charset.forName(System.getProperty("native.encoding"));
 
+    static MemorySegment allocateSegmentForString(Arena arena, String str) {
+        // Allocate a memory segment using the native-encoded bytes.
+        byte[] nativeBytes = str.getBytes(nativeCharset);
+        return arena.allocateFrom(ValueLayout.JAVA_BYTE, Arrays.copyOf(nativeBytes, nativeBytes.length + 1));
+    }
+    }
```
keithc-ca, line 70: Too many closing braces.

keithc-ca:
Jenkins test sanity.openjdk win jdknext

keithc-ca:
Testing failed because it used the openj9 branch instead of the change here; I'll launch a grinder.

keithc-ca:
`StdLibTest` passed in https://openj9-jenkins.osuosl.org/job/Grinder/3903.

keithc-ca (approved):

PR #867, opened 2024-10-24
Check if OpenSSL is in FIPS mode and use Java for some algorithms

KostasTsiounis:
If the `OpenSSL` library used is in `FIPS` mode, avoid using it for algorithms that are not `FIPS` compliant and revert to the original Java implementation.

### DIFF
--- a/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -79,6 +79,8 @@ public class NativeCrypto {
     // or one of the OPENSSL_VERSION_x_x_x constants
     private final long ossl_ver;
 
+    private final boolean isOpenSSLFIPS;
+
     private static long loadCryptoLibraries() {
         long osslVersion;
 
@@ -105,6 +107,11 @@ public class NativeCrypto {
     @SuppressWarnings("removal")
     private NativeCrypto() {
         ossl_ver = AccessController.doPrivileged((PrivilegedAction<Long>) () -> loadCryptoLibraries()).longValue();
+        if (ossl_ver != -1) {
+            isOpenSSLFIPS = isOpenSSLFIPS();
+        } else {
+            isOpenSSLFIPS = false;
+        }
     }
 
     /**
@@ -178,6 +185,54 @@ public class NativeCrypto {
         return traceEnabled;
     }
 
+    public static final boolean isOpenSSLFIPSVersion() {
+        return InstanceHolder.instance.isOpenSSLFIPS;
+    }
+
+    /**
+     * Check whether a native implementation is available in the loaded OpenSSL library.
+     * Note that, an algorithm could be unavailable due to options used to build the
+     * OpenSSL version utilized, or using a FIPS version that doesn't allow it.
+     *
+     * @param algorithm the algorithm checked
+     * @return whether a native implementation of the given crypto algorithm is available
+     */
+    public static final boolean isAlgorithmAvailable(String algorithm) {
+        boolean isAlgorithmAvailable = false;
+        if (isAllowedAndLoaded()) {
+            if (isOpenSSLFIPSVersion()) {
+                switch (algorithm) {
+                case "ChaCha20":
+                case "MD5":
+                    // not available
+                    break;
+                default:
+                    isAlgorithmAvailable = true;
+                    break;
+                }
+            } else {
+                switch (algorithm) {
+                case "MD5":
+                    isAlgorithmAvailable = isMD5Available();
+                    break;
+                default:
+                    isAlgorithmAvailable = true;
+                    break;
+                }
+            }
+        }
+
+        // Issue a message indicating whether the crypto implementation is available.
+        if (traceEnabled) {
+            if (isAlgorithmAvailable) {
+                System.err.println(algorithm + " native crypto implementation is available.");
+            } else {
+                System.err.println(algorithm + " native crypto implementation is not available.");
+            }
+        }
+        return isAlgorithmAvailable;
+    }
+
     @CallerSensitive
     public static NativeCrypto getNativeCrypto() {
         ClassLoader callerClassLoader = Reflection.getCallerClass().getClassLoader();
@@ -203,6 +258,8 @@ public class NativeCrypto {
     private static final native long loadCrypto(boolean trace);
 
     public static final native boolean isMD5Available();
+
+    private static final native boolean isOpenSSLFIPS();
 
     public final native long DigestCreateContext(long nativeBuffer,
                                                  int algoIndex);
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -82,6 +82,9 @@ int OSSL102_RSA_set0_crt_params(RSA *, BIGNUM *, BIGNUM *, BIGNUM *);
 #define EVP_CTRL_AEAD_SET_TAG EVP_CTRL_GCM_SET_TAG
 #endif
 
+/* Whether loaded library is in FIPS mode. */
+static jboolean OSSL_IS_FIPS;
+
 /* Header for EC algorithm */
 jboolean OSSL_ECGF2M;
 int setECPublicCoordinates(EC_KEY *, BIGNUM *, BIGNUM *, int);
@@ -365,6 +368,18 @@ static jlong extractVersionToJlong(const char *astring)
 }
 
 static void *crypto_library = NULL;
+
+/*
+ * Class:     jdk_crypto_jniprovider_NativeCrypto
+ * Method:    isOpenSSLFIPS
+ * Signature: ()Z
+ */
+JNIEXPORT jboolean JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_isOpenSSLFIPS
+  (JNIEnv *env, jclass clazz)
+{
+    return OSSL_IS_FIPS;
+}
+
 /*
  * Class:     jdk_crypto_jniprovider_NativeCrypto
  * Method:    loadCrypto
@@ -437,6 +452,25 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
             unload_crypto_library(crypto_library);
             crypto_library = NULL;
             return -1;
+        }
+    }
+
+    /* Check whether the loaded OpenSSL library is in FIPS mode. */
+    if (ossl_ver >= OPENSSL_VERSION_3_0_0) {
+        typedef int OSSL_fipsmode_t(OSSL_LIB_CTX *);
+        OSSL_fipsmode_t *ossl_fipsmode = (OSSL_fipsmode_t *)find_crypto_symbol(crypto_library, "EVP_default_properties_is_fips_enabled");
+        if ((NULL != ossl_fipsmode) && (1 == (*ossl_fipsmode)(NULL))) {
+            OSSL_IS_FIPS = JNI_TRUE;
+        } else {
+            OSSL_IS_FIPS = JNI_FALSE;
+        }
+    } else {
+        typedef int OSSL_fipsmode_t(void);
+        OSSL_fipsmode_t *ossl_fipsmode = (OSSL_fipsmode_t *)find_crypto_symbol(crypto_library, "FIPS_mode");
+        if ((NULL != ossl_fipsmode) && (1 == (*ossl_fipsmode)())) {
+            OSSL_IS_FIPS = JNI_TRUE;
+        } else {
+            OSSL_IS_FIPS = JNI_FALSE;
         }
     }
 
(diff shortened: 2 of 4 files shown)


closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -70,6 +71,8 @@ public class NativeCrypto {
     private static final boolean traceEnabled = Boolean.parseBoolean(
             GetPropertyAction.privilegedGetProperty("jdk.nativeCryptoTrace", "false"));
 
+    private static final Set<String> disallowedAlgosFIPS = Set.of("MD5", "ChaCha20");
```
keithc-ca, line 74: Suggest set members should be listed alphabetically:
```java
    private static final Set<String> disallowedAlgosFIPS = Set.of("ChaCha20", "MD5");
```

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -178,6 +188,45 @@ public static final boolean isTraceEnabled() {
...
+            }
+            switch (algorithm) {
+                case "MD5":
+                    return isMD5Available();
+                default:
+                    return true;
+            }
+        }
+
+        //Issue a message indicating whether the crypto implementation is available.
```
keithc-ca, line 219: Space after `//`, please.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -178,6 +188,45 @@ public static final boolean isTraceEnabled() {
...
+        if (traceEnabled) {
+            if (isAlgorithmAvailable) {
+                System.err.println(algorithm + " native crypto implementation is available.");
+            } else {
+                System.err.println(algorithm + " native crypto implementation is not available.");
+            }
+        }
+        return isAlgorithmAvailable;
```
keithc-ca, line 227: I don't think this behaves as intended: `isAlgorithmAvailable` is always false here.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -202,6 +251,8 @@ public void run() {
 
     private static final native long loadCrypto(boolean trace);
 
+    private static final native boolean isOpenSSLFIPS();
```
keithc-ca, line 254: Suggest adding this after `isMD5Available()`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -82,6 +82,9 @@ int OSSL102_RSA_set0_crt_params(RSA *, BIGNUM *, BIGNUM *, BIGNUM *);
 #define EVP_CTRL_AEAD_SET_TAG EVP_CTRL_GCM_SET_TAG
 #endif
 
+/* Check whether loaded library is in FIPS mode. */
```
keithc-ca, line 85: "Check" is not appropriate here (a variable is passive). Suggest this flag be `static`:
```c
/* Whether loaded library is in FIPS mode. */
static jboolean OSSL_IS_FIPS;
```
(Most other globals in this file should be `static`, but that might best be corrected separately.)

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -365,6 +368,18 @@ static jlong extractVersionToJlong(const char *astring)
...
 
 static void *crypto_library = NULL;
+
+/*
+ * Class:     jdk_crypto_jniprovider_NativeCrypto
+ * Method:    isOpenSSLFIPS
+ * Signature: ()Z
+ */
+JNIEXPORT jboolean JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_isOpenSSLFIPS
+  (JNIEnv *env, jclass thisObj)
```
keithc-ca, line 378: `thisObj` is a misleading name, the parameter is a class:
```c
  (JNIEnv *env, jclass clazz)
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -440,6 +455,27 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
         }
     }
 
+    /* Check whether the loaded OpenSSL library is in FIPS mode. */
+    if (ossl_ver >= OPENSSL_VERSION_3_0_0) {
+        typedef int OSSL_fipsmode_t(OSSL_LIB_CTX *);
+        OSSL_fipsmode_t* OSSL_fipsmode;
+        OSSL_fipsmode = (OSSL_fipsmode_t*)find_crypto_symbol(crypto_library, "EVP_default_properties_is_fips_enabled");
+        if ((NULL != OSSL_fipsmode) && ((*OSSL_fipsmode)(NULL) == 1)) {
```
keithc-ca, line 463: Suggestions:
* this local should be initialized in its declaration
* names beginning with a capital suggest they're global (this is not)
* missing space before `*` in pointer types
* constant belong on the left of `==` and `!=`
```c
        typedef int OSSL_fipsmode_t(OSSL_LIB_CTX *);
        OSSL_fipsmode_t *ossl_fipsmode = (OSSL_fipsmode_t *)find_crypto_symbol(crypto_library, "EVP_default_properties_is_fips_enabled");
        if ((NULL != ossl_fipsmode) && (1 == (*ossl_fipsmode)(NULL))) {
```
Similar comments for lines 470-472.

src/java.base/share/classes/sun/security/ec/SunEC.java
```diff
@@ -65,30 +65,30 @@ public final class SunEC extends Provider {
     /* The property 'jdk.nativeEC' is used to control enablement of the native
      * ECDH implementation.
      */
-    private static final boolean useNativeECDH = NativeCrypto.isAlgorithmEnabled("jdk.nativeEC", "SunEC");
+    private static final boolean useNativeECDH = NativeCrypto.isAlgorithmEnabled("jdk.nativeEC", "ECDH");
```
keithc-ca, line 68: The changes in this file could be proposed separately.

keithc-ca (changes requested):

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: Fixed.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: You're right. I added the trace later and forgot to update to set `isAlgorithmAvailable` instead of return.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: Moved.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: Changed comment and added `static`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: Changed parameter name.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: Updated in both spots.

src/java.base/share/classes/sun/security/ec/SunEC.java
KostasTsiounis: Reverted and will open a separate PR.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: > names beginning with a capital suggest they're global (this is not)

That part has not been addressed (i.e. `ossl_fipsmode` is a better name than `OSSL_fipsmode`).

KostasTsiounis: Sorry, I missed that. Changed to suggestion.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -178,6 +188,45 @@ public static final boolean isTraceEnabled() {
...
+                    isAlgorithmAvailable = false;
+                }
+            }
+            switch (algorithm) {
+                case "MD5":
+                    isAlgorithmAvailable = isMD5Available();
+                default:
+                    isAlgorithmAvailable = true;
+            }
+        }
```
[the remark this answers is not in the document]
KostasTsiounis: You're right. The switch statement needs to be in an `else` block and a `break` is required after `isAlgorithmAvailable = isMD5Available();`. Thanks for catching that.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -178,6 +188,47 @@ public static final boolean isTraceEnabled() {
...
+                }
+            } else {
+                switch (algorithm) {
+                    case "MD5":
+                        isAlgorithmAvailable = isMD5Available();
+                        break;
+                    default:
+                        isAlgorithmAvailable = true;
+                }
+            }
```
keithc-ca, line 218: This is still not right. In FIPS mode, `isAlgorithmAvailable` will always be `false`.
Suggest we can dispense with `disallowedAlgosFIPS` and just use a `switch` statement for FIPS and non-FIPS modes:
```java
            if (isOpenSSLFIPSVersion()) {
                switch (algorithm) {
                case "ChaCha20":
                case "MD5":
                    // not available
                    break;
                default:
                    isAlgorithmAvailable = true;
                    break;
                }
            } else {
                switch (algorithm) {
                case "MD5":
                    isAlgorithmAvailable = isMD5Available();
                    break;
                default:
                    isAlgorithmAvailable = true;
                    break;
                }
            }
```
Also notice `case` labels are indented the same as `switch` statements.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -440,6 +455,25 @@ JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
         }
     }
 
+    /* Check whether the loaded OpenSSL library is in FIPS mode. */
+    if (ossl_ver >= OPENSSL_VERSION_3_0_0) {
+        typedef int OSSL_fipsmode_t(OSSL_LIB_CTX *);
+        OSSL_fipsmode_t* ossl_fipsmode = (OSSL_fipsmode_t*)find_crypto_symbol(crypto_library, "EVP_default_properties_is_fips_enabled");
```
keithc-ca, line 461: Space before, not after `*`:
```c
        OSSL_fipsmode_t *ossl_fipsmode = (OSSL_fipsmode_t *)find_crypto_symbol(crypto_library, "EVP_default_properties_is_fips_enabled");
```
Similar comment for lin 469.

keithc-ca (changes requested):

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: Changed to suggestion.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: Fixed.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -25,6 +25,7 @@
 
 import java.lang.ref.Cleaner;
 import java.security.*;
+import java.util.Set;
```
keithc-ca, line 28: Please remove this unused import.

keithc-ca (changes requested):
Also, please squash.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: Removed.

KostasTsiounis:
> Also, please squash.

Squashed.

keithc-ca:
Jenkins test sanity amac,plinux jdknext

keithc-ca (approved):
The sanity.openjdk failures are known and unrelated to this change.

PR #869, opened 2024-10-28
Set AIX executables to have exempt SED status

zl-wang:
Customers complained IBM SDK8 has this status, but Semeru Runtimes don't, such that they cannot run Semeru Runtimes on their systems when migrating from IBM SDK8. This is a temporary fix without any other impacts, while we are investigating a complete solution for a larger issue (HugePages are not used for codeCache even though proper -Xlp option is specified).

### DIFF
--- a/make/autoconf/flags-ldflags.m4
+++ b/make/autoconf/flags-ldflags.m4
@@ -27,7 +27,7 @@
 #
 
 # ===========================================================================
-# (c) Copyright IBM Corp. 2019, 2019 All Rights Reserved
+# (c) Copyright IBM Corp. 2019, 2024 All Rights Reserved
 # ===========================================================================
 
 AC_DEFUN([FLAGS_SETUP_LDFLAGS],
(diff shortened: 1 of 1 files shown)


zl-wang:
@keithc-ca please review

keithc-ca (changes requested):
Please fix the commit message. The second line should be blank and following lines should be no more than 72 characters long. See the commit guidelines in [OMR](https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines).

Logically, this is incomplete as the `-brwexec_must` linker option should be applied to **all** executables, including those from OpenJ9 (e.g. `jitserver` which is currently not supported on AIX).

keithc-ca:
> This is a temporary fix without any other impacts, while we are investigating a complete solution for a larger issue (HugePages are not used for codeCache even though proper -Xlp option is specified).

Please create an issue at https://github.com/eclipse-openj9/openj9/issues which describes the larger issue and will serve to track follow-up work to replace this "temporary" fix with a more permanent solution.

zl-wang:
i have it tracked here half-year ago:  https://github.ibm.com/runtimes/openj9-jit-power/issues/414

keithc-ca:
We should have a public issue.

zl-wang:
when we don't have enough resources to make real progresses on the issue any time soon, would you recommend tracking it in GHE instead? at least, that was my motivation.

keithc-ca:
I can think of no reason why that issue shouldn't be publicly visible.

zl-wang:
opened here:  https://github.com/eclipse-openj9/openj9/issues/20440

keithc-ca:
> Please fix the commit message

zl-wang:
done!

keithc-ca:
Lines in the body should be no more than 72 characters in length, and the description should not begin with spaces.

See the commit guidelines in [OMR](https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines).

Please also correct the spelling of "temporary".

keithc-ca:
Jenkins test sanity.functional aix,amac jdknext

keithc-ca (approved):

PR #875, opened 2024-11-01
Add error message when the profile does not exist

taoliult:
In Restricted Security Mode, add the error message when the profile name is used without the version part in "-Dsemeru.customprofile=", and the specified profile does not exist in the java.security file.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -344,23 +344,29 @@ public final class RestrictedSecurity {
                         + selectedProfile);
             }
             String defaultMatch = null;
+            boolean profileExists = false;
             for (Object keyObject : props.keySet()) {
                 if (keyObject instanceof String key) {
-                    if (key.startsWith(potentialProfileID) && key.endsWith(".desc.default")) {
-                        // Check if property is set to true.
-                        if (Boolean.parseBoolean(props.getProperty(key))) {
-                            // Check if multiple defaults exist and act accordingly.
-                            if (defaultMatch == null) {
-                                defaultMatch = key.split("\\.desc")[0];
-                            } else {
-                                printStackTraceAndExit("Multiple default RestrictedSecurity"
-                                        + " profiles for " + selectedProfile);
+                    if (key.startsWith(potentialProfileID)) {
+                        profileExists = true;
+                        if (key.endsWith(".desc.default")) {
+                            // Check if property is set to true.
+                            if (Boolean.parseBoolean(props.getProperty(key))) {
+                                // Check if multiple defaults exist and act accordingly.
+                                if (defaultMatch == null) {
+                                    defaultMatch = key.substring(0, key.length() - ".desc.default".length());
+                                } else {
+                                    printStackTraceAndExit("Multiple default RestrictedSecurity"
+                                            + " profiles for " + selectedProfile);
+                                }
                             }
                         }
                     }
                 }
             }
-            if (defaultMatch == null) {
+            if (!profileExists) {
+                printStackTraceAndExit(selectedProfile + " is not present in the java.security file.");
+            } else if (defaultMatch == null) {
                 printStackTraceAndExit("No default RestrictedSecurity profile was found for "
                         + selectedProfile);
             } else {


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -344,23 +344,29 @@ private static void getProfileID(Properties props) {
                         + selectedProfile);
             }
             String defaultMatch = null;
+            Boolean profileExist = false;
```
keithc-ca, line 347: This should use the primitive type `boolean` instead of the wrapper type. A better name is `profileExists`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -344,23 +344,29 @@ private static void getProfileID(Properties props) {
...
-                    if (key.startsWith(potentialProfileID) && key.endsWith(".desc.default")) {
-                        // Check if property is set to true.
-                        if (Boolean.parseBoolean(props.getProperty(key))) {
-                            // Check if multiple defaults exist and act accordingly.
-                            if (defaultMatch == null) {
-                                defaultMatch = key.split("\\.desc")[0];
-                            } else {
-                                printStackTraceAndExit("Multiple default RestrictedSecurity"
-                                        + " profiles for " + selectedProfile);
+                    if (key.startsWith(potentialProfileID)) {
```
keithc-ca, line 350: This is not right: It will match when it should not. For example, suppose `key` is "sample2", desired profile is "sample".

This also complains when it should not. Suppose there's only one profile with a name beginning with "sample": it should not be necessary to have the `sample.desc.default` property.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -344,23 +344,29 @@ private static void getProfileID(Properties props) {
...
+                        profileExist = true;
+                        if (key.endsWith(".desc.default")) {
+                            // Check if property is set to true.
+                            if (Boolean.parseBoolean(props.getProperty(key))) {
+                                // Check if multiple defaults exist and act accordingly.
+                                if (defaultMatch == null) {
+                                    defaultMatch = key.split("\\.desc")[0];
```
keithc-ca, line 357: This need not use `split()`:
```java
                                    defaultMatch = key.substring(0, key.length() - ".desc.default".length());
```

keithc-ca (changes requested):
Please also correct the commit message and the description here: "is" should be "does" as in "profile does not exist".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca 
Yes, we found this problem weeks ago, and I already opened the issue 2 weeks ago to address this, link https://github.ibm.com/runtimes/jit-crypto/issues/573. I am planning to open another PR to fix this problem. So, do you think, I fix this problem in the current PR, or as I planned, open another PR to fix this?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: Issues relating to public repositories should be public as well (unless there's a good reason otherwise - e.g. a security vulnerability). Please move that issue to https://github.com/eclipse-openj9/openj9/issues.

keithc-ca:
> Please also correct the commit message and the description here: "is" should be "does" as in "profile does not exist".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: @keithc-ca The issue created under the openj9 repo, https://github.com/eclipse-openj9/openj9/issues/20518. I will open another PR to address this issue.

taoliult:
> > Please also correct the commit message and the description here: "is" should be "does" as in "profile does not exist".

@keithc-ca  Done.

keithc-ca:

keithc-ca (approved):
None of the test failures are new.

PR #887, opened 2024-11-14
Fix profiles with same name beginnings being treated as one

taoliult:
Fix the issue where profiles with the same beginnings of profile name are treated as a single profile.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -345,9 +345,10 @@ public final class RestrictedSecurity {
             }
             String defaultMatch = null;
             boolean profileExists = false;
+            String profilePrefix = potentialProfileID + '.';
             for (Object keyObject : props.keySet()) {
                 if (keyObject instanceof String key) {
-                    if (key.startsWith(potentialProfileID)) {
+                    if (key.startsWith(profilePrefix)) {
                         profileExists = true;
                         if (key.endsWith(".desc.default")) {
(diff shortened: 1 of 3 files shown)


taoliult:
@keithc-ca Just checking if there’s any update on this?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -347,7 +347,7 @@ private static void getProfileID(Properties props) {
             boolean profileExists = false;
             for (Object keyObject : props.keySet()) {
                 if (keyObject instanceof String key) {
-                    if (key.startsWith(potentialProfileID)) {
+                    if (key.startsWith(potentialProfileID + '.')) {
```
keithc-ca, line 350: Please capture `potentialProfileID + '.'` in a local rather than compute it for each key.
```java
            String profilePrefix = potentialProfileID + '.';
            for (Object keyObject : props.keySet()) {
                if (keyObject instanceof String key) {
                    if (key.startsWith(profilePrefix)) {
```

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

keithc-ca:
Jenkins test sanity.functional zlinux jdknext

keithc-ca:

keithc-ca (approved):

pshipton:
@keithc-ca fyi the backports

taoliult:
Working on the JDK8 PR, will create the JDK8 PR shortly.

PR #890, opened 2024-11-18

JasonFengJ9:

Conflicts:
```
	src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
	src/java.base/share/classes/jdk/internal/loader/BuiltinClassLoader.java
	src/java.base/share/classes/jdk/internal/loader/NativeLibraries.java
	src/java.base/share/classes/module-info.java
```

### DIFF
--- a/closed/openjdk-tag.gmk
+++ b/closed/openjdk-tag.gmk
@@ -1,1 +1,1 @@
-OPENJDK_TAG := jdk-24+23
+OPENJDK_TAG := jdk-24+24


src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2024, Oracle and/or its affiliates. All rights reserved.
```
keithc-ca, line 2: Rather than patch `LambdaFormEditor` and `MethodTypeForm`, I think we should instead (temporarily?) change the field `USE_SOFT_CACHE` here:
```java
    static final boolean USE_SOFT_CACHE = true;
```
OpenJ9 doesn't need to support "dumping the AOTCache" (which might require the value to be `false`).

Please also retain the import statements added by [8331497: Implement JEP 483: Ahead-of-Time Class Loading & Linking](https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/be095e78c3f87a3a6ef0746d0ec55bf14ff55e5c).

src/java.base/share/classes/jdk/internal/loader/BuiltinClassLoader.java
```diff
@@ -636,30 +617,20 @@ private List<URL> findMiscResource(String name) throws IOException {
...
+                    if (urls == null)
+                        urls = new ArrayList<>();
+                    urls.add(u.toURL());
+                } catch (MalformedURLException | IllegalArgumentException e) {
+                }
+            }
+        }
+        if (urls == null) {
+            urls = List.of();
         }
```
keithc-ca, line 634: Please remove lines 639-641: `urls == null` could never be true after this (they appear to be left over from a previous less-than-ideal merge).

keithc-ca (changes requested):

src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
JasonFengJ9: > static final boolean USE_SOFT_CACHE = true;

If this is hardcoded as `true`, the system property `java.lang.invoke.MethodHandleNatives.USE_SOFT_CACHE` won't be able to change JVM `AOTCache` behaviour.

> OpenJ9 doesn't need to support "dumping the AOTCache" (which might require the value to be false).

OpenJ9 uses OpenJDK MethodHandle implementation since JDK17 and gets the "dumping the AOTCache" function. I don't see any OpenJ9-specific support is required.

keithc-ca: I couldn't find anything in this repository, or elsewhere, that would set `java.lang.invoke.MethodHandleNatives.USE_SOFT_CACHE=false`, so I don't see any need to query that system property. It's not clear to me whether you're agreeing with me nor not.

JasonFengJ9: This is a system property like `java.lang.invoke.MethodHandle.DEBUG_xxx or TRACE_xxx` that can be set by an application at the command line.
These properties are useful for runtime controlling the application behaviours.

src/java.base/share/classes/jdk/internal/loader/BuiltinClassLoader.java
JasonFengJ9: Removed

src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
keithc-ca: Yes, in general, system properties can affect behavior, but in this case it seems the property is defined by a Hotspot JVM based on `-XX:AOTCache=xxx` command-line option. OpenJ9 ignores such options, and I found no place that uses `-Djava.lang.invoke.MethodHandleNatives.USE_SOFT_CACHE=false` so there's nothing to be gained by trying to follow that system property.

keithc-ca: > Please also retain the import statements added by [8331497: Implement JEP 483: Ahead-of-Time Class Loading & Linking](https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/be095e78c3f87a3a6ef0746d0ec55bf14ff55e5c).

`LambdaFormEditor.java` and `MethodTypeForm.java` should be identical to upstream.

JasonFengJ9: Hmm, restored the upstream version.

src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
```diff
@@ -697,6 +697,22 @@ static boolean canBeCalledVirtual(MemberName symbolicRef, Class<?> definingClass
                 symbolicRefClass.isInterface());                     // Mdef implements Msym
     }
 
+    //--- AOTCache support
```
keithc-ca, line 700: Please add the imports (even though they are unused) from upstream that were added in be095e78c3f87a3a6ef0746d0ec55bf14ff55e5c.

keithc-ca (changes requested):

src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
JasonFengJ9: Restored the imports.

PR #900, opened 2024-12-02
Remove unnecessary private ReferenceQueue constructor from OpenJ9

pshipton:


pshipton:
jenkins compile amac jdknext

closed/src/java.base/share/classes/java/lang/ref/NativeReferenceQueue.java
```diff
@@ -37,6 +37,5 @@
  */
 final class NativeReferenceQueue<T> extends ReferenceQueue<T> {
 	public NativeReferenceQueue() {
-		super(0);
```
keithc-ca, line 40: My preference is to keep the explicit `super();` call.

PR #904, opened 2024-12-05

JasonFengJ9:

Conflicts:
```
	src/java.base/share/classes/jdk/internal/ref/PhantomCleanable.java
```


src/java.base/share/classes/jdk/internal/ref/PhantomCleanable.java
```diff
@@ -71,82 +69,29 @@ public abstract class PhantomCleanable<T> extends PhantomReference<T>
...
-    }
-
     /**
      * Unregister this PhantomCleanable and invoke {@link #performCleanup()},
      * ensuring at-most-once semantics.
      */
     @Override
     public final void clean() {
-        if (remove()) {
+        if (list.remove(this)) {
```
keithc-ca, line 94: Does this method not need `@NotCheckpointSafe` (for CRIU)? `remove()` had the annotation before, now the synchronization is in `list.remove()` - same comment for `clear()`. Or should the annotation be attached to `CleanerImpl.CleanableList.remove()`?

JasonFengJ9: The `@NotCheckpointSafe` annotation was added according to the method in a stacktrace from `JVMCheckpointException` or `JVMRestoreException` when CRIU checkpoint or restore failed at single-thread mode. Usually the failure is intermittent. 
The original remove() was removed, it doesn't mean `CleanerImpl.CleanableList.remove()` needs the annotation as well.
The annotation will be added if a `JVMCRIUException` has the method in question. Yeah, it is added as per need.

keithc-ca: I guess we'll see if `cmdLineTester_criu_nonPortableRestore` start failing again.

keithc-ca (approved):

PR #907, opened 2024-12-12
Reorder the way contents of healthcenter jars are extracted

yathamravali:
There are two copies of Version.class one in healthcenter.jar and the other in monitoring-api.jar. Version.class from healthcenter.jar should be used to avoid HealthCenterNotSupportedException caused due to long version date in monitoring-api.jar.


yathamravali:
While enabling Class Histogram feature in Healthcenter getting HealthCenterNotSupportedException

```
com.ibm.java.diagnostics.healthcenter.api.HealthCenterNotSupportedException: The connected agent does not support this function, please upgrade to the latest version. at com.ibm.java.diagnostics.healthcenter.api.vmcontrol.impl.VMControlImpl.collectClassHistogramData(Unknown Source).
WorkingCase.main(WorkingCase.java:64)
```

The exception is because for Semeru Agent version being used is `3.0.21.202109031203` instead of `3.0.21.20210903` causing NumberFormatException resulting in HealthCenterNotSupportedException. After inspecting the Healthcenter code noticed that monitoring-api.jar is being used which has different timestamp pattern than the copy in healthcenter.jar. 

Provided a testpatch with this change and customer confirmed that the issue is resolved with this fix.

closed/make/modules/ibm.healthcenter/Copy.gmk
```diff
@@ -48,8 +48,8 @@ TRANSPORT_PROPERTY_SED_SCRIPT := -e 's|$(TRANSPORT_PROPERTY_REGEX)\s*=.*|$(TRANS
 HEALTHCENTER_COPY := $(JDK_OUTPUTDIR)/modules/$(MODULE)/_the.$(MODULE)_copy_marker
 
 HEALTHCENTER_JARS := \
-	$(HEALTHCENTER_HOME)/healthcenter.jar \
 	$(HEALTHCENTER_HOME)/monitoring-api.jar \
+	$(HEALTHCENTER_HOME)/healthcenter.jar \
```
pshipton, line 52: Pls add a comment to the code to explain the ordering. Otherwise they may be reordered in the future to be alphabetical.

yathamravali: Thanks Peter. Added comment, please review.

pshipton (approved):

PR #908, opened 2024-12-16

JasonFengJ9:

Conflicts:
```
	make/Main.gmk
```


pshipton (approved):

keithc-ca:
We could instead alter the rules to build `static-jdk-image` (and related targets), but that would be a bigger change. I think what we have now is fine.

PR #910, opened 2025-01-02
Apply the patch for CVE-2024-56732

pshipton:
From https://github.com/harfbuzz/harfbuzz/commit/1767f99e2e2196c3fcae27db6d8b60098d3f6d26

### DIFF
--- a/src/java.desktop/share/native/libharfbuzz/hb-utf.hh
+++ b/src/java.desktop/share/native/libharfbuzz/hb-utf.hh
@@ -458,19 +458,21 @@ struct hb_ascii_t
 template <typename utf_t>
 static inline const typename utf_t::codepoint_t *
 hb_utf_offset_to_pointer (const typename utf_t::codepoint_t *start,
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins compile amac jdknext

pshipton:
jenkins compile amac jdknext

src/java.desktop/share/native/libharfbuzz/hb-utf.hh
```diff
@@ -458,19 +458,21 @@ struct hb_ascii_t
 template <typename utf_t>
 static inline const typename utf_t::codepoint_t *
 hb_utf_offset_to_pointer (const typename utf_t::codepoint_t *start,
```
keithc-ca, line 460: This change is not really necessary: `hb_utf_offset_to_pointer()` is not used anywhere in this project (that I can see), but that also makes it safe to apply.

keithc-ca (approved):

PR #912, opened 2025-01-07
Reduce available slots to match the upper limit supported by OpenJ9

babsingh:
Related: https://github.com/eclipse-openj9/openj9/issues/20355

### DIFF
--- a/test/jdk/java/foreign/largestub/TestLargeStub.java
+++ b/test/jdk/java/foreign/largestub/TestLargeStub.java
@@ -22,6 +22,12 @@
  */
 
 /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2025, 2025 All Rights Reserved
+ * ===========================================================================
+ */
+
+/*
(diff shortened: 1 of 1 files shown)


tajila:
Is the test excluded?

babsingh:
> Is the test excluded?

No, it is not excluded; it isn't listed in https://github.com/adoptium/aqa-tests/blob/master/openjdk/excludes/ProblemList_openjdk24-openj9.txt.

keithc-ca:
Please explain how the issue relates to OpenJ9: both the test and the restriction are in openjdk code.

babsingh:
> Please explain how the issue relates to OpenJ9: both the test and the restriction are in openjdk code.

It depends on the downcall and upcall implementation, which is OpenJ9-specific. See https://github.com/eclipse-openj9/openj9/issues/20355#issuecomment-2575602828 for details.

tajila (approved):

PR #914, opened 2025-01-10
Add missing update to the shared class cache search path

theresa-m:
URLClassPath.getResource(String name, boolean check) was removed in 24
 due to the security manager removal. The OpenJ9 patch to update the
scc's search path was lost because of this change.

Port from https://github.com/ibmruntimes/openj9-openjdk-jdk24

Related to https://github.com/eclipse-openj9/openj9/issues/20702


keithc-ca (approved):
Approved based on test results in https://github.com/ibmruntimes/openj9-openjdk-jdk24/pull/8.

PR #924, opened 2025-01-22
Add a check to detect the OpenJCEPlus module

taoliult:
Add a check to detect the OpenJCEPlus module. If the module is missing but the restricted security profile requires it, print an error message and exit.

Signed-off-by: Tao Liu <tao.liu@ibm.com>

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -42,6 +42,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.regex.Matcher;
@@ -67,6 +68,7 @@ public final class RestrictedSecurity {
 
     private static final boolean isNSSSupported;
     private static final boolean isOpenJCEPlusSupported;
+    private static boolean isOpenJCEPlusModuleExist;
 
     private static final boolean userSetProfile;
     private static final boolean shouldEnableSecurity;
@@ -136,6 +138,14 @@ public final class RestrictedSecurity {
             }
         }
         isOpenJCEPlusSupported = isOsSupported && isArchSupported;
+
+        // Check whether the OpenJCEPlus module exists.
+        isOpenJCEPlusModuleExist = false;
+        ModuleLayer layer = ModuleLayer.boot();
+        Optional<Module> module = layer.findModule("openjceplus");
+        if (module.isPresent()) {
+            isOpenJCEPlusModuleExist = true;
+        }
 
         // Check the default solution to see if FIPS is supported.
         isFIPSSupported = isNSSSupported;
@@ -385,6 +395,11 @@ public final class RestrictedSecurity {
         if (profileID.contains("OpenJCEPlus") && !isOpenJCEPlusSupported) {
             printStackTraceAndExit("OpenJCEPlus RestrictedSecurity profiles are not supported"
                     + " on this platform.");
+        }
+
+        if (profileID.contains("OpenJCEPlus") && !isOpenJCEPlusModuleExist) {
+            printStackTraceAndExit("FIPS 140-3 profile specified. Required OpenJCEPlus"
+                    + " module not found.");
         }
 
         if (debug != null) {


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -137,6 +139,14 @@ public String[] run() {
         }
         isOpenJCEPlusSupported = isOsSupported && isArchSupported;
 
+        // Check whether the OpenJCEPlus module exists.
+        isOpenJCEPlusModuleExist = false;
+        ModuleLayer layer = ModuleLayer.boot();
+        Optional<Module> module = layer.findModule("openjceplus");
+        if (module.isPresent()) {
+            isOpenJCEPlusModuleExist = true;
+        }
```
keithc-ca, line 148: The field should be `final` and initialized as:
```java
        isOpenJCEPlusModuleExist = module.isPresent();
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -387,6 +397,11 @@ private static void checkIfKnownProfileSupported() {
                     + " on this platform.");
         }
 
+        if (profileID.contains("OpenJCEPlus") && !isOpenJCEPlusModuleExist) {
```
keithc-ca, line 400: The operands of `&&` should be reversed (doing the cheaper test first).

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -387,6 +394,11 @@ private static void checkIfKnownProfileSupported() {
                     + " on this platform.");
         }
 
+        if (!isOpenJCEPlusModuleExist && profileID.contains("OpenJCEPlus")) {
```
keithc-ca, line 397: This doesn't seem right. Trying to use a profile called "not-OpenJCEPlus" shouldn't demand the presence of the openjceplus module. Also profiles that don't contain that string might actually depend on the openjceplus module.

I would expect to just get "no such provider" or something similar if I try to use an algorithm that requires a module (perhaps openjceplus, but it could be another module) that isn't present in the runtime in use. What are the current symptoms of that situation? Are they really worth potentially breaking legitimate use-cases?

taoliult: The initially discussion about this is, we want to remove the OpenJCEPlus profile from the java.security file in the OpenJ9 builds. The OpenJCEPlus profile should only appear in the java.security file for the Semeru and personal builds. To do this, our first proposal was to modify how the java.security file is generated by introducing a new variable during the process. But, you suggested keeping the profile in the file and instead using a general mechanism to handle any broken profiles. So I am thinking about this current solution to detect the OpenJCEPlus module.

But after the team discussions last Friday, we realized that using this detect the OpenJCEPlus module solution won’t work for the OpenJCEPlus on z/OS. Since the OpenJCEPlus doesn’t exist as a Java module on z/OS.

So, we still think the best solution is to remove the OpenJCEPlus profile from the java.security file in the OpenJ9 builds. That said, it needs to add a new variable during the java.security file generation process. Or, do you have any other suggestions?

keithc-ca: Yes, we could conditionally include sections of `java.security` depending on whether `BUILD_OPENJCEPLUS` is `true` or `false`, but that wouldn't address the problem of a runtime created via `jlink` that excludes the `openjceplus` module.

It's not immediately obvious that we have a serious problem that needs to be solved. What are the symptoms if an algorithm is requested that is not present in the runtime?

taoliult: For the jlink issue, let me think about it. 

It's not a serious problem, it's just that the profile required provider, OpenJCEPlusFIPS, is not exist in the OpenJ9 builds. I will move this PR to the draft mode for now.

PR #925, opened 2025-01-27

JasonFengJ9:

Conflicts:
```
	src/java.base/share/classes/java/security/Provider.java
```

Requires
* https://github.com/eclipse-openj9/openj9/pull/21022


src/java.base/share/classes/java/security/Provider.java
```diff
@@ -1156,7 +1156,7 @@ public Service getService(String type, String algorithm) {
             if (s != null && (!s.isValid() || !RestrictedSecurity.isServiceAllowed(s))) {
                 legacyMap.remove(key, s);
                 // don't return invalid or disallowed legacy services
```
keithc-ca, line 1158: Please delete line 1158 which is not present upstream.

keithc-ca (changes requested):

src/java.base/share/classes/java/security/Provider.java
JasonFengJ9: Removed

PR #879, opened 2024-11-08
Set java.lang.Thread state on transitions

ThanHenderson:
Signed-off-by: Nathan Henderson <nathan.henderson@ibm.com>

PR #927, opened 2025-01-30
Update getState() API to use holder.threadStatus field

fengxue-IS:
depend: https://github.com/eclipse-openj9/openj9/pull/21043
Rework of https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/879

### DIFF
--- a/src/java.base/share/classes/java/lang/Thread.java
+++ b/src/java.base/share/classes/java/lang/Thread.java
@@ -2460,20 +2460,52 @@ public class Thread implements Runnable {
     }
 
     /**
+     * Returns the translation from a J9VMThread state to a Thread::State.
+     *
+     * @param status thread status value set by VM.
+     * @return this thread's state.
+     *
+     * @see State
+     */
+    private State translateJ9VMThreadStateToThreadState(int status) {
+        switch (status) {
+        case 0x1:   // J9VMTHREAD_STATE_RUNNING
+            return State.RUNNABLE;
+        case 0x2:   // J9VMTHREAD_STATE_BLOCKED
+            return State.BLOCKED;
+        case 0x4:   // J9VMTHREAD_STATE_WAITING
+        case 0x80:  // J9VMTHREAD_STATE_PARKED
+            return State.WAITING;
+        case 0x8:   // J9VMTHREAD_STATE_SLEEPING
+        case 0x40:  // J9VMTHREAD_STATE_WAITING_TIMED
+        case 0x100: // J9VMTHREAD_STATE_PARKED_TIMED
+            return State.TIMED_WAITING;
+        case 0x20:  // J9VMTHREAD_STATE_DEAD
+            return State.TERMINATED;
+        default:
+            synchronized (interruptLock) {
+                if (eetop == NO_REF) {
+                    return State.TERMINATED;
+                }
+                return State.values()[getStateImpl(eetop)];
+            }
+        }
+    }
+
+    /**
      * Returns the state of this thread.
      * This method can be used instead of getState as getState is not final and
      * so can be overridden to run arbitrary code.
      */
     State threadState() {
-        synchronized (interruptLock) {
+        if (started) {
             if (eetop == NO_REF) {
-                if (isDead()) {
-                    return State.TERMINATED;
-                }
-                return State.NEW;
+                return State.TERMINATED;
             }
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -2426,21 +2426,57 @@ public State getState() {
...
+            }
+            if (holder == null) {
+                if (eetop == NO_REF) {
                     return State.TERMINATED;
                 }
-                return State.NEW;
+                return State.values()[getStateImpl(eetop)];
             }
-            return State.values()[getStateImpl(eetop)];
+            return translateJ9VMThreadStateToThreadState(holder.threadStatus);
```
keithc-ca, line 2477: Please explain how this can work: I don't see anything in the VM that uses 
`J9VMJAVALANGTHREADFIELDHOLDER(_SET)?_THREADSTATUS`.

fengxue-IS: yes, I've update the comment to depend, sorry for the confusion.

pshipton:
@keithc-ca is this waiting for your review?

keithc-ca:
> @keithc-ca is this waiting for your review?

Yes, but more VM changes are coming, see https://github.com/eclipse-openj9/openj9/pull/21043#issuecomment-2660236597.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -2426,21 +2426,56 @@ public State getState() {
...
+            return State.BLOCKED;
+        case 4: // J9VMTHREAD_STATE_WAITING
+        case 0x80: // J9VMTHREAD_STATE_PARKED
+            return State.WAITING;
+        case 8: // J9VMTHREAD_STATE_SLEEPING
+        case 64: // J9VMTHREAD_STATE_WAITING_TIMED
+        case 0x100: // J9VMTHREAD_STATE_PARKED_TIMED
+            return State.TIMED_WAITING;
+        case 32: // J9VMTHREAD_STATE_DEAD
+            return State.TERMINATED;
```
keithc-ca, line 2451: Please use the same (hex) form of constants as appear in `j9nonbuilder.h`:
```java
        case 0x1: // J9VMTHREAD_STATE_RUNNING
            return State.RUNNABLE;
        case 0x2: // J9VMTHREAD_STATE_BLOCKED
            return State.BLOCKED;
        case 0x4: // J9VMTHREAD_STATE_WAITING
        case 0x80: // J9VMTHREAD_STATE_PARKED
            return State.WAITING;
...
```

keithc-ca (changes requested):
Please rebase to avoid implicit merges.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -2426,21 +2426,56 @@ public State getState() {
...
-                }
-                return State.NEW;
+                return State.TERMINATED;
+            }
+            if (holder == null) {
+                // Gets the thread state via native API under interruptLock.
+                return translateJ9VMThreadStateToThreadState(0);
             }
-            return State.values()[getStateImpl(eetop)];
+            return translateJ9VMThreadStateToThreadState(holder.threadStatus);
```
keithc-ca, line 2476: Suggest using a ternary expression for the argument:
```java
            return translateJ9VMThreadStateToThreadState(
                    (holder != null) ? holder.threadStatus : 0);
```

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -2459,21 +2459,53 @@ public State getState() {
...
             if (eetop == NO_REF) {
-                if (isDead()) {
-                    return State.TERMINATED;
-                }
-                return State.NEW;
+                return State.TERMINATED;
             }
-            return State.values()[getStateImpl(eetop)];
+            return translateJ9VMThreadStateToThreadState(
+                (holder != null) ? holder.threadStatus : 0);
```
keithc-ca, line 2506: Please indent four more spaces as is the convention here (see e.g. line 1376).

keithc-ca:
Jenkins test sanity zlinux jdknext depends eclipse-openj9/openj9#21043

keithc-ca (approved):
I don't think the test failures are related to this change.

PR #928, opened 2025-01-30
Remove --without-cmake support

pshipton:


closed/autoconf/custom-hook.m4
```diff
@@ -52,47 +52,17 @@ AC_DEFUN_ONCE([CUSTOM_EARLY_HOOK],
...
-  if test "x$with_cmake" = xno ; then
-    OPENJ9_ENABLE_CMAKE=false
-
-    # Currently, mixedrefs mode is only available with CMake enabled
-    if test "x$OMR_MIXED_REFERENCES_MODE" != xoff ; then
-      AC_MSG_ERROR([[--with-mixedrefs=[static|dynamic] requires --with-cmake]])
-    fi
+  if AS_EXECUTABLE_P(cmake) ; then
+    CMAKE=cmake
   else
```
keithc-ca, line 63: Please explain this use of `AS_EXECUTABLE_P()` why not just
```
  UTIL_REQUIRE_PROGS([CMAKE], cmake)
```
to allow `CMAKE` from the environment to have effect?

keithc-ca (changes requested):

closed/autoconf/custom-hook.m4
pshipton: I don't understand it so I left it as it was.

pshipton: I'm thinking the idea was that you could specify a different cmake executable, but we don't seem to support that or use that. I'll remove it.

keithc-ca: Actually, I think this should be simpler still; `CMAKE` was already probed in `basic_tools.m4`, we just need to insist if was found:
```
 AC_DEFUN([OPENJ9_CONFIGURE_CMAKE],
[
  if test "x$CMAKE" = x ; then
    AC_MSG_ERROR([CMAKE is required to build OpenJ9])
  fi
])
```

closed/OpenJ9.gmk
```diff
@@ -78,21 +78,10 @@ else
...
-  endif
-else
-  # MAKEFLAGS, as inherited from openjdk, includes options (e.g. '-r' and '-R')
-  # that don't work well with OpenJ9 and OMR when not using cmake.
-  unexport MAKEFLAGS
-  # Filter out unwanted make flags.
-  MAKE_ARGS := $(filter-out -R -r -s,$(MAKE_ARGS))
-  ifneq (,$(JOBS))
-    MAKE_ARGS += -j $(JOBS)
-  endif
```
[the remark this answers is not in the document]
pshipton: I've removed them.

closed/autoconf/custom-hook.m4
pshipton: Changed.

keithc-ca (approved):

keithc-ca:
Jenkins compile amac jdknext

PR #939, opened 2025-02-17
Enforce a more strict FIPS 140-3 JSSE profile definition

jasonkatonica:
The default JSSE provider enables a few services that we would like to
disable by default whenever users are making use of the strict 140-3
FIPS profile. Specific services disabled includes the `PKCS12` KeyStore,
`MD5andSHA1withRSA` Signature, and SSLContexts of name `DTLSv1.0`,
`TLSv1`, and `TLSv1.1`.


keithc-ca:
Please explain how adding a new provider makes the profile more strict. To me, this seems to do the opposite by permitting use of more algorithms via the new provider.

jasonkatonica:
Hi Keith in case it helps, I listed all the JSSE services while running in FIPS strict mode before and after this change and the differences in services listed by the JCE framework looks as follows:

[Difference.html.zip](https://github.com/user-attachments/files/18874135/Difference.html.zip)

keithc-ca (approved):

PR #947, opened 2025-02-24
Prohibit the insertion of 3rd-party provider

JinhangZhang:
Some applications want to use CRIU and enable the -XX:CRIUSecProvider JVM option. That's being said only CRIUSec provider can be used during the pre checkpoint phase of CRIU. However, some non-CRIUSec providers might be inserted, which contradicts the purpose of our -XX:CRIUSecProvider JVM option. Therefore, we added a check in the insertProviderAt function to prevent the addition of non-CRIUSec providers.

Depends on https://github.com/eclipse-openj9/openj9/pull/21176

### DIFF
--- a/src/java.base/share/classes/java/security/Security.java
+++ b/src/java.base/share/classes/java/security/Security.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2024 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2025 All Rights Reserved
  * ===========================================================================
  */
 
@@ -500,11 +500,24 @@ public final class Security {
     public static synchronized int insertProviderAt(Provider provider,
             int position) {
 
+        ProviderList list = Providers.getFullProviderList();
+
         /*[IF CRIU_SUPPORT]*/
+        if (InternalCRIUSupport.enableCRIUSecProvider()) {
+            for (Provider existingProvider : list.providers()) {
+                if ("CRIUSEC".equals(existingProvider.getName())) {
+                    if (criuDebug) {
+                        System.out.println("Trying to insert + " + provider.getName()
+                                + " during the pre-checkpoint which is not allowed.");
+                    }
+                    throw new RuntimeException("Inserting " + provider.getName()
+                            + " during the pre-checkpoint is not allowed");
+                }
+            }
+        }
         CRIUConfigurator.invalidateAlgorithmCache();
         /*[ENDIF] CRIU_SUPPORT */
 
-        ProviderList list = Providers.getFullProviderList();
         ProviderList newList = ProviderList.insertAt(list, provider, position - 1);
         if (list == newList) {
             return -1;


JinhangZhang:
@jasonkatonica FYI

src/java.base/share/classes/java/security/Security.java
```diff
@@ -502,6 +502,16 @@ public static synchronized int insertProviderAt(Provider provider,
...
         CRIUConfigurator.invalidateAlgorithmCache();
+        if (InternalCRIUSupport.enableCRIUSecProvider()) {
+            for (Provider p : Security.getProviders()) {
+                if ("CRIUSEC".equals(p.getName())) {
+                    InternalCRIUSupport.setThirdPartyProviderIsInserted(true);
+                }
+            }
+            if (InternalCRIUSupport.getThirdPartyProviderIsInserted()) {
+                throw new RuntimeException("Inserting a " + provider.getName() + " during the pre-checkpoint is not allowed.");
+            }
```
keithc-ca, line 513: Please avoid single-letter variable names (suggest `existingProvider` instead of `p`).

Please explain why this must depend on API changes to `InternalCRIUSupport`. Why is local state here not sufficient?

This can also use `list` rather than make a copy:
```java
        ProviderList list = Providers.getFullProviderList();

        /*[IF CRIU_SUPPORT]*/
        CRIUConfigurator.invalidateAlgorithmCache();
        if (InternalCRIUSupport.enableCRIUSecProvider()) {
            for (Provider existingProvider : list.providers()) {
                if ("CRIUSEC".equals(existingProvider.getName())) {
                    throw new RuntimeException("Inserting " + provider.getName() + " during the pre-checkpoint is not allowed");
                }
            }
        }
        /*[ENDIF] CRIU_SUPPORT */
```

keithc-ca (changes requested):

src/java.base/share/classes/java/security/Security.java
JinhangZhang: Update the variable name;

The change here, along with the one in OpenJ9, is to prevent some applications from bypassing the insertProviderAt check by catching exceptions, which could lead to third-party providers being incorrectly added during pre-checkpoint. The correct approach is to add the third-party provider during post-restore, otherwise, the third-party provider will be ignored after the checkpoint.

keithc-ca: This throws an exception before adding the offending provider to the list: how would catching that (or any exception) change that?

JinhangZhang: update the code

JinhangZhang: After re-thinking about this, here might be a case: an user want to insert a 3rd party provider, but he encounter this RuntimeException, he catch the exception and threw it anyway. Then he go to checkpoint, but the checkpoint is fine.

I think the whole idea here is that we want to fail the pre-checkpoint or checkpoint if someone want to add a 3rd party provider.

keithc-ca: I don't see a real problem: that unwanted provider is not present and so there's no reason for the checkpoint to fail (it's as if no attempt was made to add the provider).

keithc-ca:
I'm happy with this change, but I won't start testing or merge until you confirm you're satisfied with my [response](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/947#discussion_r1968431145) to the scenario you describe above.

src/java.base/share/classes/java/security/Security.java
JinhangZhang: Hi Keith, after talking with @jasonkatonica , we find it will be hard to diagnose a problem during the post-restore if the exception was caught during the pre-checkpoint and the application bypass the checkpoint. We still think it would be best to fail the checkpoint. Or is it ok that we call `System.exit()` in this location to ensure that users dont reach checkpoint?

keithc-ca: I don't think calling `System.exit()` is appropriate. In earlier versions the security manager can prevent that from terminating the process, so that strategy can't be relied upon for those versions.

I remain unconvinced that it's worthwhile trying to anticipate all misuses of the API.

src/java.base/share/classes/java/security/Security.java
```diff
@@ -500,11 +500,19 @@ public static String getAlgorithmProperty(String algName,
...
+
         /*[IF CRIU_SUPPORT]*/
         CRIUConfigurator.invalidateAlgorithmCache();
+        if (InternalCRIUSupport.enableCRIUSecProvider()) {
+            for (Provider existingProvider : list.providers()) {
+                if ("CRIUSEC".equals(existingProvider.getName())) {
+                    throw new RuntimeException("Inserting " + provider.getName() + " during the pre-checkpoint is not allowed");
+                }
+            }
+        }
```
keithc-ca, line 513: I suggest this addition should appear before line 506: if the provider is not added, that call is inappropriate.

keithc-ca (changes requested):

src/java.base/share/classes/java/security/Security.java
JinhangZhang: Update the code; Also add another debug trace inside if you think that is appropriate?

src/java.base/share/classes/java/security/Security.java
JinhangZhang: Ok, Im fine without using System.exit()

src/java.base/share/classes/java/security/Security.java
```diff
@@ -500,11 +500,22 @@ public static String getAlgorithmProperty(String algName,
...
+        ProviderList list = Providers.getFullProviderList();
+
         /*[IF CRIU_SUPPORT]*/
+        if (InternalCRIUSupport.enableCRIUSecProvider()) {
+            for (Provider existingProvider : list.providers()) {
+                if ("CRIUSEC".equals(existingProvider.getName())) {
+                    if (criuDebug) {
+                        System.out.println("Trying to insert + " + provider.getName() + " during the pre-checkpoint but is not allowed.");
+                    }
+                    throw new RuntimeException("Inserting " + provider.getName() + " during the pre-checkpoint is not allowed");
```
keithc-ca, line 512: In the debug message, I think "but" should be "which", and the lines here are too long:
```java
                    if (criuDebug) {
                        System.out.println("Trying to insert + " + provider.getName()
                                + " during the pre-checkpoint which is not allowed.");
                    }
                    throw new RuntimeException("Inserting " + provider.getName()
                            + " during the pre-checkpoint is not allowed");
```

JinhangZhang: update the code

keithc-ca:

keithc-ca (approved):

PR #952, opened 2025-02-26
Include System.c natives to make use of SystemProps$Raw

pshipton:
Related to https://github.com/eclipse-openj9/openj9/issues/21189

Remove the Java_java_lang_System_mapLibraryName native so it can't conflict with the OpenJ9 native of the same name.

Depends on https://github.com/eclipse-openj9/openj9/pull/21203


keithc-ca:
Do we need to be concerned about the definition of `Java_java_lang_System_mapLibraryName()` that clashes with OpenJ9?
They're in different shared libraries, but is there any danger of the wrong one being used?

keithc-ca:
How does this help if nothing uses `SystemProps`?

pshipton:
It enables https://github.com/eclipse-openj9/openj9/pull/21200

pshipton:
> Do we need to be concerned about the definition of Java_java_lang_System_mapLibraryName() that clashes with OpenJ9?
They're in different shared libraries, but is there any danger of the wrong one being used?

I'm not sure which would be used, I don't think we should have natives with the same names. Do you agree I should rename the one in OpenJ9 to avoid the conflict?

keithc-ca:
> rename the one in OpenJ9

That would involve making that method in OpenJ9 non-native so it can call a new native method. I don't expect performance is critical here, and stack traces should not matter (or we could hide the new native). Or it could just be removed from the extension repository (surround with `#if 0` / `#endif`)?

pshipton:
For what it's worth, it appears the OpenJ9 native is called, at last on xlinux. Doing some tracing, I see libjclse29.so loaded before libjava.so.

pshipton:
Added the `#if 0`.

src/java.base/share/native/libjava/System.c
```diff
@@ -297,6 +303,7 @@ static void cpchars(jchar *dst, char *src, int n)
     }
 }
 
+#if 0
```
keithc-ca, line 306: A comment explaining why this is omitted would be welcome, with another on the `#endif` linking it to here.

pshipton: Added a comment.

src/java.base/share/native/libjava/System.c
```diff
@@ -297,6 +303,7 @@ static void cpchars(jchar *dst, char *src, int n)
     }
 }
 
+#if 0 /* Exclude so it doesn't conflict with the OpenJ9 native. */
```
keithc-ca, line 306: I expect we need to move this before `cpchars()` else warnings will ensue that that function is unused.

pshipton: Moved it.

keithc-ca:
Jenkins compile amac,win jdknext

keithc-ca (approved):

PR #954, opened 2025-02-27
Allow multiple constraints for each algorithm

KostasTsiounis:
If a constraint for an algorithm is found and the class attempting to utilize it doesn't match the accepted uses, the algorithm is considered not allowed and loading it does not succeed.

Instead, we want to check all available constraints for an algorithm before deciding if it is allowed to be used by a specific module, package and/or class.

Additional test cases are added to check this functionality.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -789,6 +789,7 @@ public final class RestrictedSecurity {
             if (debug != null) {
                 debug.println("Security constraints check of provider.");
             }
+            constraints:
             for (Constraint constraint : constraints) {
                 String cType = constraint.type;
                 String cAlgorithm = constraint.algorithm;
@@ -807,14 +808,14 @@ public final class RestrictedSecurity {
                     if (debug != null) {
                         debug.println("The constraint doesn't apply to the service type.");
                     }
-                    continue;
+                    continue constraints;
                 }
                 if (!isAsterisk(cAlgorithm) && !algorithm.equalsIgnoreCase(cAlgorithm)) {
                     // The constraint doesn't apply to the service algorithm.
                     if (debug != null) {
                         debug.println("The constraint doesn't apply to the service algorithm.");
                     }
-                    continue;
+                    continue constraints;
                 }
 
                 // For type and algorithm match, and attribute is not *.
@@ -836,7 +837,8 @@ public final class RestrictedSecurity {
                                     + "\nagainst the service attribute value: " + sValue);
                         }
                         if ((sValue == null) || !cValue.equalsIgnoreCase(sValue)) {
-                            // If any attribute doesn't match, return service is not allowed.
+                            // If any of the attributes don't match,
+                            // then this constraint doesn't match so move on.
                             if (debug != null) {
                                 debug.println("Attributes don't match!");
                                 debug.println("The following service:"
@@ -845,7 +847,7 @@ public final class RestrictedSecurity {
                                             + "\n\tAttribute: " + cAttribute
                                             + "\nis NOT allowed in provider: " + providerClassName);
                             }
-                            return false;
+                            continue constraints;
                         }
                         if (debug != null) {
                             debug.println("Attributes match!");
@@ -903,7 +905,7 @@ public final class RestrictedSecurity {
                     }
 
                     // If nothing matching the accepted uses is found in the call stack,
-                    // this service is not allowed.
+                    // then this constraint doesn't match so move on.
                     if (!found) {
                         if (debug != null) {
                             debug.println("Classes in call stack are not part of accepted uses!");
@@ -914,7 +916,7 @@ public final class RestrictedSecurity {
                                         + "\n\tAccepted uses: " + cAcceptedUses
                                         + "\nis NOT allowed in provider: " + providerClassName);
                         }
-                        return false;
+                        continue constraints;
                     }
                 }
 
--- a/closed/test/jdk/openj9/internal/security/TestConstraintsSuccess.java
+++ b/closed/test/jdk/openj9/internal/security/TestConstraintsSuccess.java
@@ -77,6 +77,16 @@ public class TestConstraintsSuccess {
         KeyManagerFactory.getInstance("SunX509");
         TrustManagerFactory.getInstance("SunX509");
         SSLContext.getInstance("TLSv1.3");
+
+        // Since there are three constraints for MD5, with only the middle one
+        // allowing for use by this class, successfully getting the algorithm
+        // verifies that all constraints are checked.
+        MessageDigest.getInstance("MD5");
+
+        // Since there are three constraints for SHA512withECDSA, with only the
+        // middle one having the correct attributes, successfully getting the
+        // algorithm verifies that all constraints are checked.
+        Signature.getInstance("SHA512withECDSA");
     }
 
     @Test
--- a/closed/test/jdk/openj9/internal/security/constraints-java.security
+++ b/closed/test/jdk/openj9/internal/security/constraints-java.security
@@ -21,7 +21,7 @@
 RestrictedSecurity.TestConstraints.Version.desc.name = Test Base Profile
 RestrictedSecurity.TestConstraints.Version.desc.default = false
 RestrictedSecurity.TestConstraints.Version.desc.fips = false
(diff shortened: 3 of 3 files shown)


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -914,7 +914,7 @@ boolean isRestrictedServiceAllowed(Service service, boolean checkUse) {
                                         + "\n\tAccepted uses: " + cAcceptedUses
                                         + "\nis NOT allowed in provider: " + providerClassName);
                         }
-                        return false;
+                        continue;
```
keithc-ca, line 917: Should line 848 also continue with the next constraint?

KostasTsiounis: I thought about it, but I don't see a use case where one would have multiple constraints with different attributes. I can switch that for consistency though, if you feel that's better.

keithc-ca: I think it was necessary even before the extension of constraints. Suppose the profile contains:
```
    {KeyFactory, EC, KeySize=192},
    {KeyFactory, EC, KeySize=256},
```
Attempts to use keys of size 256 would fail otherwise. If those entries were reversed, 192 bit keys could not be used.

KostasTsiounis: That makes sense. I updated line 848 to move on to the next constraint, instead of returning false. I've also updated the comments to reflect this behaviour.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -845,7 +846,7 @@ boolean isRestrictedServiceAllowed(Service service, boolean checkUse) {
                                             + "\n\tAttribute: " + cAttribute
                                             + "\nis NOT allowed in provider: " + providerClassName);
                             }
-                            return false;
+                            continue;
```
keithc-ca, line 849: This needs to continue the loop that starts on line 792, otherwise constraint attributes have no meaning.
Suggest a label on that outer loop:
```java
            constraints:
            for (Constraint constraint : constraints) {
                String cType = constraint.type;
...
                    for (String attribute : cAttributeArray) {
...
                            continue constraints; /* replaces return false; */
```
For consistency, the new continue statement below should use the same label:
```java
                        continue constraints;
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -836,7 +837,8 @@ boolean isRestrictedServiceAllowed(Service service, boolean checkUse) {
                                     + "\nagainst the service attribute value: " + sValue);
                         }
                         if ((sValue == null) || !cValue.equalsIgnoreCase(sValue)) {
-                            // If any attribute doesn't match, return service is not allowed.
+                            // If any of the attributes doesn't match,
```
keithc-ca, line 840: mixing cardinalities: either "any attribute doesn't" or "any of the attributes don't".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -814,7 +815,7 @@ boolean isRestrictedServiceAllowed(Service service, boolean checkUse) {
                     if (debug != null) {
                         debug.println("The constraint doesn't apply to the service algorithm.");
                     }
-                    continue;
+                    continue constraints;
```
keithc-ca, line 818: Please also use the label on line 811 and line 919, or only on line 850.

keithc-ca (changes requested):
I took pause at the use of `Thread.currentThread().getStackTrace()` when reviewing #935, but only today did I remember we have a better tool available in the `StackWalker` class (available since Java 9). I created https://github.com/eclipse-openj9/openj9/issues/21238 to track that improvement.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Went with the second option.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Used the label everywhere.

keithc-ca (commented):
Please squash.

keithc-ca:
Perhaps the test suite should be expanded to validate this change?

KostasTsiounis:
> Perhaps the test suite should be expanded to validate this change?

Yeah, probably a good idea. I'm thinking of adding a constraint that won't be used before the one that is supposed to work and let the algorithm be loaded. I can probably try the same with the attributes.

KostasTsiounis:
> > Perhaps the test suite should be expanded to validate this change?
> 
> Yeah, probably a good idea. I'm thinking of adding a constraint that won't be used before the one that is supposed to work and let the algorithm be loaded. I can probably try the same with the attributes.

Does something like 21e358bf54aae36fe79f2e47f2a69f0f5feb04f3 make sense?

KostasTsiounis:
> It occurs to me now, that constraints might overlap. I can't think of any negative consequences, but do wonder whether that should elicit a warning.
> 
> Some examples. Here the second entry is entirely included in the first; they could be reversed to express equivalent constraints:
> 
> ```
>     {MessageDigest, SHA-256, *}, \
>     {MessageDigest, SHA-256, *, FullClassName:TestConstraintsSuccess}, \
> ```
> 
> Here the entries are identical:
> 
> ```
>     {MessageDigest, SHA-512, *}, \
>     {MessageDigest, SHA-512, *}, \
> ```
> 
> What, if anything, do you think should be done about situations like these?

In this case the first one will be considered. It's the same behaviour as in other properties within the default `java.security` file. I think that parsing all of them and performing additional logic to consolidate them is beyond the scope of what we are trying to do here and contradicts the behaviour of the `java.security` properties as well.

KostasTsiounis:
> > It occurs to me now, that constraints might overlap. I can't think of any negative consequences, but do wonder whether that should elicit a warning.
> > Some examples. Here the second entry is entirely included in the first; they could be reversed to express equivalent constraints:
> > ```
> >     {MessageDigest, SHA-256, *}, \
> >     {MessageDigest, SHA-256, *, FullClassName:TestConstraintsSuccess}, \
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Here the entries are identical:
> > ```
> >     {MessageDigest, SHA-512, *}, \
> >     {MessageDigest, SHA-512, *}, \
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What, if anything, do you think should be done about situations like these?
> 
> In this case the first one will be considered. It's the same behaviour as in other properties within the default `java.security` file. I think that parsing all of them and performing additional logic to consolidate them is beyond the scope of what we are trying to do here and contradicts the behaviour of the `java.security` properties as well.

Actually, come to think of it, the inverse is happening in existing `java.security` properties (i.e., the last one is taken into account). 

A warning could be printed. Perhaps in debug mode... I'll open an issue for us to discuss this with @jasonkatonica and the rest of the team.

KostasTsiounis:
> I suggest [21e358b](https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/21e358bf54aae36fe79f2e47f2a69f0f5feb04f3) would be improved by using a name other than `TestConstraintsFailure` (so as not to affect that test), and adding a third MD5 entry so the matching entry is neither first nor last (and updating the comment in the test code).
> 
> Similarly, the attribute test would be improved if the matching entry is neither first nor last.

Updated with suggestions in 42f0fa3960d0cab572d60bc3773bd8f1e95d37d8. Rebuilding to ensure test works as expected.

KostasTsiounis:
Rebuilt and run test with latest changes and it works fine.

keithc-ca:
Please squash; if you think it make sense to capture the test changes in a second commit that would be fine.

keithc-ca:
Jenkins test sanity.functional amac jdknext

keithc-ca (approved):

PR #956, opened 2025-03-03
Fix selectedProfile overridden by a null property semeru.customprofile

JasonFengJ9:
Fix `selectedProfile` overridden by a `null` property `semeru.customprofile`

Resolve [an internal issue](https://github.ibm.com/runtimes/backlog/issues/1612)

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -138,9 +138,14 @@ public final class RestrictedSecurity {
             }
         }
 
-        // If user has specified a profile, use that
-        selectedProfile = System.getProperty("semeru.customprofile");
-        userSetProfile = selectedProfile != null;
+        // If user has specified a profile, use that.
+        String semeruCustomProfile = System.getProperty("semeru.customprofile");
+        if (semeruCustomProfile != null) {
+            selectedProfile = semeruCustomProfile;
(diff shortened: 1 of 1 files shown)


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -138,9 +138,12 @@ public final class RestrictedSecurity {
...
 
-        // If user has specified a profile, use that
-        selectedProfile = System.getProperty("semeru.customprofile");
-        userSetProfile = selectedProfile != null;
+        // If user has specified a profile, use that.
+        String semeruCustomprofile = System.getProperty("semeru.customprofile");
+        if (semeruCustomprofile != null) {
+            selectedProfile = semeruCustomprofile;
+        }
+        userSetProfile = semeruCustomprofile != null;
```
keithc-ca, line 146: "p" should be capitalized; suggest `customProfile` for consistency with other fields/locals.
Also suggest the null check need not be repeated:
```java
        String customProfile = System.getProperty("semeru.customprofile");
        if (customProfile != null) {
            selectedProfile = customProfile;
            userSetProfile = true;
        } else {
            userSetProfile = false;
        }
```

JasonFengJ9: Modified

keithc-ca (approved):

PR #957, opened 2025-03-04
Only check and set FIPS mode in FIPS profiles

KostasTsiounis:
A `RestrictedSecurity` profile doesn't have to be a `FIPS` profile, in which case a `FIPS mode` property is not required and the corresponding system property should not be set.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -558,11 +558,17 @@ public final class RestrictedSecurity {
         propsMapping.put("jdk.tls.legacyAlgorithms", restricts.jdkTlsLegacyAlgorithms);
         propsMapping.put("jdk.certpath.disabledAlgorithms", restricts.jdkCertpathDisabledAlgorithms);
         propsMapping.put("jdk.security.legacyAlgorithms", restricts.jdkSecurityLegacyAlgorithms);
-        String fipsMode = System.getProperty("com.ibm.fips.mode");
-        if (fipsMode == null) {
-            System.setProperty("com.ibm.fips.mode", restricts.jdkFipsMode);
-        } else if (!fipsMode.equals(restricts.jdkFipsMode)) {
-            printStackTraceAndExit("Property com.ibm.fips.mode is incompatible with semeru.customprofile and semeru.fips properties");
+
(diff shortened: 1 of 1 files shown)


keithc-ca:
Please provide some background information motivating this change.

KostasTsiounis:
> Please provide some background information motivating this change.

If one tries to load a non-FIPS profile that doesn't contain the `fips.mode` property, the result is a `NullPointerException`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -558,11 +558,14 @@ private static void setProperties(Properties props) {
...
-        String fipsMode = System.getProperty("com.ibm.fips.mode");
-        if (fipsMode == null) {
-            System.setProperty("com.ibm.fips.mode", restricts.jdkFipsMode);
-        } else if (!fipsMode.equals(restricts.jdkFipsMode)) {
-            printStackTraceAndExit("Property com.ibm.fips.mode is incompatible with semeru.customprofile and semeru.fips properties");
+
+        if (restricts.descIsFIPS) {
+            String fipsMode = System.getProperty("com.ibm.fips.mode");
+            if (fipsMode == null) {
+                System.setProperty("com.ibm.fips.mode", restricts.jdkFipsMode);
```
keithc-ca, line 565: Please explain what guarantees that `restricts.jdkFipsMode != null`. If there is no such guarantee, I don't see how this change necessarily avoids the `NullPointerException`.

KostasTsiounis: Nothing guarantees that `restricts.jdkFipsMode != null`, but it should be set when the profile is a FIPS one and an exception should be thrown if it's not. However, a `NullPointerException` is not the most appropriate here, so I will add another check and throw an exception with a better message.

KostasTsiounis: Added extra check through 44ad7f705c98468e06f01b0078048fb1fb1e23dd.

keithc-ca:
Please squash and rebase to avoid the implicit merge.

keithc-ca:
Jenkins compile amac jdknext

keithc-ca (approved):

PR #944, opened 2025-02-20
CRIU adds @NotCheckpointSafe for PhantomCleanable constructor

JasonFengJ9:
CRIU adds `@NotCheckpointSafe` for `PhantomCleanable` constructor

`@NotCheckpointSafe` should be added for the caller of the `synchronized` method `list.insert(this)`, i.e., `PhantomCleanable(T referent, Cleaner cleaner)`.

Related to
* https://github.com/eclipse-openj9/openj9/issues/21111

Signed-off-by: Jason Feng <fengj@ca.ibm.com>

PR #936, opened 2025-02-13
CRIU adds @NotCheckpointSafe for CleanerImpl.CleanableList.insert()

JasonFengJ9:
CRIU adds `@NotCheckpointSafe` for `CleanerImpl.CleanableList.insert()`

Removed `jdk/internal/ref/PhantomCleanable.java` from `COPY_OVERLAY_FILES` list which only applies to JDK21 and earlier versions due to the change of `CleanerImpl.CleanableList` since JDK24.

Related to
* https://github.com/eclipse-openj9/openj9/issues/21111

Signed-off-by: Jason Feng <fengj@ca.ibm.com>

PR #960, opened 2025-03-05
CRIU adds @NotCheckpointSafe for CleanerImpl.CleanableList methods

JasonFengJ9:
CRIU adds `@NotCheckpointSafe` for `CleanerImpl.CleanableList` methods

Added `@NotCheckpointSafe` to `CleanerImpl.CleanableList` `isEmpty()`, `insert()` and `remove()`.

Related to 
* https://github.com/eclipse-openj9/openj9/issues/21111#issuecomment-2698203372

The thread `"Common-Cleaner"` entered lock: ` jdk/internal/ref/CleanerImpl$CleanableList@0x000000070A36BCE8`.
```
3XMTHREADINFO      "Common-Cleaner" J9VMThread:0x0000000000172F00, omrthread_t:0x00007F19A4186B60, java/lang/Thread:0x000000070A367470, state:R, prio=8
3XMJAVALTHREAD            (java/lang/Thread getId:0x3, isDaemon:true)
3XMJAVALTHRCCL            No Java context classloader associated with this thread
3XMTHREADINFO1            (native thread ID:0x22BCBB, native priority:0x8, native policy:UNKNOWN, vmstate:CW, vm thread flags:0x00200081)
3XMTHREADINFO2            (native stack address range from:0x00007F1957C8F000, to:0x00007F1957D0F000, size:0x80000)
3XMCPUTIME               CPU usage total: 0.001903862 secs, current category="Application"
3XMHEAPALLOC             Heap bytes allocated since last GC cycle=0 (0x0)
3XMTHREADINFO3           Java callstack:
4XESTACKTRACE                at jdk/internal/ref/CleanerImpl$CleanableList.isEmpty(CleanerImpl.java:270)
5XESTACKTRACE                   (entered lock: jdk/internal/ref/CleanerImpl$CleanableList@0x000000070A36BCE8, entry count: 1)
4XESTACKTRACE                at jdk/internal/ref/CleanerImpl.run(CleanerImpl.java:132)
4XESTACKTRACE                at java/lang/Thread.runWith(Thread.java:1471)
4XESTACKTRACE                at java/lang/Thread.run(Thread.java:1458)
4XESTACKTRACE                at jdk/internal/misc/InnocuousThread.run(InnocuousThread.java:148)
```

The following thread is the checkpoint main thread.
```
22:02:54   [OUT] Caused by: openj9.internal.criu.JVMCheckpointException: Blocking operation is not allowed in CRIU single thread mode.
22:02:54   [OUT] 	at java.base/jdk.internal.ref.PhantomCleanable.<init>(PhantomCleanable.java:86)
22:02:54   [OUT] 	at java.base/jdk.internal.ref.CleanerImpl$PhantomCleanableRef.<init>(CleanerImpl.java:164)
22:02:54   [OUT] 	at java.base/java.lang.ref.Cleaner.register(Cleaner.java:225)
22:02:54   [OUT] 	at java.base/java.lang.invoke.MethodHandleNatives$CallSiteContext.make(MethodHandleNatives.java:92)
22:02:54   [OUT] 	at java.base/java.lang.invoke.CallSite.<init>(CallSite.java:145)
22:02:54   [OUT] 	at java.base/java.lang.invoke.ConstantCallSite.<init>(ConstantCallSite.java:50)
22:02:54   [OUT] 	at java.base/java.lang.invoke.InnerClassLambdaMetafactory.buildCallSite(InnerClassLambdaMetafactory.java:229)
```
`PhantomCleanable.<init>` attempts to enter the same lock via `CleanerImpl$CleanableList.insert()`.

The `@NotCheckpointSafe` annotation should be placed at a method invoked by the non-checkpoint thread that entered the lock.
There are other `synchronized` methods within `CleanerImpl$CleanableList` such as `reset()`, `insert()`, and `remove()`. Not adding annotations for these methods for now. The failure was highly intermittent and can't be reproduced in a regular-size grinder.

### DIFF
--- a/src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
+++ b/src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2025, 2025 All Rights Reserved
+ * ===========================================================================
+ */
+
 package jdk.internal.ref;
 
 import java.lang.ref.Cleaner;
@@ -35,6 +41,10 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 
 import jdk.internal.misc.InnocuousThread;
+
+/*[IF CRIU_SUPPORT]*/
+import openj9.internal.criu.NotCheckpointSafe;
+/*[ENDIF] CRIU_SUPPORT */
 
 /**
  * CleanerImpl manages a set of object references and corresponding cleaning actions.
@@ -265,6 +275,9 @@ public final class CleanerImpl implements Runnable {
          *
          * @return true if the list is empty
          */
+        /*[IF CRIU_SUPPORT]*/
+        @NotCheckpointSafe
+        /*[ENDIF] CRIU_SUPPORT */
         public synchronized boolean isEmpty() {
             // Head node size is zero only when the entire list is empty.
             return head.size == 0;
@@ -273,6 +286,9 @@ public final class CleanerImpl implements Runnable {
         /**
          * Insert this PhantomCleanable in the list.
          */
+        /*[IF CRIU_SUPPORT]*/
+        @NotCheckpointSafe
+        /*[ENDIF] CRIU_SUPPORT */
         public synchronized void insert(PhantomCleanable<?> phc) {
             if (head.size == NODE_CAPACITY) {
                 // Head node is full, insert new one.
@@ -303,6 +319,9 @@ public final class CleanerImpl implements Runnable {
          * @return true if Cleanable was removed or false if not because
          * it had already been removed before
          */
+        /*[IF CRIU_SUPPORT]*/
+        @NotCheckpointSafe
+        /*[ENDIF] CRIU_SUPPORT */
         public synchronized boolean remove(PhantomCleanable<?> phc) {
             if (phc.node == null) {
                 // Not in the list.
--- a/closed/GensrcJ9JCL.gmk
+++ b/closed/GensrcJ9JCL.gmk
@@ -51,7 +51,7 @@ $(eval $(call SetupCopyFiles,COPY_OVERLAY_FILES, \
 		src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java \
 		src/java.base/share/classes/java/util/zip/ZipFile.java \
 		src/java.base/share/classes/jdk/internal/access/JavaNetInetAddressAccess.java \
-		src/java.base/share/classes/jdk/internal/ref/PhantomCleanable.java \
+		src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java \
 		src/java.base/share/classes/module-info.java \
 		src/java.base/share/classes/sun/security/jca/ProviderConfig.java \
 		src/java.base/share/classes/sun/security/jca/ProviderList.java \
--- a/src/java.base/share/classes/jdk/internal/ref/PhantomCleanable.java
+++ b/src/java.base/share/classes/jdk/internal/ref/PhantomCleanable.java
@@ -23,22 +23,12 @@
  * questions.
  */
 
-/*
- * ===========================================================================
- * (c) Copyright IBM Corp. 2025, 2025 All Rights Reserved
- * ===========================================================================
- */
-
 package jdk.internal.ref;
 
 import java.lang.ref.Cleaner;
 import java.lang.ref.Reference;
 import java.lang.ref.PhantomReference;
 import java.util.Objects;
-
-/*[IF CRIU_SUPPORT]*/
-import openj9.internal.criu.NotCheckpointSafe;
-/*[ENDIF] CRIU_SUPPORT */
 
 /**
  * PhantomCleanable subclasses efficiently encapsulate cleanup state and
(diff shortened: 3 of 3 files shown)


JasonFengJ9:
```
16:23:18  The following files were modified and have incorrect copyrights
[Pipeline] echo
16:23:18  src/java.base/share/classes/jdk/internal/ref/PhantomCleanable.java
16:23:18  	IBM Copyright (basic) is missing from the file
```
The upstream version is restored.

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
```diff
@@ -265,6 +275,9 @@ synchronized void reset() {
          *
          * @return true if the list is empty
          */
+        /*[IF CRIU_SUPPORT]*/
+        @NotCheckpointSafe
```
keithc-ca, line 279: This annotation was also applied to `insert()` [earlier](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/936): what guides the decision where the annotation is needed?

JasonFengJ9: Before JVM enters the single-thread mode and performs a checkpoint, JVM is going to pause other Java threads except the checkpoint thread, however, other threads might hold a lock and can't proceed (and release the lock) in the single-thread mode, in this case, if the checkpoint thread attempts to acquire the lock, a `JVMCheckpointException` is thrown otherwise JVM hangs.

The `@NotCheckpointSafe` annotation is introduced to allow JVM to check if a thread holds a lock before entering the single-thread mode. If there is such a method, JVM waits a timeout, usually the thread finishes and releases the lock, afterwards, the checkpoint can be performed.

In past cases, the last method showing in the stack trace usually is the method to be annotated because they are all synchronized blocks. 
For the failure https://github.com/eclipse-openj9/openj9/issues/21111#issuecomment-2698203372, the non-checkpoint thread entered a synchronized method `CleanerImpl$CleanableList.isEmpty()`, the `@NotCheckpointSafe` annotation should be applied to this method, and allow it finishes before entering the single-thread mode.
The last two failed attempts were adding the annotation to the method hit by the checkpoint method, and won't affect the thread holding the lock`jdk/internal/ref/CleanerImpl$CleanableList@0x000000070A36BCE8`.
The two thread stack traces are at https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/960#issue-2898543766

As discussed a while ago, this annotation is added case by case according to a known checkpoint failure.

keithc-ca: Thanks, but that doesn't really answer my question about why the annotation is not needed for `insert()` - it was previously added in response to some failure scenario: is that scenario not still possible?

JasonFengJ9: `CleanerImpl$CleanableList.insert()` was invoked by the checkpoint thread that is the only thread not paused, there is no `@NotCheckpointSafe` annotation check for this thread.

There are other synchronized methods within CleanerImpl$CleanableList such as `reset()`, `insert()`, and `remove()`.
Not adding annotations for these methods for now since there is no known failure showing these methods holding the lock in a thread other than the checkpoint thread.

keithc-ca: The annotation was added to `insert()` by #936, then removed by #944; #944 is effectively reverted by this so why should the annotation not be restored to `insert()` in this change?

JasonFengJ9: `insert()` was only invoked by the checkpoint thread, which isn't subject to the annotation check. Thus, it won't be checked and has no impact.
`isEmpty()` was invoked by a non-checkpoint thread which holds a lock `jdk/internal/ref/CleanerImpl$CleanableList@0x000000070A36BCE8` and caused the `JVMCheckpointException`.
Adding `@NotCheckpointSafe` annotation at `isEmpty()` allows JVM to detect its presence in the thread stack before entering the single thread mode and wait for a timeout to let this method finish.
Hence only `isEmpty()` needs the annotation for the current failure known.

keithc-ca: In https://github.com/eclipse-openj9/openj9/issues/21111#issuecomment-2698203372 you say
> The @NotCheckpointSafe annotation should be placed at the non-checkpoint thread that entered the lock.

This makes no sense: annotations decorate _methods_, not _threads_.

Earlier you said
> @NotCheckpointSafe should be for the caller method of [the synchronized method]

That's not what you're proposing this time. Why the course change?

JasonFengJ9: > > The @NotCheckpointSafe annotation should be placed at the non-checkpoint thread that entered the lock.

> This makes no sense: annotations decorate methods, not threads.

Correction - The @NotCheckpointSafe annotation should be placed at **a method invoked by** the non-checkpoint thread that entered the lock.

> > @NotCheckpointSafe should be for the caller method of [the synchronized method]

> That's not what you're proposing this time. Why the course change?

That change was invalid since the method affected was invoked by the checkpoint thread.
The past cases requiring `@NotCheckpointSafe` annotation were synchronized blocks hence both the checkpoint thread and non-checkpoint threads share same method.
This failure occurred at a synchronized method, only the method invoked by non-checkpoint thread should be annotated.

keithc-ca: This seems incredibly fragile to me. If a method is used by both the checkpoint thread and by non-checkpoint threads, then the annotation is both needed and not allowed. I don't see how this strategy can work.

JasonFengJ9: > not allowed

No, it is not "_not allowed_" for the checkpoint thread.
The checkpoint thread doesn't need to check the annotation because it always runs without waiting for a timeout. Only the non-checkpoint threads will be checked and be waited if the annotation is detected.

keithc-ca: Ok, I think I better understand the implications of `@NotCheckpointSafe` annotations. I think the annotation should also be applied to `insert()` (as was done in #936) and applied to `remove()`, unless you can offer an explanation for why those annotations are not necessary.

keithc-ca (approved):

keithc-ca:
Validated locally.

PR #919, opened 2025-01-21
Temporarily restore ASM and remove test call to ClearAllFramePops

pshipton:
Restore jdk.internal.org.objectweb in java.base, but don't revert OpenJDK test changes.

Related to
https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/a778fb5131

Issue https://github.com/eclipse-openj9/openj9/issues/20938

Temporarily remove call in test to ClearAllFramePops

Issue https://github.com/eclipse-openj9/openj9/issues/20980

PR #964, opened 2025-03-13
Revert "Temporarily restore ASM until DDR is fixed to use classfile API"

keithc-ca:
This reverts commit ce7d84ce4868c77a1b514d099b189f6c0bb3b775 from #919. ASM is not used now that https://github.com/eclipse-openj9/openj9/pull/21341 is merged.


pshipton:
jenkins compile amac jdknext

PR #965, opened 2025-03-18
Add error visibility for SunPKCS11 provider configuration failures

sarwat12:
When using `SunPKCS11` at the top of the security provider list, if configuration errors are present in the pkcs11 config file (e.g., a typo in the list of `disabledMechanisms{}`), the `SunPKCS11` provider fails to initialize. If the provider is unspecified during the cryptographic operation, the application continues to run without any indication of sunpkcs11 configuration issues, while moving onto the next provider in  list, leading to silent initialization failures.

The lack of error messages makes it difficult for users to diagnose and fix sunpkcs11 related configuration issues. This change modifies the `ConfigurationException` class in [pkcs11/Config.java](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/Config.java) to output error messages when `-Djava.security.debug=sunpkcs11` debug flag is enabled, providing better visibility into initialization failures due to potential errors present in the configuration file.

### DIFF
--- a/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/Config.java
+++ b/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/Config.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * (c) Copyright IBM Corp. 2023, 2025 All Rights Reserved
  * ===========================================================================
  */
 
@@ -37,6 +37,7 @@ import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
 
+import sun.security.util.Debug;
 import sun.security.util.PropertyExpander;
 
 import sun.security.pkcs11.wrapper.*;
@@ -1084,11 +1085,23 @@ final class Config {
 class ConfigurationException extends IOException {
     @Serial
     private static final long serialVersionUID = 254492758807673194L;
+
+    private static final Debug configDebug = Debug.getInstance("sunpkcs11");
+
+    private static void debug(String msg) {
+         // If debugging is enabled, use the Debug class for additional sunpkcs11 logging.
+         if (configDebug != null) {
+             configDebug.println(msg);
+         }
+     }
+
     ConfigurationException(String msg) {
         super(msg);
+        debug(msg);
     }
 
     ConfigurationException(String msg, Throwable e) {
         super(msg, e);
+        debug(msg);
     }
 }


pshipton:
Pls modify the end date in the IBM copyright on line 28.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/Config.java
```diff
@@ -1084,11 +1085,23 @@ private void parseHandleStartupErrors(String keyword) throws IOException {
 class ConfigurationException extends IOException {
     @Serial
     private static final long serialVersionUID = 254492758807673194L;
+
+    static final Debug configDebug = Debug.getInstance("sunpkcs11");
```
keithc-ca, line 1089: This field should be private.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/Config.java
```diff
@@ -1084,11 +1085,23 @@ private void parseHandleStartupErrors(String keyword) throws IOException {
...
+
+    private static void debug(Object o) {
+         // If debugging is enabled, use the Debug class for additional sunpkcs11 logging
+         if (configDebug != null) {
+             configDebug.println(o.toString());
```
keithc-ca, line 1094: Please finish each sentence with a period.
There's no need to assume `o` is not null; suggest the argument should just be a string (matching all the current uses).
```java
    private static void debug(String msg) {
         // If debugging is enabled, use the Debug class for additional sunpkcs11 logging.
         if (configDebug != null) {
             configDebug.println(msg);
         }
     }
```

keithc-ca (changes requested):
Please update the IBM copyright date.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/Config.java
sarwat12: Done.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/Config.java
sarwat12: Argument changed to string.

keithc-ca:
Jenkins compile amac jdknext

keithc-ca:
Please fix the commit message: lines in the body should be no more than 72 characters in length.

sarwat12:
> Please fix the commit message: lines in the body should be no more than 72 characters in length.

Done.

keithc-ca (approved):
Build passed: https://openj9-jenkins.osuosl.org/job/PullRequest-OpenJDKnext/366

pshipton:
In case it's not obvious, this needs to be backported to every version where you want it, i.e. jdk21, jdk17, jdk11.

pshipton:
@keithc-ca fyi the backports.

PR #969, opened 2025-03-21
Avoid checking RestrictedSecurity profile hash during jar verification

KostasTsiounis:
If the process of verifying a jar is started before the `RestrictedSecurity` profile is loaded, the hash calculation is triggered as part of it leading to a nested jar verification and a subsequent error.

To avoid that, the hash calulation of a profile is skipped if triggered by a jar verification process and is performed later in the loading process.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -79,6 +79,8 @@ public final class RestrictedSecurity {
 
     private static RestrictedSecurityProperties restricts;
 
+    private static boolean profileHashChecked = false;
+
     private static final Set<String> unmodifiableProperties = new HashSet<>();
 
     private static final Map<String, List<String>> supportedPlatformsNSS = new HashMap<>();
@@ -244,6 +246,10 @@ public final class RestrictedSecurity {
      */
     public static boolean isServiceAllowed(Service service) {
         if (securityEnabled) {
+            if (!(profileHashChecked || isJarVerifierinStackTrace())) {
+                profileHashChecked = true;
+                checkHashValues();
+            }
             return restricts.isRestrictedServiceAllowed(service, true);
         }
         return true;
@@ -257,6 +263,10 @@ public final class RestrictedSecurity {
      */
     public static boolean canServiceBeRegistered(Service service) {
         if (securityEnabled) {
+            if (!(profileHashChecked || isJarVerifierinStackTrace())) {
+                profileHashChecked = true;
+                checkHashValues();
+            }
             return restricts.isRestrictedServiceAllowed(service, false);
         }
         return true;
@@ -270,6 +280,10 @@ public final class RestrictedSecurity {
      */
     public static boolean isProviderAllowed(String providerName) {
         if (securityEnabled) {
+            if (!(profileHashChecked || isJarVerifierinStackTrace())) {
+                profileHashChecked = true;
+                checkHashValues();
+            }
             // Remove argument, e.g. -NSS-FIPS, if present.
             int pos = providerName.indexOf('-');
             if (pos >= 0) {
@@ -289,6 +303,10 @@ public final class RestrictedSecurity {
      */
     public static boolean isProviderAllowed(Class<?> providerClazz) {
         if (securityEnabled) {
+            if (!(profileHashChecked || isJarVerifierinStackTrace())) {
+                profileHashChecked = true;
+                checkHashValues();
+            }
             String providerClassName = providerClazz.getName();
 
             // Check if the specified class extends java.security.Provider.
@@ -376,6 +394,18 @@ public final class RestrictedSecurity {
                 profileID = defaultMatch;
             }
         }
+    }
+
+    private static boolean isJarVerifierinStackTrace() {
+        StackTraceElement[] elements = Thread.currentThread().getStackTrace();
+        for (int i = 1; i < elements.length; i++) {
+            StackTraceElement stackTraceElement = elements[i];
+            if ("java.util.jar.JarVerifier".equals(stackTraceElement.getClassName())
+                && "java.base".equals(stackTraceElement.getModuleName())) {
+                return true;
+            }
+        }
+        return false;
     }
 
     private static void checkIfKnownProfileSupported() {
--- a/src/java.base/share/classes/sun/security/jca/Providers.java
+++ b/src/java.base/share/classes/sun/security/jca/Providers.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * (c) Copyright IBM Corp. 2024, 2025 All Rights Reserved
  * ===========================================================================
  */
 
@@ -111,7 +111,6 @@ public class Providers {
         // triggers a getInstance() call (although that should not happen)
         providerList = ProviderList.EMPTY;
         providerList = ProviderList.fromSecurityProperties();
-        RestrictedSecurity.checkHashValues();
     }
 
     // Return Sun provider.


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -378,6 +396,18 @@ private static void getProfileID(Properties props) {
...
+        StackTraceElement[] elements = Thread.currentThread().getStackTrace();
+        for (int i = 1; i < elements.length; i++) {
+            StackTraceElement stackTraceElement = elements[i];
+            if ("java.util.jar.JarVerifier".equals(stackTraceElement.getClassName())
+                && "java.base".equals(stackTraceElement.getModuleName())) {
+                return true;
+            }
+        }
+        return false;
+    }
```
keithc-ca, line 409: This should use `StackWalker` to avoid producing the whole stack trace at once.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -79,6 +79,8 @@ public final class RestrictedSecurity {
 
     private static RestrictedSecurityProperties restricts;
 
+    private static boolean profileHashChecked = false;
```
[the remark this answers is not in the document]
KostasTsiounis: I wanted to use this additional variable to short-circuit the call to `isJarVerifierinStackTrace()` so that it's not called every time. `profileParser != null` indicates that the hash needs to be checked, but there is no way to tell if the hash has already been checked. With this extra variable, and after the hashes have been checked, the expression `!(profileHashChecked || isJarVerifierinStackTrace())` evaluates to `false` without calling `isJarVerifierinStackTrace()`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -244,6 +246,10 @@ public static boolean isFIPSEnabled() {
      */
     public static boolean isServiceAllowed(Service service) {
         if (securityEnabled) {
+            if (!(profileHashChecked || isJarVerifierinStackTrace())) {
+                profileHashChecked = true;
+                checkHashValues();
+            }
```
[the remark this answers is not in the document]
KostasTsiounis: Due to the comment above, moving the `isJarVerifierinStackTrace()` into the `checkHashValues()` will cause it to be executed every time. The `checkHashValues()` will also be executed every time someone tries to check a service or a provider, which is not what we want. We only want this to run once but not while verifying a jar.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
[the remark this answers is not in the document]
keithc-ca: `profileParser` is non-null exactly when the check is required:
```java
    private static void checkHashValues() {
        if ((profileParser != null) && !isJarVerifierInStackTrace()) {
            profileParser.checkHashValues();
            profileParser = null;
        }
    }
```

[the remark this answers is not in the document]
KostasTsiounis: This will lead to an infinite loop. Initiating the `checkHashValues()` will trigger another `canServiceBeRegistered()` or `isServiceAllowed()` which will trigger another `checkHashValues()` etc. I had the same issue when I was setting the `profileHashChecked = true` after calling `checkHashValues()` but I cannot set `profileParser = null` before since I need it to check.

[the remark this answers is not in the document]
keithc-ca: Infinite recursion can be avoided with a local, clearing the field before checking hash values:
```java
        if ((parser != null) && !isJarVerifierInStackTrace()) {
            profileParser = null;
            parser.checkHashValues();
        }
```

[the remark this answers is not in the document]
KostasTsiounis: Sure. I just thought having a separate variable with a name with a clear meaning would be better for readability, given that it's just a boolean value.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -257,6 +249,11 @@ public static boolean isServiceAllowed(Service service) {
      */
     public static boolean canServiceBeRegistered(Service service) {
         if (securityEnabled) {
+            ProfileParser parser = profileParser;
+            if ((parser != null) && !isJarVerifierInStackTrace()) {
+                profileParser = null;
+                parser.checkHashValues();
+            }
```
keithc-ca, line 256: Please keep the `checkHashValues()` method (making it `private`) with this code rather than repeating it.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -378,6 +385,15 @@ private static void getProfileID(Properties props) {
         }
     }
 
+    private static boolean isJarVerifierInStackTrace() {
+        return StackWalker.getInstance(Option.RETAIN_CLASS_REFERENCE)
+                          .walk(sf -> sf.map(s -> s.toStackTraceElement())
```
keithc-ca, line 390: Please use uppercase consistently: "In" instead of "in".
Please qualify `StackWalker.Option` rather than importing `Option`.
Suggest mapping to `Class` rather than `StackTraceElement`.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Kept method, turned it private and put this code in there.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Avoided importing `StackWalker.Option` and changed the command to get and use `Class` instead of `StackTraceElement`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -378,6 +383,15 @@ private static void getProfileID(Properties props) {
...
     }
 
+    private static boolean isJarVerifierInStackTrace() {
+        return StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE)
+                          .walk(sf -> sf.map(s -> s.getDeclaringClass())
+                                        .anyMatch(s -> ("java.util.jar.JarVerifier".equals(s.getName())
+                                                         && "java.base".equals(s.getModule().getName()))
+                                                 )
+                               );
+    }
```
keithc-ca, line 393: Suggest a better location for this method is near the one place it is used.

The return expression is hard to follow expressed like this.

Better names might help: `sf` has type `Stream<StackFrame>` (not `StackFrame` like one might first think; the `s` should not be used for two different types (the first is `StackFrame`, the second is `Class`).

It might also help to decompose it into smaller pieces, perhaps like this:
```java
        java.util.function.Predicate<Class<?>> isJarVerifier =
                clazz -> "java.util.jar.JarVerifier".equals(clazz.getName())
                      && "java.base".equals(clazz.getModule().getName());

        java.util.function.Function<Stream<StackWalker.StackFrame>, Boolean> matcher =
                stream -> stream.map(StackWalker.StackFrame::getDeclaringClass)
                                .anyMatch(isJarVerifier);

        return StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).walk(matcher);
```

KostasTsiounis: Moved method and changed to suggestion.

keithc-ca:
Please squash.

keithc-ca:

keithc-ca (approved):

PR #980, opened 2025-04-03
Rename bundled openssl libraries

jasonkatonica:
OpenSSL libraries that are bundled within the runtime are being renamed to avoid conflicts with system libraries that may have the same name. Additional updates were made to the loading logic associated with native cryptography to make use of the correct bundled version by default.

### DIFF
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -662,6 +662,17 @@ find_crypto_library(jboolean traceEnabled, const char *chomepath)
 #endif /* defined(_AIX) */
     };
 
+    /** OpenSSL library names associated with bundled library. */
+#if defined(_AIX)
+    static const char bundledLibName[] = "libcrypto-semeru.so";
+#elif defined(__APPLE__) /* defined(_AIX) */
+    static const char bundledLibName[] = "libcrypto-semeru.dylib";
+#elif defined(_WIN32) /* defined(__APPLE__) */
+    static const char bundledLibName[] = "crypto-semeru.dll";
+#else /* defined(_WIN32) */
+    static const char bundledLibName[] = "libcrypto-semeru.so";
+#endif /* defined(_AIX) */
+
     const size_t numOfLibs = sizeof(libNames) / sizeof(libNames[0]);
     void *result = NULL;
     size_t i = 0;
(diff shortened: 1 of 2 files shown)


jasonkatonica:
This pull request has been opened as a draft since an additional update to the JIT server code is required to avoid failure. This update is expected to be done in openj9 project and I will update this PR once it is available. This update is still ready for review at this time.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -662,6 +662,17 @@ find_crypto_library(jboolean traceEnabled, const char *chomepath)
 #endif /* defined(_AIX) */
     };
 
+    /** OpenSSL library names associated with bundled library. */
+#if defined(_AIX)
+    static const char * const bundledLibName = "libcrypto-semeru.so";
```
keithc-ca, line 667: Each definition of `` should be an array:
```c
    static const char bundledLibName[] = "libcrypto-semeru.so";
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -693,33 +704,31 @@ find_crypto_library(jboolean traceEnabled, const char *chomepath)
...
+                fprintf(stderr, "\tFailed to allocate memory for file name with path.\n");
             }
-            result = load_crypto_library(traceEnabled, libNameWithPath);
+            free(libPath);
+            return NULL;
+        }
 
-            free(libNameWithPath);
+        strcpy(libNameWithPath, libPath);
+        strcat(libNameWithPath, bundledLibName);
```
keithc-ca, line 720: Just one allocation is needed; line 689 should become
```c
        size_t path_len = strlen(chomepath) + sizeof(pathSuffix) - 1 + sizeof(bundledLibName) - 1;
```
and this should be:
```c
        strcat(libPath, bundledLibName);
```
Then line 728 should be removed.

keithc-ca (changes requested):

jasonkatonica:
Updated and tested latest code in this PR. This update was moved from `Draft` status to `Ready for review` since the dependent update in j9 has been merged for the JIT server [as seen here](https://github.com/eclipse-openj9/openj9/pull/21566).

keithc-ca:
Please squash.

jasonkatonica:
> Please squash.

Squashed

keithc-ca:

keithc-ca (approved):

PR #989, opened 2025-04-14
Add flag to skip loading the library bundled with the JDK

KostasTsiounis:
The `-Djdk.native.openssl.lib.skipBundled` flag is added to allow users to indicate that they want to skip the bundled version of `OpenSSL` and look for one in their system.


closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -90,14 +90,23 @@ private static long loadCryptoLibraries() {
             // Load jncrypto JNI library.
             System.loadLibrary("jncrypto");
 
+            // Get user-specified option to skip bundled OpenSSL library.
+            boolean skipBundledLib = System.getProperty("jdk.native.openssl.lib.skipBundled") != null;
```
keithc-ca, line 94: Suggest the name should be abbreviated to `skipBundled`, like the comment in the native code.
This should use `getBoolean()` to allow a command-line option to override environment variables:
```java
            boolean skipBundled = Boolean.getBoolean("jdk.native.openssl.lib.skipBundled");
```

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -90,14 +90,23 @@ private static long loadCryptoLibraries() {
...
+
             // Get user-specified OpenSSL library to use, if available.
-            String nativeLibName = System.getProperty("jdk.native.openssl.lib", "");
+            String nativeLibName = System.getProperty("jdk.native.openssl.lib");
+
+            // Check that these mutually exclusive flags are not used at the same time.
+            if (skipBundledLib && (nativeLibName != null)) {
+                throw new RuntimeException("You cannot specify 'jdk.native.openssl.lib.skipBundled'" +
+                        " and 'jdk.native.openssl.lib' at the same time");
+            }
```
keithc-ca, line 103: Is this necessary? We could just document that `jdk.native.openssl.lib` is ignore if `jdk.native.openssl.lib.skipBundled=true`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -611,11 +611,12 @@ load_crypto_library(jboolean traceEnabled, const char *libName)
 }
 
 /* Look for a crypto library in java.home or the system.
+ * Skip looking in java.home when skipBundled is true.
  * NULL is returned when an appropriate crypto library
  * cannot be found.
  */
 static void *
-find_crypto_library(jboolean traceEnabled, const char *chomepath)
+find_crypto_library(jboolean traceEnabled, jboolean skipBundledLib, const char *chomepath)
```
keithc-ca, line 619: The name `skipBundledLib` should be `skipBundled` throughout (see line 614).

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -678,52 +679,58 @@ find_crypto_library(jboolean traceEnabled, const char *chomepath)
     size_t i = 0;
     long tempVersion = 0;
 
-    /* If JAVA_HOME is not null or empty and no library has been loaded yet, try there. */
-    if ((NULL != chomepath) && ('\0' != *chomepath) && (NULL == crypto_library)) {
+    if (JNI_TRUE == skipBundledLib) {
```
keithc-ca, line 682: A boolean variable should be just used directly when testing if is true:
```c
    if (skipBundled) {
```

keithc-ca (changes requested):

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: I will change the name, but we wanted to just look for the presence of the option, not `skipBundled=true`.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: We discussed this with @mstoodle and @jasonkatonica and we decided we want to fail in this scenario.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: We've used this pattern before since it's a `jboolean`, not a `boolean`, but I can change that if you feel it's not needed.

keithc-ca: I consider it an anti-pattern. Reasonable uses of boolean variables are:
```c
    if (flag) { ...}
    if (!flag) { ...}
    if (JNI_FALSE == flag) { ...}
```
Problems arise when you consider `else` blocks:
```c
   if (JNI_TRUE == flag) { yes(); } else { no(); }
```
is not the same as
```c
   if (flag) { yes(); } else { no(); }
```
because `jboolean` variables are represented using a type with more than one bit (meaning there are values possible other than `JNI_FALSE` and `JNI_TRUE` which should be treated consistently as _true_).

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
keithc-ca: As we discussed off-line, that presents a usability challenge; there is no way to unset a system property via a command-line option, so if it's set (to anything) in an environment variable the only way to cancel that is to change the environment variable. My suggestion would allow someone to say `-D...skipBundled=false` on the command line independently of what any environment variable might say.

I'd also note it's inconsistent with existing system properties (like `semeru.fips` and `semeru.fips.allowsetproperties`).

KostasTsiounis: I abbreviated the name and changed the option to look for `=true` instead of just the presence of it.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: Do you think I should just change to `if (skipBundled)` or would it be better if I inverted the logic to use `if (JNI_FALSE == skipBundled)`?

keithc-ca: I don't think inverting that `if` statement improves it. A rule-of-thumb I like to follow is to put the smaller branch first so when you're reading code after the `else` keyword you are more likely to be able to see the condition (that's false) there. I stand by my original suggestion (just remove `JNI_TRUE ==`).

KostasTsiounis: Changed to suggestion.

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -90,14 +90,23 @@ private static long loadCryptoLibraries() {
...
+            boolean skipBundled = Boolean.getBoolean("jdk.native.openssl.skipBundled");
+
             // Get user-specified OpenSSL library to use, if available.
-            String nativeLibName = System.getProperty("jdk.native.openssl.lib", "");
+            String nativeLibName = System.getProperty("jdk.native.openssl.lib");
+
+            // Check that these mutually exclusive flags are not used at the same time.
+            if (skipBundled && (nativeLibName != null)) {
+                throw new RuntimeException("You cannot specify 'jdk.native.openssl.lib.skipBundled'" +
+                        " and 'jdk.native.openssl.lib' at the same time");
```
keithc-ca, line 102: The message should refer the actual property name; suggest alternate wording without the single quotes:
```java
                throw new RuntimeException("Conflicting properties " +
                        "jdk.native.openssl.skipBundled and jdk.native.openssl.lib");
```

keithc-ca (changes requested):

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: Changed to suggestion.

keithc-ca:
Please squash.

keithc-ca:
Jenkins compile amac jdknext

keithc-ca (approved):

PR #991, opened 2025-04-16
Check RestrictedSecurity profile hash after Providers init

KostasTsiounis:
If the `Providers` class hasn't had a chance to initialize, the required message digest algorithms are not available to check the hash value of the `RestrictedSecurity` profile.

With this change, the check is only allowed after the appropriate initialization has been performed.

Issue https://github.com/eclipse-openj9/openj9/issues/21615

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -77,6 +77,8 @@ public final class RestrictedSecurity {
 
     private static ProfileParser profileParser;
 
+    private static boolean enableCheckHashes;
+
     private static RestrictedSecurityProperties restricts;
 
     private static final Set<String> unmodifiableProperties = new HashSet<>();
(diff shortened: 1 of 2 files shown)


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -180,9 +182,20 @@ private static boolean isJarVerifierInStackTrace() {
...
+
         ProfileParser parser = profileParser;
-        if ((parser != null) && !isJarVerifierInStackTrace()) {
+        if ((parser != null)
+                && enableCheckHashes
+                && !isJarVerifierInStackTrace()
+        ) {
             profileParser = null;
             parser.checkHashValues();
         }
```
keithc-ca, line 201: Suggest restructuring this:
* if `parser` is null, there's no need to do anything (this will be the vast majority of calls)
* we expect at most one call where `fromProviders` is true, so checking `!enableCheckHashes` isn't helpful
```java
        if (parser != null) {
            if (fromProviders) {
                enableCheckHashes = true;
            }
            if (enableCheckHashes && !isJarVerifierInStackTrace()) {
                profileParser = null;
                parser.checkHashValues();
            }
        }
```

keithc-ca:
Please squash.

keithc-ca:

keithc-ca (approved):

PR #997, opened 2025-04-28
Add tabindex to grid columns which has only plain text to make it tababble

psoujany:
This PR fixes the accessibility violations seen with achecker tool.

```
Message: The scrollable element <div> with non-interactive content is not tabbable
 Level: violation
 XPath: /html[1]/body[1]/div[1]/main[1]/div[2]/div[2]/div[1]/div[3122]
 Snippet: <div class="col-last odd-row-color all-classes-table all-classes-table-tab2">
 Help: https://able.ibm.com/rules/archives/latest/doc/en-US/element_scrollable_tabbable.html#%7B%22message%22%3A%22The%20scrollable%20element%20%3Cdiv%3E%20with%20non-interactive%20content%20is%20not%20tabbable%22%2C%22snippet%22%3A%22%3Cdiv%20class%3D%5C%22col-last%20odd-row-color%20all-classes-table%20all-classes-table-tab2%5C%22%3E%22%2C%22value%22%3A%5B%22VIOLATION%22%2C%22FAIL%22%5D%2C%22reasonId%22%3A%22fail_scrollable%22%2C%22ruleId%22%3A%22element_scrollable_tabbable%22%7D
```


psoujany:
@pshipton As per your suggestion here https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/472, raising this PR. Thank you.

psoujany:
Once the PR is merged, will backport the changes to 24/21/17 repos. Thank you.

psoujany:
Personal build is successful with the changes and violation(The scrollable element <div> with non-interactive content is not tabbable) is not seen with the changes in accessibility testing.

@smoinud1 FYI

keithc-ca (changes requested):

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/Table.java
```diff
@@ -86,6 +92,8 @@ public class Table<T> extends Content {
     private HtmlStyle gridStyle;
     private final List<Content> bodyRows;
     private HtmlId id;
+    private static String regex = "<(?:a|area|button|input|object|select|textarea)\\b";
+    private static final Pattern pattern = Pattern.compile(regex);
```
keithc-ca, line 96: Please inline `regex` into the pattern initial value, and use a more descriptive name than just `pattern`.

keithc-ca (changes requested):
Please avoid "tababble" - if it's a word, it would end with "able" (just one "b", not two).

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/Table.java
psoujany: Updated changes as suggested. Thank you/

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/Table.java
```diff
@@ -326,10 +333,19 @@ public void addRow(T item, List<Content> contents) {
...
+                cell.put(HtmlAttr.ROLE, "tablist")
+                    .put(HtmlAttr.TABINDEX, "0");
+            }
 
             for (String tabClass : tabClasses) {
                 cell.addStyle(tabClass);
+                if (!matchFound) {
+                    cell.put(HtmlAttr.ROLE, "tablist")
+                        .put(HtmlAttr.TABINDEX, "0");
+                }
```
keithc-ca, line 348: These lines are redundant; see lines 338-341.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/Table.java
```diff
@@ -326,10 +333,19 @@ public void addRow(T item, List<Content> contents) {
             HtmlStyle cellStyle = columnStyles.get(colIndex);
             // Always add content to make sure the cell isn't dropped
             var cell = HtmlTree.DIV(cellStyle).addUnchecked(c.isEmpty() ? Text.EMPTY : c);
+            boolean matchFound = c.isEmpty() || checkFormElements.matcher(c.toString()).find();
             cell.addStyle(rowStyle);
+            if (!matchFound) {
+                cell.put(HtmlAttr.ROLE, "tablist")
+                    .put(HtmlAttr.TABINDEX, "0");
+            }
```
keithc-ca, line 341: This seems to apply to many more places than the title suggests - I don't think this is what we want.

keithc-ca (changes requested):

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/Table.java
psoujany: Yes Keith, here we have 2 styles.
1. rowStyle without tabClasses : where div is `<div class="col-first even-row-color">` 
2. rowStyle with tabClasses : where div is <div class="col-last odd-row-color all-classes-table all-classes-table-tab2">

For both the styles tab should get focus, to do that we need to add `role=tablist tabindex=0`.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/Table.java
psoujany: These values should be applied to all grid rows with/without tab elements which has only plain text.

PR #998, opened 2025-04-30
Prevent AES/CBC native buffer overflow

pshipton:
Native AES acceleration is available when using AES and CBC mode of operation. While calling the underlying OpenSSL library operations often do not supply a large enough buffer which is documented by the OpenSSL EVP_EncryptUpdate as input length + an additional block size. This update ensures buffers sent to this API are large enough.


keithc-ca:

keithc-ca:
Test failures for sanity.openjdk seem unrelated.

keithc-ca (approved):

PR #1004, opened 2025-05-05
Target z10 architecture for OpenSSL Linux on z

jasonkatonica:
This update is to prepare for migrating to `OpenSSL` version `3.5.0`.
`OpenSSL` requires `z10` or higher architectures specified for the
compiler.

### DIFF
--- a/closed/openssl.gmk
+++ b/closed/openssl.gmk
@@ -92,6 +92,11 @@ ifeq (,$(OPENSSL_TARGET))
   $(error Unsupported platform $(OPENJDK_TARGET_OS)-$(OPENJDK_TARGET_CPU))
 endif # OPENSSL_TARGET
 
+OPENSSL_CFLAGS :=
+ifeq ($(OPENJDK_TARGET_CPU), s390x)
+  OPENSSL_CFLAGS := -march=z10
+endif
+
 ifneq (,$(CCACHE))
   # If ccache is enabled and the environment contains either CC or CXX, their
   # values (as defined in this make) are propagated to the instance of make
@@ -101,8 +106,8 @@ ifneq (,$(CCACHE))
 endif # CCACHE
 
 build_openssl :
-	@$(ECHO) Compiling OpenSSL in $(OPENSSL_DIR) for $(OPENSSL_TARGET)
-	( $(OPENSSL_CONFIG_SETUP) $(CD) $(OPENSSL_DIR) && $(PERL) Configure $(OPENSSL_TARGET) shared )
+	@$(ECHO) Compiling OpenSSL in $(OPENSSL_DIR) for $(OPENSSL_TARGET) with additional CFLAGS $(OPENSSL_CFLAGS)
+	( $(OPENSSL_CONFIG_SETUP) $(CD) $(OPENSSL_DIR) && $(PERL) Configure $(OPENSSL_CFLAGS) $(OPENSSL_TARGET) shared )
 	$(OPENSSL_PATCH)
 	( $(OPENSSL_MAKE_SETUP) $(CD) $(OPENSSL_DIR) && $(OPENSSL_MAKE) )
 


pshipton:
@joransiu @r30shah fyi

jasonkatonica:
The error message and discussion associated with the openssl failure I was seeing on s390x linux can be seen [here](https://github.com/openssl/openssl/issues/27323)

closed/openssl.gmk
```diff
@@ -92,6 +92,11 @@ ifeq (,$(OPENSSL_TARGET))
   $(error Unsupported platform $(OPENJDK_TARGET_OS)-$(OPENJDK_TARGET_CPU))
 endif # OPENSSL_TARGET
 
+OPENSSL_CFLAGS :=
+ifeq ($(OPENJDK_TARGET_CPU), s390x)
+  OPENSSL_CFLAGS := -march=z196
```
keithc-ca, line 97: Why not `z10` like [flags-cflags.m4](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/db84234aaa983a6b9f271b76d60dbb4ceb2e7db6/make/autoconf/flags-cflags.m4#L801)?

jasonkatonica: I originally tested `z10` as the arch and that does work fine. I ended up later choosing `z196` since this is the oldest z architecture listed on our support page https://www.ibm.com/support/pages/semeru-runtimes-support . I could set this to `z10` to be safe?

r30shah: Looking at the supported distros for IBM Semeru, I think we should reevaluate the table in https://www.ibm.com/support/pages/semeru-runtimes-support table), as some of the distros even moved from z196 and need min. zEC12 [1] (For SLES 15SP5). For Ubuntu 20.04 it seems like min. supported hardware level is z13 [2] and for RHEL 8 and RHEL 9 it is z13 and z14 respectively.

I do not want to start the discussion of min. supported ALS for IBM Semeru on Z here, but to add to the level set here and question, Keith asked -  I can't think of any other reason to keep this to z10, and based on the finding, setting this to z196 is actually fine (May be zEC1 if we end up with updating our support documentation). 
@joransiu what do you think ?


[1]. https://documentation.suse.com/sles/15-SP5/html/SLES-all/cha-zseries.html
[2]. https://wiki.ubuntu.com/FocalFossa/ReleaseNotes

keithc-ca: According to [1], the `cijne` instruction is supported in z10, so specifying `z196` seems unnecessary.

I think it should be a separate discussion whether we update the minimum supported architecture in openj9, omr and the extensions repositories.

[1] https://www.ibm.com/docs/en/hla-and-tf/1.6.0?topic=instructions-table-all-supported

jasonkatonica: For now I changed the target to z10 to match the same spot [flags-cflags.m4](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/db84234aaa983a6b9f271b76d60dbb4ceb2e7db6/make/autoconf/flags-cflags.m4#L801) that Keith sent. 

It prob does make sense to figure out what arch we want to target for the entire JDK and move both the JDK and the OpenSSL targets accordingly. I do agree that this seems like a separate discussion.

joransiu: Yeah, I agree with @r30shah's assessment that moving to z196 is fine. That said, instead of proactively bumping up OpenSSL to use z196, we should make the same change across OMR, OpenJ9 and extensions repo components for consistency.

keithc-ca: Please update the commit message and the description to reflect the choice of `z10` instead of `z196`.

jasonkatonica: Updated the PR title, PR description, and commit text to represent targeting of z10.

joransiu (approved):

keithc-ca (approved):

PR #1006, opened 2025-05-06
Tolerate RSASSA-PSS in OpenJCEPlus signature

JinhangZhang:
The mask generation function (MGF) based on a hash algorithm is recommended to use the same hash function as the hash function fingerprinting the message for RSASSA-PSS. However, the structures in [PKCS#1v2.1] allow for separate parameterization of the MGF and the message digest.

OpenJCEPlus restricts to use the same hash algorithm for both MGF and message digest. This will cause test failures since the tests are expecting different MGF and message digest behavior.

### DIFF
--- a/test/jdk/javax/xml/crypto/dsig/SecureValidation.java
+++ b/test/jdk/javax/xml/crypto/dsig/SecureValidation.java
@@ -31,6 +31,13 @@
  *          java.base/sun.security.x509
  * @run main/othervm SecureValidation
  */
+
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2025, 2025 All Rights Reserved
+ * ===========================================================================
+ */
+
 import jdk.test.lib.Asserts;
 import jdk.test.lib.security.XMLUtils;
 import jdk.test.lib.Utils;
@@ -48,6 +55,7 @@ import javax.xml.xpath.XPath;
 import javax.xml.xpath.XPathConstants;
 import javax.xml.xpath.XPathFactory;
 import java.security.PrivateKey;
+import java.security.Signature;
 import java.security.cert.X509Certificate;
 import java.security.spec.MGF1ParameterSpec;
 import java.security.spec.PSSParameterSpec;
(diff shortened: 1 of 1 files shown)


keithc-ca:
Is there a compelling argument in favour of this restriction in OpenJCEPlus? It seems somewhat arbitrary and unnecessary to me.

JinhangZhang:
> Is there a compelling argument in favour of this restriction in OpenJCEPlus? It seems somewhat arbitrary and unnecessary to me.

`engineSetParameter` in OpenJCEPlus has the restriction to check if the messagedigest used for mgf is the same as messagedigest used for pss. If they are not the same, then an exception will be thrown which cause this test case failed with OpenJCEPlus provider.

keithc-ca:
I understand OpenJCEPlus **_has_** that restriction: I was asking **_why_**.

JinhangZhang:
> I understand OpenJCEPlus **_has_** that restriction: I was asking **_why_**.

The original PSS from OpenJCEPlus is designed to follow the RFC doc. According to this [docs](https://datatracker.ietf.org/doc/html/rfc3447#page-29), it is recommend to use a same hash function. Therefore, the code in OpenJCEPlus expects them to match.
```
Therefore, it is recommended that the EMSA-
   PSS mask generation function be based on the same hash function.  In
   this manner the entire encoded message will be dependent on the hash
   function and it will be difficult for an opponent to substitute a
   different hash function than the one intended by the signer.  This
   matching of hash functions is only for the purpose of preventing hash
   function substitution, and is not necessary if hash function
   substitution is addressed by other means (e.g., the verifier accepts
   only a designated hash function)
```

FYI, @jasonkatonica

keithc-ca:
That same RFC also says:

> The provable security of RSASSA-PSS does not rely on
> the hash function in the mask generation function being the same as
> the hash function applied to the message.

As I understand it, OpenJCEPlus is intended to augment or replace algorithms also implemented by OpenJDK; replacements that come with extra restrictions are not simple substitutions. Rather than continue to modify tests here to tolerate those restrictions (and hide issues users may encounter), I suggest the better course of action is to remove those restrictions that don't have strong justifications.

@jasonkatonica Please share your thoughts.

jasonkatonica:
We did have some discussions on this topic and decided to keep this behavior the same as z/OS and IBM Java 8, mainly based on the RFC segment above that Jinhang pasted where this is recommended anyway.

test/jdk/javax/xml/crypto/dsig/SecureValidation.java
```diff
@@ -70,10 +78,24 @@ public static void main(String[] args) throws Exception {
...
-                .sm(SignatureMethod.RSA_PSS, new RSAPSSParameterSpec(pspec))
-                .sign(doc);
+                .sm(SignatureMethod.RSA_PSS, new RSAPSSParameterSpec(pspec));
+        Document signed;
+        try {
+            signed = signer.sign(doc);
+        } catch (javax.xml.crypto.dsig.XMLSignatureException xmlse) {
+            Throwable cause = xmlse.getCause();
+            if (cause instanceof java.security.InvalidAlgorithmParameterException) {
+                if (Signature.getInstance("RSA-PSS").getProvider().getName().equals("OpenJCEPlus")
```
keithc-ca, line 90: Why does this check `RSA-PSS` instead of `RSASSA-PSS`?

keithc-ca: The commit message should also be consistent, referring to `RSASSA-PSS` instead.

keithc-ca: Please squash. The commit message should refer to `RSASSA-PSS`, not `RSA-PSS`; likewise for the description here.

JinhangZhang: Squashed

keithc-ca (approved):
Tested in https://openj9-jenkins.osuosl.org/job/Grinder/4329.

PR #1010, opened 2025-05-12
Use CXX_VERSION_STRING instead of COMPILER_VERSION_STRING

pshipton:
COMPILER_VERSION_STRING seems an internal variable that shouldn't be used.


closed/autoconf/custom-hook.m4
```diff
@@ -574,7 +574,7 @@ AC_DEFUN([OPENJ9_PLATFORM_SETUP],
 
   AC_SUBST(OPENJ9_BUILDSPEC)
   AC_SUBST(OPENJ9_PLATFORM_CODE)
-  AC_SUBST(COMPILER_VERSION_STRING)
+  AC_SUBST(CXX_VERSION_STRING)
   AC_SUBST(OPENJ9_LIBS_SUBDIR)
   AC_SUBST(OMR_MIXED_REFERENCES_MODE)
```
keithc-ca, line 579: While you're here, could you put these in alphabetical order?

keithc-ca (changes requested):

closed/autoconf/custom-hook.m4
pshipton: Done.

keithc-ca:
Checking `openj9_version_info.h`, I notice that the version string is longer than I expected (although this didn't make it that way):
```c
#define J9COMPILER_VERSION_STRING "Apple clang version 15.0.0 (clang-1500.1.0.2.5) Target: arm64-apple-darwin22.6.0 Thread model: posix InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin"
```
I think some of that could be omitted (e.g. "Thread model", and "InstalledDir"):
```c
#define J9COMPILER_VERSION_STRING "Apple clang version 15.0.0 (clang-1500.1.0.2.5) Target: arm64-apple-darwin22.6.0"
```
Thoughts?

keithc-ca (approved):

PR #1011, opened 2025-05-13
Properly check if jdk.native.openssl.lib property is set

KostasTsiounis:
Whether the property has been set at all needs to be checked before checking if it's not empty.

Related to issue https://github.com/eclipse-openj9/openj9/issues/21867


keithc-ca:

keithc-ca (approved):

PR #1016, opened 2025-05-20
Add key algorithm check for native CBC

JinhangZhang:
Native implementation of CBC mode is used as default while initializing an AES cipher instance. A secret key will be used to initialize an AES cipher instance. However, there is a lack of check to see if the implementation algorithm corresponding to the secret key is consistent with the cipher's algorithm.

Issue: [#21886](https://github.com/eclipse-openj9/openj9/issues/21886).


keithc-ca (approved):
Validated locally.

PR #1017, opened 2025-05-21
Update NullStringToKey test

taoliult:
The NullStringToKey test attempts to catch the expected IllegalArgumentException by removing the SUN and SunJCE providers. However, the service remains unexpectedly available through SunPKCS11-NSS-FIPS, OpenJCEPlus, and OpenJCEPlusFIPS, causing the test to fail. This update modifies the test to remove these additional providers during execution.

### DIFF
--- a/test/jdk/sun/security/krb5/NullStringToKey.java
+++ b/test/jdk/sun/security/krb5/NullStringToKey.java
@@ -21,6 +21,11 @@
  * questions.
  */
 /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2025, 2025 All Rights Reserved
+ * ===========================================================================
+ */
+/*
  * @test
  * @bug 8343622
  * @summary KerberosKey created with null key bytes
@@ -35,8 +40,11 @@ import java.util.List;
 public class NullStringToKey {
     public static void main(String[] args) throws Exception {
 
+        Security.removeProvider("OpenJCEPlus");
+        Security.removeProvider("OpenJCEPlusFIPS");
         Security.removeProvider("SUN");
         Security.removeProvider("SunJCE");
+        Security.removeProvider("SunPKCS11-NSS-FIPS");
 
         var name = new KerberosPrincipal("me@ME.COM");
         var pass = "password".toCharArray();


test/jdk/sun/security/krb5/NullStringToKey.java
```diff
@@ -35,8 +35,11 @@
 public class NullStringToKey {
     public static void main(String[] args) throws Exception {
 
+        Security.removeProvider("OpenJCEPlus");
+        Security.removeProvider("OpenJCEPlusFIPS");
         Security.removeProvider("SUN");
         Security.removeProvider("SunJCE");
+        Security.removeProvider("SunPKCS11-NSS-FIPS");
```
keithc-ca, line 42: This file should have an IBM copyright notice.

Should we remove all providers instead of a fixed list of names?

keithc-ca (changes requested):

test/jdk/sun/security/krb5/NullStringToKey.java
taoliult: Copyright, updated.

The original test case was designed to trigger an `IllegalArgumentException` when testing the Kerberos-related module. To achieve this, the `SUN` and `SunJCE` providers were removed during testing. However, the expected services remained unexpectedly available through other providers, `SunPKCS11-NSS-FIPS`, `OpenJCEPlus`, and `OpenJCEPlusFIPS` when running the FIPS tests, causing the test to fail. To fix this, we also removed those providers. 

Since we're not entirely certain which provider the test case was originally intended to target, and to keep the test case structure to avoid potential conflicts if it's updated in the future, we believe that removing the unexpected providers is a better way.

keithc-ca:

keithc-ca (approved):
Test failures are unrelated to this change.

taoliult:
Since this test does not exist in JDK 21 or earlier versions, there is no need to back port it to JDK 21 or earlier versions.

taoliult:
The test has been backported to JDK 11, create the PR to backport this corresponding updates.

PR #1020, opened 2025-05-23
Improve OpenSSL build integration

keithc-ca:
Instead of yes/no, use true/false for make variable values like other component flags.

Correct a reference to the `--enable-openssl-bundling` option.

Allow building openssl in parallel.


pshipton:
jenkins compile amac jdknext

pshipton (approved):

PR #1023, opened 2025-05-27
Fix thread interrupt status

tajila:
It is possible for the Java and native interrupt status to get out of
sync. Sending two interrupts to the same thread in quick succession can
cause this. It is critical that clearing the interrupt status also
clears the native status. If the Java interrupt status is false, it is
important to verify that this is consistent with the native status to
prevent an unexpected interrupt exception the next time sleep() or
wait() is called.

Related to https://github.com/eclipse-openj9/openj9/issues/21804

### DIFF
--- a/src/java.base/share/classes/java/lang/Thread.java
+++ b/src/java.base/share/classes/java/lang/Thread.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2021, 2024 All Rights Reserved
+ * (c) Copyright IBM Corp. 2021, 2025 All Rights Reserved
  * ===========================================================================
  */
 
@@ -1646,7 +1646,7 @@ public class Thread implements Runnable {
 
     final void clearInterrupt() {
         // assert Thread.currentCarrierThread() == this;
-        if (interrupted) {
+        if (interrupted || isInterruptedImpl()) {
             interrupted = false;
             clearInterruptEvent();
         }
@@ -1657,17 +1657,23 @@ public class Thread implements Runnable {
         if (com.ibm.oti.vm.VM.isJVMInSingleThreadedMode()) {
             return interruptedImpl();
         }
-        boolean oldValue = interrupted;
+
+        /*
+         * It is possible for the Java and native interrupt status to get out of sync. Sending
+         * two interrupts to the same thread in quick succession can cause this. It is critical
+         * that clearing the interrupt status also clears the native status. If the Java
+         * interrupt status is false, it is important to verify that this is consistent with the
+         * native status to prevent an unexpected interrupt exception the next time
+         * sleep() or wait() is called.
+         */
+        boolean oldValue = interrupted || isInterruptedImpl();
         if (oldValue) {
             synchronized (interruptLock) {
-                oldValue = interrupted;
                 // We may have been interrupted the moment after we read the field,
                 // so only clear the field if we saw that it was set and will return
                 // true; otherwise we could lose an interrupt.
-                if (oldValue) {
-                    interrupted = false;
-                    clearInterruptEvent();
-                }
+                interrupted = false;
+                clearInterruptEvent();
             }
         }
         return oldValue;


src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1657,10 +1657,19 @@ boolean getAndClearInterrupt() {
...
         }
-        boolean oldValue = interrupted;
+
+        /**
+         * Its possible for the java and native interrupt status to get out of sync. Sending
+         * two interrupts to the same thread in quick succession can cause this. Its critical
+         * that when clearing the interrupt status we also clear the native status. If the
+         * java interrupt status is false, we need to double check that this agrees with the
+         * native status so we dont get an interrupt exception the next time sleep() or wait()
+         * is called.
```
babsingh, line 1667:
```suggestion
         * It is possible for the Java and native interrupt status to get out of sync. Sending
         * two interrupts to the same thread in quick succession can cause this. It is critical
         * that clearing the interrupt status also clears the native status. If the Java
         * interrupt status is false, it is important to verify that this is consistent with the
         * native status to prevent an unexpected interrupt exception the next time
         * sleep() or wait() is called.
```

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1657,10 +1657,19 @@ boolean getAndClearInterrupt() {
...
+
+        /**
+         * It is possible for the Java and native interrupt status to get out of sync. Sending
+         * two interrupts to the same thread in quick succession can cause this. It is critical
+         * that clearing the interrupt status also clears the native status. If the Java
+         * interrupt status is false, it is important to verify that this is consistent with the
+         * native status to prevent an unexpected interrupt exception the next time
+         * sleep() or wait() is called.
+         */
+        boolean oldValue = interrupted ? true : isInterruptedImpl();
```
keithc-ca, line 1669: Why not use `||`?
```java
        boolean oldValue = interrupted || isInterruptedImpl();
```

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1657,10 +1657,19 @@ boolean getAndClearInterrupt() {
...
         if (oldValue) {
             synchronized (interruptLock) {
-                oldValue = interrupted;
+                oldValue = interrupted || oldValue;
```
keithc-ca, line 1672: Why not `|=`?
```java
                oldValue |= interrupted;
```

keithc-ca:
Did you mean to reference https://github.com/eclipse-openj9/openj9/issues/21804? It's already closed.

src/java.base/share/classes/java/lang/Thread.java
keithc-ca: Please explain how this helps. `oldValue` (a local) is already `true`, otherwise we would not be here. Reading the field again and or-ing it won't change the value of that local, so there's no need to test it again a few lines below.

tajila:
> Did you mean to reference https://github.com/eclipse-openj9/openj9/issues/21804? It's already closed.

For some reason, it closed as soon as I opened the PR. This PR addresses that issue.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1657,10 +1657,19 @@ boolean getAndClearInterrupt() {
         if (com.ibm.oti.vm.VM.isJVMInSingleThreadedMode()) {
             return interruptedImpl();
         }
-        boolean oldValue = interrupted;
+
+        /**
```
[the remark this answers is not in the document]
tajila: changed

src/java.base/share/classes/java/lang/Thread.java
tajila: changed

keithc-ca:
> For some reason, it closed as soon as I opened the PR.

I suspect that reason is that commit messages say "fixed 21804". That clause should appear in the _final_ pull request description, that, after merging, resolves the issue.

src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1657,17 +1657,23 @@ boolean getAndClearInterrupt() {
...
                 // so only clear the field if we saw that it was set and will return
                 // true; otherwise we could lose an interrupt.
-                if (oldValue) {
-                    interrupted = false;
-                    clearInterruptEvent();
-                }
+                interrupted = false;
+                clearInterruptEvent();
             }
         }
```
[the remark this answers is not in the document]
keithc-ca: > wait in OMR is interrupted, clears omrthread thread interrupt status

Please explain why the interrupt status **_should_** be cleared at that point.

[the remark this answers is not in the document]
tajila: The other way of looking at it is, once `omrthread_monitor_wait_interruptable` has reported the interrupt, why should the thread omrthread still have the interrupt status.

tajila:
Anymore comments?

keithc-ca:
Ok, I've convinced myself that this should fix the issue reported in https://github.com/eclipse-openj9/openj9/issues/21804. Given the limited uses of `getAndClearInterrupt()` I can find, I don't see how the multiple thread scenario I was concerned about could arise.

I do wonder, though, whether a similar change should be made in `VirtualThread`?

keithc-ca (approved):
Approved based on the testing completed in https://github.com/ibmruntimes/openj9-openjdk-jdk24/pull/96.

PR #1024, opened 2025-05-27
Remove unnecessary export of OPENJ9_BUILD=true

keithc-ca:
See https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/832#pullrequestreview-2871736796.


keithc-ca:
The only files that I can find that would be affected by having `OPENJ9_BUILD` in the environment are
* `runtime/compiler/build/toolcfg/common.mk`
* `runtime/makelib/mkconstants.mk.ftl`
* `runtime/makelib/targets.mk.ftl`

and they're only used in UMA builds.

PR #1025, opened 2025-05-27
Improve building OpenSSL on Windows

keithc-ca:
Add configuration option `no-makedepend` which, according to [NOTES-WINDOWS.md](https://github.com/openssl/openssl/blob/master/NOTES-WINDOWS.md), "can speed up build times by up to 50%". Specifying that option appears to have no effect on other platforms.

### DIFF
--- a/closed/openssl.gmk
+++ b/closed/openssl.gmk
@@ -108,7 +108,7 @@ endif # CCACHE
 
 build_openssl :
 	@$(ECHO) Compiling OpenSSL in $(OPENSSL_DIR) for $(OPENSSL_TARGET)$(if $(OPENSSL_CONFIG_CFLAGS), with additional CFLAGS $(OPENSSL_CONFIG_CFLAGS))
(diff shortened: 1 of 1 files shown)


pshipton (approved):

jasonkatonica:
I have also noticed that we spend quite a bit of time compiling the tests. We could avoid this to save time too. Iv seen discussion at https://github.com/openssl/openssl/issues/1511 talking about options to avoid compiling test binaries ( and I would assume each of the test files ).

keithc-ca:
With everything else already built (so there's no contention with other parts of the build), the current state yields an OpenSSL build time of about 48 seconds (on my Linux laptop); adding `no-tests` to the configuration command drops that build time to about 33 seconds. Fifteen seconds isn't huge in the context of a 20-60+ minute build (depending on the platform), but I don't think we benefit from building those tests so we may as well avoid that work. I'll create another batch of pull requests to make that change.

PR #1028, opened 2025-05-28
Increase Deadlock test timeout for FIPS

JasonFengJ9:
Increase `Deadlock` test timeout for FIPS

In weak fips mode `-Dsemeru.fips=true` `-Dsemeru.customprofile=OpenJCEPlusFIPS` there is a bit more processing to
be done during startup to load and read the FIPS profiles from the `java.security` file and setup the JCE algorithms and providers available according to the profile.

Related to 
* https://github.com/eclipse-openj9/openj9/issues/21919

### DIFF
--- a/test/jdk/java/security/Security/ClassLoaderDeadlock/Deadlock.java
+++ b/test/jdk/java/security/Security/ClassLoaderDeadlock/Deadlock.java
@@ -22,11 +22,17 @@
  */
 
 /*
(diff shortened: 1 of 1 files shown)


pshipton (approved):

JasonFengJ9:
There was no timeout failure at JDK 8/11, only backporting to JDK 17/21/24.

PR #1005, opened 2025-05-05
Add sleep in StopThreadTest to ensure virtual thread unmounts

babsingh:
The test blocks a virtual thread on monitorenter, expecting it to be
preempted and unmounted before JVMTI functions are invoked. These JVMTI
calls require the virtual thread to be fully unmounted. Without a delay,
the thread may still be in transition, causing intermittent failures.
Adding a 2-second sleep ensures the vthread has time to preempt and
unmount, improving test reliability.

Related: https://github.com/eclipse-openj9/openj9/issues/21695

PR #1029, opened 2025-05-30
Fix race between notifyAll and GetCurrentContendedMonitor

babsingh:
contmon01 intermittently fails due to a race condition between
notifyAll and JVMTI GetCurrentContendedMonitor, caused by the
asynchronous scheduling of virtual threads. After being notified, a
virtual thread may not immediately contend for the monitor, resulting
in unexpected null values from GetCurrentContendedMonitor.

To mitigate this, a controlled delay is added after the virtual
thread reaches the BLOCKED state. This provides enough time for the
thread to become contended on the monitor before inspection, making
the test behavior more deterministic.

Related: https://github.com/eclipse-openj9/openj9/issues/21647

### DIFF
--- a/test/hotspot/jtreg/serviceability/jvmti/thread/GetCurrentContendedMonitor/contmon01/contmon01.java
+++ b/test/hotspot/jtreg/serviceability/jvmti/thread/GetCurrentContendedMonitor/contmon01/contmon01.java
@@ -20,7 +20,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
-
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2025, 2025 All Rights Reserved
+ * ===========================================================================
+ */
 
 /*
  * @test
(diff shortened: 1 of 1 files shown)


babsingh:
@tajila Could you review this PR?

JDK24 port: https://github.com/ibmruntimes/openj9-openjdk-jdk24/pull/100

test/hotspot/jtreg/serviceability/jvmti/thread/GetCurrentContendedMonitor/contmon01/contmon01.java
```diff
@@ -119,13 +123,19 @@ public static void test(boolean isVirtual) {
...
             System.out.println("\nMain thread entered lockFld's monitor"
                     + "\n\tand calling lockFld.notifyAll() to awake auxiliary thread");
             lockFld.notifyAll();
+
             System.out.println("\nCheck #4: verifying a contended monitor of auxiliary thread ...");
+            while (thread.getState() != Thread.State.BLOCKED) {
+                doSleep(10);
+            }
+            // Wait for 2 seconds to allow the virtual thread to be contended.
+            doSleep(2000);
```
keithc-ca, line 138: This looks like the same sort of "solution" I have been objecting to in #1005. I don't think this is any more correct.

PR #1030, opened 2025-06-03
Return Thread.State.RUNNABLE for VirtualThread.BLOCKING

babsingh:
VirtualThread.BLOCKING currently maps to Thread.State.BLOCKED. This
confuses tests like StopThreadTest and contmon01, which rely on
Thread.getState() to infer if a virtual thread is fully unmounted.

However, BLOCKING is an intermediate state. Mapping it to BLOCKED can
falsely indicate that a virtual thread is unmounted, leading to
incorrect assumptions and intermittent test failures.

To prevent these false positives, BLOCKING will now return
Thread.State.RUNNABLE. This aligns with other intermediate states such
as PARKING, TIMED_PARKING, WAITING, TIMED_WAITING, and YIELDING, which
also return RUNNABLE. Treating BLOCKING similarly ensures consistency
across intermediate states.

Only VirtualThread.BLOCKED will return Thread.State.BLOCKED, which
accurately reflects a fully unmounted virtual thread.

Related: https://github.com/eclipse-openj9/openj9/issues/21695
Related: https://github.com/eclipse-openj9/openj9/issues/21647

This fix also reverts the below PR:
- https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1029

### DIFF
--- a/src/java.base/share/classes/java/lang/VirtualThread.java
+++ b/src/java.base/share/classes/java/lang/VirtualThread.java
@@ -1134,6 +1134,7 @@ final class VirtualThread extends BaseVirtualThread {
                 }
                 // runnable, mounted
                 return Thread.State.RUNNABLE;
+            case BLOCKING:
             case PARKING:
             case TIMED_PARKING:
             case WAITING:
@@ -1149,7 +1150,6 @@ final class VirtualThread extends BaseVirtualThread {
             case TIMED_PINNED:
             case TIMED_WAIT:
                 return Thread.State.TIMED_WAITING;
-            case BLOCKING:
             case BLOCKED:
                 return Thread.State.BLOCKED;
             case TERMINATED:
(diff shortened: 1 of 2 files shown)


babsingh:
JDK24 backport for this fix: https://github.com/ibmruntimes/openj9-openjdk-jdk24/pull/101

If this PR gets merged, then
- https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1005 can be closed since this PR resolves the same issue.
- @backwaterred We would also like to port this fix to OpenJDK. Would you be able to help propose it upstream once this PR gets merged?

@keithc-ca Here are the changes we discussed offline. Requesting your review. I will initiate the necessary testing to evaluate any side effects in https://github.com/ibmruntimes/openj9-openjdk-jdk24/pull/101.

src/java.base/share/classes/java/lang/VirtualThread.java
```diff
@@ -1139,6 +1139,7 @@ Thread.State threadState() {
             case WAITING:
             case TIMED_WAITING:
             case YIELDING:
+            case BLOCKING:
```
keithc-ca, line 1142: These cases are somewhat sorted (by pair X and TIMED_X); please keep them that way by putting `BLOCKING` before `PARKING`.

keithc-ca (changes requested):

keithc-ca:

keithc-ca (approved):
None of the test failures appear related to this change.

PR #1036, opened 2025-06-10
Allow additional characters in RestrictedSecurity profile constraints

KostasTsiounis:
Some more characters, like `#` and `:`, need to be allowed in RestrictedSecurity profile constraints, specifically the algorithm portion.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -1550,7 +1550,7 @@ public final class RestrictedSecurity {
 
             // Check whether constraints are properly specified.
             final String typeRE = "\\w+";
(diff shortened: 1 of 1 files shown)


keithc-ca (changes requested):
Please share examples where the new characters are used.

keithc-ca:
That shows `#` used in a constraint. Please give examples showing why `:` and `#` should be accepted in algorithm names.

KostasTsiounis:
> That shows `#` used in a constraint. Please give examples showing why `:` and `#` should be accepted in algorithm names.

Not sure what you mean. The algorithm part of the constraint (i.e., `http://www.w3.org/2001/10/xml-exc-c14n#`) has both.

keithc-ca:
Jenkins compile plinux jdknext

keithc-ca (approved):

PR #1038, opened 2025-06-13
Generate keypair using tested provider

KostasTsiounis:
For the keypair generator tests, the keypairs are generated by the first available provider and not the provider that is to be tested.

With this change the service from the tested provider is explicitly requested where needed.

When a keypair from any provider is good enough, we allow the test to get it from the first available.


keithc-ca:
I think this looks good, but was unable to test it in a grinder because things are not ready for Java 26 which requires jtreg version 7.5.2+1. Could you prepare the change for jdk25 which can be tested while we work to update the test infrastructure?

PR #1043, opened 2025-06-29
Use default charset for PrintWriter on z/OS to avoid hardcoded UTF-8

yathamravali:
Modified JFR tool `print` suboption to use the platform's default charset when printing to System.out on z/OS systems, for Java level below 18. This avoids forcing UTF-8 output on EBCDIC-based consoles, ensuring proper encoding and display.

For Java21 on zos I don't see any issue - https://www.ibm.com/docs/en/semeru-runtime-ce-z/21.0.0?topic=guide-file-encoding-utf-8-as-default-charset

### DIFF
--- a/src/jdk.jfr/share/classes/jdk/jfr/internal/tool/Print.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/internal/tool/Print.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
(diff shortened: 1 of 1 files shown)


yathamravali:
@pshipton - Sorry about that. I wan in the impression that this should be opened here first before backporting. 

Can I just open in JDK11 and JD17 zos specific repo's or should it go to https://github.ibm.com/runtimes/openj9-openjdk-jdk-zos first then backport to JDK 11,17?

PR #710, opened 2023-12-07
Integrate OpenJCEPlus into Semeru OpenJDK

taoliult:
Depends on https://github.com/eclipse-openj9/openj9/pull/18544 being merged first.

Integrate the building of OpenJCEPlus with Semeru OpenJDK as a module "openjceplus". The java codes will be built with Semeru OpenJDK as a jmod, and the native codes will be built as a ".so" or ".dll" library.

The "bash get_source.sh" will download the OpenJCEPlus repo and also the ICC binaries which needed when building the OpenJCEPlus native codes.

Example:

bash get_source.sh -openjceplus-repo=[OpenJCEPlus repo] -openjceplus-branch=[OpenJCEPlus branch] -gskit-bin=[ICC binary tar file] -gskit-sdk-bin=[ICC SDK binary tar file] -gskit-credential=[Credential for downloading ICC]

bash configure --with-boot-jdk=[Boot JDK] --enable-openjceplus

Signed-off-by: Tao Liu [tao.liu@ibm.com](mailto:tao.liu@ibm.com)

PR #1046, opened 2025-07-03
Add filter when copying OCK libraries

taoliult:
This commit adds a filter to the OCK library copying process, allowing only library and .txt files to be copied.

### DIFF
--- a/closed/make/modules/openjceplus/Copy.gmk
+++ b/closed/make/modules/openjceplus/Copy.gmk
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2023, 2024 All Rights Reserved
+# (c) Copyright IBM Corp. 2023, 2025 All Rights Reserved
 # ===========================================================================
 # This code is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 only, as
@@ -33,7 +33,7 @@ ifeq (true,$(BUILD_OPENJCEPLUS))
   # Copy OpenJCEPlus native libraries.
   $(eval $(call SetupCopyFiles, OPENJCEPLUS_JGSKIT_LIBS_COPY, \
       SRC := $(OPENJCEPLUS_TOPDIR)/target, \
-      FILES := $(filter %.dll %.dylib %.so %.x, $(call FindFiles, $(OPENJCEPLUS_TOPDIR)/target)), \
+      FILES := $(filter %$(SHARED_LIBRARY_SUFFIX), $(call FindFiles, $(OPENJCEPLUS_TOPDIR)/target)), \
       FLATTEN := true, \
       DEST := $(LIB_DST_DIR), \
   ))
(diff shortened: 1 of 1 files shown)


pshipton:
The copyright date needs to be updated.

taoliult:
> The copyright date needs to be updated.

Updated.

closed/make/modules/openjceplus/Copy.gmk
```diff
@@ -51,7 +51,7 @@ ifeq (true,$(BUILD_OPENJCEPLUS))
   $(eval $(call SetupCopyFiles, OPENJCEPLUS_OCK_COPY, \
       SRC := $(OPENJCEPLUS_OCK_DIR), \
       DEST := $(SUPPORT_OUTPUTDIR)/$(OPENJCEPLUS_OCK_SUB_DIR)/$(MODULE), \
-      FILES := $(call FindFiles, $(OPENJCEPLUS_OCK_DIR)), \
+      FILES := $(filter %.dll %.dylib %.so %.x %.txt, $(call FindFiles, $(OPENJCEPLUS_OCK_DIR))), \
```
keithc-ca, line 54: Suggest we should have a variable for this pattern:
```
      FILES := $(filter $(OPENJCEPLUS_FILTER_PATTERN) %.txt, $(call FindFiles, $(OPENJCEPLUS_OCK_DIR))), \
```
where `OPENJCEPLUS_FILTER_PATTERN` is defined above
```
  OPENJCEPLUS_FILTER_PATTERN := %$(SHARED_LIBRARY_SUFFIX) %.txt
  ifeq ($(call isTargetOs, zos), true)
    OPENJCEPLUS_FILTER_PATTERN += %.x
  endif
```
and also used on line 36.

Can you give some examples of files that would be omitted by this new filter?

keithc-ca: Perhaps just two uses don't fully support the introduction of a variable, but I do think it was a mistake that the pattern included elements that should not match (e.g. `%.dll` should only apply on Windows, `%.dylib` only on macOS).

I was hoping you could be more specific than "an unexpected file".

pshipton: We want to remove the `Standard_International_Program_License_Agreement.pdf` file that started showing up in the lib directory.

keithc-ca: I understand, but the filter should not match files that are not appropriate for the target platform.

If `.x` files are not needed (even on z/OS), I'm fine if you don't want to capture the pattern list in a variable
```
      FILES := $(filter %$(SHARED_LIBRARY_SUFFIX) %.txt, $(call FindFiles, $(OPENJCEPLUS_OCK_DIR))), \
```
but I will insist that the filter not admit any files that would only be relevant on a different platform.

I believe it was a mistake not to use `SHARED_LIBRARY_SUFFIX` originally in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/710; let's fix that now.

taoliult: Replaced with SHARED_LIBRARY_SUFFIX and removed .x, as Z/OS does not use the open-source OpenJCEPlus and has a different build process.

keithc-ca (approved):

PR #1045, opened 2025-07-02
Update HKDF constraints in RestrictedSecurity profile

KostasTsiounis:
In the `RestrictedSecurity` profile that includes the `OpenJCEPlusFIPS` provider, the constraints that pertain to the `HKDF` services are updated to the new implementation.

Signed-off-by: Kostas Tsiounis <kostas.tsiounis@ibm.com>

PR #1047, opened 2025-07-04
Enable HKDF in OpenJCEPlus restrict Profile

JinhangZhang:
Starting in Java25, key derivation functions can be registered as services extending the KDFSpi, and OpenJCEPlus refactors KDF from KeyGeneratorSpi to KDFSpi. So we need to explicitly register KDF in OpenJCEPlusFIPS restrict profile.


JinhangZhang:
@jasonkatonica FYI

JinhangZhang:
Right, it's a duplicate of https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1045

PR #1048, opened 2025-07-08
Remove the keystore property in Restricted profile

JinhangZhang:
We dont need to restrict the keystore in OpenJCEPlusFIPS in FIPS140-3.


JinhangZhang:
@jasonkatonica FYI

jasonkatonica:
FYI - This update is NOT intended to be put into the .53 release branches.

jasonkatonica:
The only properties spanned by the hash value for a profile are the properties in the profile itself. For example in this case all the properties that start with `RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3` should be included. I don't expect the hash to include all the regular properties such as `javax.net.ssl.keyStore` since this could cause problems as these values are changed from OpenJDK.

keithc-ca:
> all the properties that start with `RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3` should be included

I was referring to:
```
RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.javax.net.ssl.keyStore = NONE
```
It's removal doesn't appear to affect the expected hash value. So my question remains: is that intentional or an oversight?

JinhangZhang:
> OK, I expect this code in this PR to fail if the line you mentioned is removed.
> 
> ```
> RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.javax.net.ssl.keyStore = NONE
> ```
> 
> @JinhangZhang Is this PR working as you opened it without any update to the hash value? If so we will need to investigate why the removal of this property did not cause failure as expected....

There should be a new hash value there. My first commit has a new hash value, but since Kostas also have a PR for adding KDF in restricted profile and that PR is merged in ahead of mine. I have to re-calculate the hash.

keithc-ca:
> It's removal doesn't appear to affect the expected hash value.

Perhaps it was the way I tested the updated `java.security` file; I may have not used the double `=` in `-Djava.security.properties==java.security`. With `==`, I see that the hash value is sensitive to that property.

PR #1054, opened 2025-07-21
Reduce LotsOfContendedMonitorEnter test size

fengxue-IS:
There is a known performance issue with OpenJ9 causing this test to
timeout, this change is to temporarily reduce the size of the test to
retain test coverage while the perf issue is being addressed.

Tracking issue: eclipse-openj9/openj9#22153

JDK24 extension repo: https://github.com/ibmruntimes/openj9-openjdk-jdk24/pull/114
JDK25 entension repo: https://github.com/ibmruntimes/openj9-openjdk-jdk25/pull/14


keithc-ca:
Please update the commit message and the description (here and in back-ports) to explain that:
* there is a known performance issue
* this is instead of disabling the test for at least some test coverage
* this is intended to be temporary

See https://github.com/ibmruntimes/openj9-openjdk-jdk24/pull/114#issuecomment-3098014030.

PR #1057, opened 2025-07-26
Fix bootcycle-images target

fridrich:
The Cmake needs the toolchain.cmake file that is normally in the output directory of the first pass as generated during the configure phase. So, for the bootcycle-images target, the CMake has to look for it in OLD_OUTPUTDIR and then the bootcycle build works just fine.


pshipton:
We can't accept non-IBM contributions to this project.

PR #1058, opened 2025-07-30
Allow FIPS mode on all Semeru platforms

jasonkatonica:
OpenJCEPlus can now be enabled in developer mode for use on all Semeru
platforms. This update allows the restricted security features
associated with FIPS 140-3 to be enabled for use including FIPS security
profiles associated with the OpenJCEPlusFIPS provider when running in
FIPS mode.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -64,7 +64,7 @@ public final class RestrictedSecurity {
     private static final boolean allowSetProperties;
 
     private static final boolean isNSSSupported;
-    private static final boolean isOpenJCEPlusSupported;
+    private static final boolean isOpenJCEPlusCertifiedPlatform;
 
     private static final boolean userSetProfile;
     private static final boolean shouldEnableSecurity;
@@ -83,49 +83,20 @@ public final class RestrictedSecurity {
 
     private static final Set<String> unmodifiableProperties = new HashSet<>();
 
-    private static final Map<String, List<String>> supportedPlatformsNSS = new HashMap<>();
-    private static final Map<String, List<String>> supportedPlatformsOpenJCEPlus = new HashMap<>();
-
     static {
-        supportedPlatformsNSS.put("Arch", List.of("amd64", "ppc64le", "s390x"));
-        supportedPlatformsNSS.put("OS", List.of("Linux"));
-
-        supportedPlatformsOpenJCEPlus.put("Arch", List.of("amd64", "ppc64", "s390x"));
-        supportedPlatformsOpenJCEPlus.put("OS", List.of("Linux", "AIX", "Windows"));
-
+        final List<String> nssSupportedArch = List.of("amd64", "ppc64le", "s390x");
+        final List<String> openjceplusCertifiedArch = List.of("amd64", "ppc64", "s390x");
+        final List<String> openjceplusCertifiedOS = List.of("AIX", "Linux", "Windows");
         String osName = System.getProperty("os.name");
         String osArch = System.getProperty("os.arch");
 
-        boolean isOsSupported, isArchSupported;
         // Check whether the NSS FIPS solution is supported.
-        isOsSupported = false;
-        for (String os: supportedPlatformsNSS.get("OS")) {
-            if (osName.contains(os)) {
-                isOsSupported = true;
-            }
-        }
-        isArchSupported = false;
-        for (String arch: supportedPlatformsNSS.get("Arch")) {
-            if (osArch.contains(arch)) {
-                isArchSupported = true;
-            }
-        }
-        isNSSSupported = isOsSupported && isArchSupported;
+        isNSSSupported = osName.contains("Linux")
+                && containsAny(osArch, nssSupportedArch);
 
-        // Check whether the OpenJCEPlus FIPS solution is supported.
-        isOsSupported = false;
-        for (String os: supportedPlatformsOpenJCEPlus.get("OS")) {
-            if (osName.contains(os)) {
-                isOsSupported = true;
-            }
-        }
-        isArchSupported = false;
-        for (String arch: supportedPlatformsOpenJCEPlus.get("Arch")) {
-            if (osArch.contains(arch)) {
-                isArchSupported = true;
-            }
-        }
-        isOpenJCEPlusSupported = isOsSupported && isArchSupported;
+        // Check whether the OpenJCEPlus FIPS solution is certified.
+        isOpenJCEPlusCertifiedPlatform = containsAny(osName, openjceplusCertifiedOS)
+                && containsAny(osArch, openjceplusCertifiedArch);
 
         // Check the default solution to see if FIPS is supported.
         isFIPSSupported = isNSSSupported;
@@ -412,12 +383,14 @@ public final class RestrictedSecurity {
                     + " on this platform.");
         }
 
-        if (profileID.contains("OpenJCEPlus") && !isOpenJCEPlusSupported) {
-            printStackTraceAndExit("OpenJCEPlus RestrictedSecurity profiles are not supported"
-                    + " on this platform.");
-        }
-
         if (debug != null) {
+            if (profileID.contains("OpenJCEPlus")) {
+                if (isOpenJCEPlusCertifiedPlatform) {
+                    debug.println("OpenJCEPlus RestrictedSecurity profile running on a certified platform.");
+                } else {
+                    debug.println("OpenJCEPlus RestrictedSecurity profile running in developer mode.");
+                }
+            }
             debug.println("RestrictedSecurity profile " + profileID
                     + " is supported on this platform.");
         }
@@ -707,6 +680,15 @@ public final class RestrictedSecurity {
      */
     private static boolean isAsterisk(String string) {
         return "*".equals(string);
+    }
+
+    private static boolean containsAny(String string, List<String> substrings) {
+        for (String substring : substrings) {
+            if (string.contains(substring)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     /**
--- a/src/java.base/share/conf/security/java.security
+++ b/src/java.base/share/conf/security/java.security
@@ -184,7 +184,6 @@ RestrictedSecurity.NSS.140-2.securerandom.provider = SunPKCS11-NSS-FIPS
 RestrictedSecurity.NSS.140-2.securerandom.algorithm = PKCS11
 #endif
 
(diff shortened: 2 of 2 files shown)


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -84,14 +84,14 @@ public final class RestrictedSecurity {
     private static final Set<String> unmodifiableProperties = new HashSet<>();
 
     private static final Map<String, List<String>> supportedPlatformsNSS = new HashMap<>();
-    private static final Map<String, List<String>> supportedPlatformsOpenJCEPlus = new HashMap<>();
+    private static final Map<String, List<String>> certifiedPlatformsOpenJCEPlus = new HashMap<>();
```
keithc-ca, line 87: These were never used outside the static initializer; I think it's time to remove the fields and just use the relevant lists in the static initializer. I suggest we add this helper method (or use patterns):
```java
    private static boolean containsAny(String string, List<String> substrings) {
        for (String substring : substrings) {
            if (string.contains(substring)) {
                return true;
            }
        }
        return false;
    }
```
Then lines 90-128 are simplified to:
```java
        final List<String> nssSupportedArch = List.of("amd64", "ppc64le", "s390x");
        final List<String> certifiedArch = List.of("amd64", "ppc64le", "s390x");
        final List<String> certifiedOS = List.of("AIX", "Linux", "Windows");

        String osName = System.getProperty("os.name");
        String osArch = System.getProperty("os.arch");

        // Check whether the NSS FIPS solution is supported.
        isNSSSupported = osName.contains("Linux") && containsAny(osArch, nssSupportedArch);

        // Check whether the OpenJCEPlus FIPS solution is supported.
        isOpenJCEPlusCertifiedPlatform = containsAny(osName, certifiedOS) && containsAny(osArch, certifiedArch);
```
You'll notice that `nssSupportedArch` and `certifiedArch` are the same. If the logical difference is not worth keeping, one of them can be removed.

Instead of `containsAny()` we could instead use patterns; e.g. for testing the architecture: `Pattern.compile("amd64|ppc64le|s390x").matcher(osArch).find()`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -412,9 +412,14 @@ private static void checkIfKnownProfileSupported() {
...
+            if (profileID.contains("OpenJCEPlus")) {
+                if (isOpenJCEPlusCertifiedPlatform) {
+                    debug.println("OpenJCEPlus RestrictedSecurity profile running on certified platform.");
+                } else {
+                    debug.println("OpenJCEPlus RestrictedSecurity profile running in developer mode.");
+                }
+            }
         }
 
         if (debug != null) {
```
keithc-ca, line 425: These two `if` blocks can be merged.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -412,9 +412,14 @@ private static void checkIfKnownProfileSupported() {
                     + " on this platform.");
         }
 
-        if (profileID.contains("OpenJCEPlus") && !isOpenJCEPlusSupported) {
-            printStackTraceAndExit("OpenJCEPlus RestrictedSecurity profiles are not supported"
-                    + " on this platform.");
+        if (debug != null) {
+            if (profileID.contains("OpenJCEPlus")) {
+                if (isOpenJCEPlusCertifiedPlatform) {
+                    debug.println("OpenJCEPlus RestrictedSecurity profile running on certified platform.");
```
keithc-ca, line 418: Perhaps "a" is missing in "on a certified platform"?

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
jasonkatonica: Added "a"

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -83,49 +83,19 @@ public final class RestrictedSecurity {
...
 
     static {
-        supportedPlatformsNSS.put("Arch", List.of("amd64", "ppc64le", "s390x"));
-        supportedPlatformsNSS.put("OS", List.of("Linux"));
-
-        supportedPlatformsOpenJCEPlus.put("Arch", List.of("amd64", "ppc64", "s390x"));
-        supportedPlatformsOpenJCEPlus.put("OS", List.of("Linux", "AIX", "Windows"));
-
+        final List<String> nssSupportedArch = List.of("amd64", "ppc64le", "s390x");
+        final List<String> openjceplusCertifiedArch = List.of("amd64", "ppc64", "s390x");
```
keithc-ca, line 89: I realize the previous version did the same, but is Linux/ppc64le not a certified platform?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -83,49 +83,19 @@ public final class RestrictedSecurity {
 
     private static final Set<String> unmodifiableProperties = new HashSet<>();
 
-    private static final Map<String, List<String>> supportedPlatformsNSS = new HashMap<>();
-    private static final Map<String, List<String>> supportedPlatformsOpenJCEPlus = new HashMap<>();
```
keithc-ca, line 86: Please delete one of the two blank lines here.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -83,49 +83,19 @@ public final class RestrictedSecurity {
...
-            }
-        }
-        isArchSupported = false;
-        for (String arch: supportedPlatformsNSS.get("Arch")) {
-            if (osArch.contains(arch)) {
-                isArchSupported = true;
-            }
-        }
-        isNSSSupported = isOsSupported && isArchSupported;
+        isNSSSupported = osName.contains("Linux") && containsAny(osArch, nssSupportedArch);
```
keithc-ca, line 95: I think this and line 98 deserve to be folded:
```java
        isNSSSupported = osName.contains("Linux")
                && containsAny(osArch, nssSupportedArch);
```
```java
        isOpenJCEPlusCertifiedPlatform = containsAny(osName, openjceplusCertifiedOS)
                && containsAny(osArch, openjceplusCertifiedArch);
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -412,14 +382,16 @@ private static void checkIfKnownProfileSupported() {
...
+            if (profileID.contains("OpenJCEPlus")) {
+                if (isOpenJCEPlusCertifiedPlatform) {
+                    debug.println("OpenJCEPlus RestrictedSecurity profile running on a certified platform.");
+                } else {
+                    debug.println("OpenJCEPlus RestrictedSecurity profile running in developer mode.");
+                }
+            }
             debug.println("RestrictedSecurity profile " + profileID
-                    + " is supported on this platform.");
+                + " is supported on this platform.");
```
keithc-ca, line 394: The indentation here should not change.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -709,6 +681,15 @@ private static boolean isAsterisk(String string) {
         return "*".equals(string);
     }
 
+    private static boolean containsAny(String string, List<String> substrings) {
```
keithc-ca, line 684: No trailing spaces, please.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
jasonkatonica: I left `ppc64` in here since I believe that both AIX ( arch=ppc64 ) and Linux ( arch=ppc64le ) are both supported and certified architectures. The new `containsAny` hopefully returns `true` for both these strings when we state that `ppc64` substring is certified.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
jasonkatonica: Deleted an additional line here.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
jasonkatonica: Folded both of these lines mentioned.

jasonkatonica:
> The final version of this should be squashed; the commit message should spell "FIPS" consistently (i.e. not "fips mode").

Sqashed these commits and updated the commit an PR text to make "FIPS" more consistent.

keithc-ca:
Jenkins test all,sanity,extended aix,amac jdknext

jasonkatonica:
I did finish testing this at a high level by building on `macos aarch64` and `linux x86`. I was able to validate the java.security profiles are present in both builds as I expect and was able to use a command as so to enable fips mode to confirm debug traces along with algorithms reported being registered from OpenJCEPlusFIPS. 
```
java -Djava.security.auth.debug=all -Dsemeru.fips=true -Dsemeru.customprofile=OpenJCEPlusFIPS.FIPS140-3 SecurityProvidersAndAlgorithms.java
```

jasonkatonica:
> Is an equivalent change in the works for jdk8?

The `OpenJCEPlusFIPS` provider ( FIPS 140-3 ) is only supported on Java 11 and higher. This update is not applicable for Semeru Java 8.

keithc-ca:
The code suggests otherwise: the field `isOpenJCEPlusSupported` mirrors newer versions - see https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java#L70.

jasonkatonica:
It seems the code was made similar between 8 and 11+ in this logic however we do not support OpenJCEPlus or OpenJCEPlusFIPS on Semeru 8. We could backport this change also for consistancy sake if it makes sense however we dont plan on ever bundling or supporting OpenJCEPlus and OpenJCEPlusFIPS nor the 140-3 profiles there.

keithc-ca:
If we don't ever intend to support OpenJCEPlus in Semeru 8, we should remove the useless code there.

PR #1062, opened 2025-08-11
Add and fix debug messages in RestrictedSecurity

KostasTsiounis:
The exception message when trying to add or remove constraints to a previously non-existing provider is fixed.

A debug message is added to the `ProviderList` to indicate when a provider not allowed in `RestrictedSecurity` mode is not added to the providers list.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -1892,8 +1892,8 @@ public final class RestrictedSecurity {
 
                 String action = m.group(4);
                 if (!update && !isNullOrBlank(action)) {
-                    printStackTraceAndExit("You cannot add or remove to provider "
-                            + m.group(1) + ". This is the base profile.");
+                    printStackTraceAndExit("Constraints of provider not previously specified"
+                            + " cannot be modified: " + providerName);
                 }
             } else {
                 printStackTraceAndExit("Provider format is incorrect: " + providerInfo);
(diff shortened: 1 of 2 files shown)


closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1892,8 +1892,8 @@ private static void checkProviderFormat(String providerInfo, boolean update) {
 
                 String action = m.group(4);
                 if (!update && !isNullOrBlank(action)) {
-                    printStackTraceAndExit("You cannot add or remove to provider "
-                            + m.group(1) + ". This is the base profile.");
+                    printStackTraceAndExit("You cannot add or remove to a provider "
+                            + "not previously specified: " + m.group(1));
```
keithc-ca, line 1896: Suggest reusing `providerName`, and updating message:
```java
                    printStackTraceAndExit("Provider not previously specified"
                            + " cannot be modified: " + providerName);
```

KostasTsiounis: Changed it to:
```java
printStackTraceAndExit("Constraints of provider not previously specified"
                            + " cannot be modified: " + providerName);
```

keithc-ca: Even better. Thanks.

keithc-ca:
Actually, I expect that build to fail because the openj9 repo has changes necessary to allow the openj9-staging branch of this repo to promote, but that hasn't happened yet, so it will complain about an `@Overrides` annotation. I verified this locally, adjusted for that issue.

keithc-ca (approved):

PR #1069, opened 2025-08-18
Support a list of java.security property files

JinhangZhang:
This PR proposes a new system property named java.security.propertiesList. This new property will allow multiple java.security profiles from different files to be used rather than put those different profiles into a single java.security file. This new property supports a way to configure the superset of all files in this set of property files.

### DIFF
--- a/src/java.base/share/classes/java/security/Security.java
+++ b/src/java.base/share/classes/java/security/Security.java
@@ -31,6 +31,7 @@
 
 package java.security;
 
+import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -126,6 +127,9 @@ public final class Security {
         private static final String EXTRA_SYS_PROP =
                 "java.security.properties";
 
+        private static final String EXTRA_SYS_PROP_LIST =
+                "java.security.propertiesList";
+
         private static Path currentPath;
 
         private static final Set<Path> activePaths = new HashSet<>();
@@ -134,7 +138,15 @@ public final class Security {
             // first load the master properties file to
             // determine the value of OVERRIDE_SEC_PROP
             loadMaster();
-            loadExtra();
+
+            // If java.security.propertiesList is present, it always takes precedence
+            // and java.security.properties will be ignored.
+            String propList = System.getProperty(EXTRA_SYS_PROP_LIST);
+            if ((propList != null) && !propList.isBlank()) {
+                loadExtraFromList(propList);
+            } else {
+                loadExtra();
+            }
         }
 
         static boolean isInclude(String key) {
@@ -176,6 +188,33 @@ public final class Security {
                                     "properties from " + propFile);
                             e.printStackTrace();
                         }
+                    }
+                }
+            }
+        }
+
+        private static void loadExtraFromList(String propList) {
+            if ("true".equalsIgnoreCase(props.getProperty(OVERRIDE_SEC_PROP))) {
+                for (String file : propList.split(File.pathSeparator)) {
+                    // propertiesList does not support OVERRIDE mode
+                    if (file.startsWith("=")) {
+                        throw new IllegalArgumentException(
+                                "java.security.propertiesList does not support '=' prefix: " + file);
+                    }
+
+                    if (sdebug != null) {
+                        sdebug.println("java.security.propertiesList list file: " + file);
+                    }
+
+                    LoadingMode mode = LoadingMode.APPEND;
+                    try {
+                        loadExtraHelper(file, mode);
+                    } catch (Exception e) {
+                        if (sdebug != null) {
+                            sdebug.println("unable to load security properties from list file: " + file);
+                            e.printStackTrace();
+                        }
+                        throw new RuntimeException("Failed to load security properties from list file: " + file, e);
                     }
                 }
             }


src/java.base/share/classes/java/security/Security.java
```diff
@@ -40,6 +40,7 @@
 import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
+import java.util.ArrayList;
```
keithc-ca, line 43: Please remove this unused import.

src/java.base/share/classes/java/security/Security.java
```diff
@@ -134,7 +138,15 @@ static void loadAll() {
             // first load the master properties file to
             // determine the value of OVERRIDE_SEC_PROP
             loadMaster();
-            loadExtra();
+
+            // If java.security.propertieslist is present, it always takes precedence
```
keithc-ca, line 142: Please correct references to the property name in comments and exception messages.

src/java.base/share/classes/java/security/Security.java
```diff
@@ -181,6 +193,37 @@ private static void loadExtra() {
             }
         }
 
+        private static void loadExtraFromList(String propList) {
+            if ("true".equalsIgnoreCase(props.getProperty(OVERRIDE_SEC_PROP))) {
+                for (String file : propList.trim().split(",")) {
```
keithc-ca, line 198: The use of `trim()` here is redundant given the use on line 199.

keithc-ca (changes requested):

src/java.base/share/classes/java/security/Security.java
```diff
@@ -181,6 +193,71 @@ private static void loadExtra() {
...
+                        cur.append(c);
+                    }
+                }
+            }
+            if (inQuote) {
+                throw new IllegalArgumentException("wrong format");
+            }
+            out.add(cur.toString());
+            return out;
+        }
```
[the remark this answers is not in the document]
JinhangZhang: Updated; just use the split(",") function

src/java.base/share/classes/java/security/Security.java
```diff
@@ -134,7 +138,15 @@ static void loadAll() {
             // first load the master properties file to
             // determine the value of OVERRIDE_SEC_PROP
             loadMaster();
-            loadExtra();
+
+            // If java.security.propertieslist is present, it always takes precedence
+            // and java.security.properties will be ignored.
+            String propList = System.getProperty(EXTRA_SYS_PROP_LIST);
+            if (propList != null && !propList.isBlank()) {
```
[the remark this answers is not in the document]
JinhangZhang: Updated

src/java.base/share/classes/java/security/Security.java
```diff
@@ -181,6 +193,71 @@ private static void loadExtra() {
             }
         }
 
+        private static void loadExtraFromList(String propList) {
+            if ("true".equalsIgnoreCase(props.getProperty(OVERRIDE_SEC_PROP))) {
+                String body = propList.trim();
+                List<String> files = splitList(body);
+
+                for (String file : files) {
```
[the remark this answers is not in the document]
JinhangZhang: Updated

src/java.base/share/classes/java/security/Security.java
```diff
@@ -181,6 +193,71 @@ private static void loadExtra() {
...
+                    String item = file.trim();
+                    if (item.isEmpty()) continue;
```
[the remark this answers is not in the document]
JinhangZhang: Updated

src/java.base/share/classes/java/security/Security.java
```diff
@@ -181,6 +193,71 @@ private static void loadExtra() {
...
+
+                    // propertieslist does not support OVERRIDE mode
+                    if (item.startsWith("=")) {
+                        throw new IllegalArgumentException(
+                            "java.security.propertieslist does not support '=' prefix: " + file);
```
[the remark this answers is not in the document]
JinhangZhang: Updated

src/java.base/share/classes/java/security/Security.java
JinhangZhang: updated

src/java.base/share/classes/java/security/Security.java
JinhangZhang: Updated both

JinhangZhang: updated

src/java.base/share/classes/java/security/Security.java
```diff
@@ -181,6 +192,37 @@ private static void loadExtra() {
             }
         }
 
+        private static void loadExtraFromList(String propList) {
+            if ("true".equalsIgnoreCase(props.getProperty(OVERRIDE_SEC_PROP))) {
+                for (String file : propList.split(",")) {
```
jasonkatonica, line 197: Could we use `File.pathSeparator` instead of a `,` here as the character we split by for the files? This might make it more compatible with other path handling in Java on a given OS/arch?

jasonkatonica (changes requested):

src/java.base/share/classes/java/security/Security.java
keithc-ca: I think this is a good suggestion. Please update to use `File.pathSeparator`.

Also, this new property should be documented: Please create an issue at https://github.com/eclipse-openj9/openj9-docs/issues.

keithc-ca: Consistent with using `pathSeparator`, I don't think we should be using `trim()`, and empty entries should be an error rather than being just ignored.

src/java.base/share/classes/java/security/Security.java
```diff
@@ -181,6 +193,36 @@ private static void loadExtra() {
             }
         }
 
+        private static void loadExtraFromList(String propList) {
+            if ("true".equalsIgnoreCase(props.getProperty(OVERRIDE_SEC_PROP))) {
+                for (String file : propList.split(File.pathSeparator)) {
+                    if (file.isEmpty()) {
+                        throw new IllegalArgumentException("unable to load an empty file in java.security.propertiesList");
+                    }
```
keithc-ca, line 201: I'm not sure this is necessary; there's no equivalent explicit check in `loadExtra()`. I suggest you just you omit these three lines.

src/java.base/share/classes/java/security/Security.java
```diff
@@ -181,6 +193,36 @@ private static void loadExtra() {
...
+                    // propertiesList does not support OVERRIDE mode
+                    if (file.startsWith("=")) {
+                        throw new IllegalArgumentException(
+                                "java.security.propertiesList does not support '=' prefix: " + file);
+                    }
+
+                    if (sdebug != null) {
+                        sdebug.println("java.security.propertiesList file value: " + file);
```
keithc-ca, line 209: Suggest this message would be more consistent with "list file" instead of "file value"; see lines 217, 220.

keithc-ca (changes requested):

keithc-ca:
Please fix the commit message; there should be a space in "propertiesList. This".

src/java.base/share/classes/java/security/Security.java
JinhangZhang: updated

src/java.base/share/classes/java/security/Security.java
JinhangZhang: omit that check.

JinhangZhang:
> Please fix the commit message; there should be a space in "propertiesList. This".

updated

JinhangZhang:
New issue is created: https://github.com/eclipse-openj9/openj9-docs/issues/1573

keithc-ca:

keithc-ca:
The build failed because it included https://github.com/eclipse-openj9/openj9/pull/22465 prior to promotion.

PR #1077, opened 2025-08-22
Add thread safety for generateKeyPair in NativeXDHKeyPairGenerator

dev-koan:
NativeXDHKeyPairGenerator is updated to use the correct instance of XDHKeyPairGenerator for varying
curves in multithreaded environment for `generateKeyPair()` when all the threads share the same instance of KeyPairGenerator. 

Signed-off-by: Dev Agarwal <dev.agarwal@ibm.com>

### DIFF
--- a/closed/src/java.base/share/classes/sun/security/ec/NativeXDHKeyPairGenerator.java
+++ b/closed/src/java.base/share/classes/sun/security/ec/NativeXDHKeyPairGenerator.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * (c) Copyright IBM Corp. 2023, 2025 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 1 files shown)


closed/src/java.base/share/classes/sun/security/ec/NativeXDHKeyPairGenerator.java
```diff
@@ -111,6 +111,7 @@ private void initializeImpl(XECParameters params, SecureRandom random) {
         ops = new XECOperations(params);
         this.random = (random != null) ? random : JCAUtil.getSecureRandom();
 
+        initializeJavaImplementation();
```
keithc-ca, line 114: We get here via the constructor, but before `lockedParams` is set leading to incorrect behavior in `initializeJavaImplementation()`.

keithc-ca:
Please elaborate on what goes wrong if we don't make this change.

dev-koan:
> Please elaborate on what goes wrong if we don't make this change.

When multiple threads share the same instance of `NativeXDHKeyPairGenerator` and they call `generateKeyPair()` (and `useJavaImpl` is true or OpenSSL fails for some reason) each of those threads will call `javaImplGenerateKeyPair()`. 

For example two threads will reach `initializeJavaImplementation()` they both enter the if block (javaImplementation == null) one of them returns from the function but hasn't  executed `return javaImplementation.generateKeyPair()` (in `javaImplGenerateKeyPair()`) and the other thread now changes the `javaImplementation` but before it can execute `javaImplementation.initialize(ops.getParameters().getBits(), random);` context switches back to the first thread.
This is the statement (`javaImplementation.initialize(ops.getParameters().getBits(), random);`) which makes sure that for different curves correct key size is used which never executed for the second thread for the changed `javaImplementation`.

keithc-ca:
Thank you for the explanation.

Given that `javaImplementation` will always be initialized in the constructor, I think some refactoring is in order. That field should be `final`, `initializeJavaImplementation()` need not check the field and, should return a new object to be stored by the constructors.

keithc-ca:

PR #1081, opened 2025-08-29
Support SHA-512/224 and SHA-512/256 in strict profile

JinhangZhang:


jasonkatonica:
Hi Jinhang, lets double check that we can support this algorithm with Atsec and John Peck to ensure that it is compliant in or FIPS module before delivering this.

PR #1082, opened 2025-08-29
Fix minimum allocation size

AditiS11:
Fix for Issue eclipse-openj9/openj9#22219
Ensure minimum allocation size of Long.BYTES.
initNativeMemory() writes a full long (Long.BYTES). 
1-byte allocation is not enough and can corrupt the footer added by OMR.


AditiS11:
@pshipton @tajila @theresa-m 
FYI

pshipton:
Tested via https://github.com/ibmruntimes/openj9-openjdk-jdk25/pull/32

PR #1086, opened 2025-09-04
Fix minimum allocation size

AditiS11:
Fix for Issue eclipse-openj9/openj9#22219
If the init flag is set to true
ensure min allocation size of Long.BYTES.
initNativeMemory() writes a full long (Long.BYTES). 
1-byte allocation is not enough and can corrupt the footer.


src/java.base/share/classes/jdk/internal/foreign/SegmentFactories.java
```diff
@@ -203,8 +203,9 @@ private static long allocateNativeInternal(long byteSize, long byteAlignment, Me
             throw new OutOfMemoryError();
         }
         // Always allocate at least some memory so that zero-length segments have distinct
-        // non-zero addresses.
-        alignedSize = Math.max(Long.BYTES, alignedSize);
+        // non-zero addresses. If we are initializing the allocated memory, then use a minimum
+        // size of 8 because we init the memory with longs.
+        alignedSize = Math.max((init ? Long.BYTES : 1), alignedSize);
```
keithc-ca, line 208: Is there a test case where `byteSize` is zero? In that case `alignedSize` would also be zero and the old code would update it to 8. How does reducing `alignedSize` to 1 help (when `init` is false)?

keithc-ca: We should allow that change to flow downstream naturally. I see no urgency to make that change now.

keithc-ca: I disagree: I don't think it's better and I don't expect a conflict when the upstream change arrives here. I also disagree with the upstream suggestion.

PR #1097, opened 2025-09-17
Fix two deadlocks in RestrictedSecurity mode DeadLock test

taoliult:
This PR fixes two issues when running DeadLock test in RestrictedSecurity mode.

One issue arises from the RestrictedSecurity.isPolicySunset() method.

In the Deadlock test, thread 1 attempts to obtain a SecureRandom instance. As part of this process, it calls into the RestrictedSecurity code, which invokes isPolicySunset(). This method uses LocalDate() to retrieve the current date, which in turn triggers the loading of time zone data (tzdb) from JAR files.

At the same time, thread 2 in the same test is attempting to load a JAR file to verify a signed class. The JAR loading process acquires a lock, and thread 2 is left waiting to obtain the signature verification algorithm.

This leads to a deadlock. Thread 2 holds the lock on JAR loading, waiting for the provider to initialize so it can perform signature verification. Thread 1 is blocked in RestrictedSecurity.isPolicySunset(), attempting to load tzdb from the JAR, which is already locked by thread 2.

As a result, both threads wait indefinitely, causing the test to hang. This issue is fixed by avoiding to use the LocalDate() method.

Another deadlock occurs in ProfileParser.checkHashValues().

In thread 1, when attempting to obtain a SecureRandom instance, the call proceeds past isPolicySunset() and into checkHashValues(). This method tries to obtain a MessageDigest from OpenJCEPlusFIPS. However, since OpenJCEPlusFIPS resides in a module that must be loaded, the loading process is blocked—because thread 2 has already locked the JAR loader while waiting for the signature verification algorithms to verify a signed class. As a result, the deadlock occurs.

We already had a method, isJarVerifierInStackTrace(), designed to skip the hash check if the JAR verifier is present in the stack trace. However, this method only checked the current thread’s stack. I updated it to check the stack traces of all threads, which resolves the issue.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -42,6 +42,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
+import java.util.TimeZone;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -135,15 +136,22 @@ public final class RestrictedSecurity {
     }
 
     private static boolean isJarVerifierInStackTrace() {
-        java.util.function.Predicate<Class<?>> isJarVerifier =
-                clazz -> "java.util.jar.JarVerifier".equals(clazz.getName())
-                      && "java.base".equals(clazz.getModule().getName());
+        final String targetClass = "java.util.jar.JarVerifier";
+        final String targetModule = "java.base";
 
-        java.util.function.Function<Stream<StackWalker.StackFrame>, Boolean> matcher =
-                stream -> stream.map(StackWalker.StackFrame::getDeclaringClass)
-                                .anyMatch(isJarVerifier);
-
-        return StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).walk(matcher);
+        for (java.util.Map.Entry<Thread, StackTraceElement[]> e : Thread.getAllStackTraces().entrySet()) {
+            StackTraceElement[] stack = e.getValue();
+            if (stack == null)
+                continue;
+            for (StackTraceElement ste : stack) {
+                if (targetClass.equals(ste.getClassName())) {
+                    if (targetModule.equals(ste.getModuleName())) {
+                        return true;
+                    }
+                }
+            }
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


taoliult:
@keithc-ca 

The `isJarVerifierInStackTrace()` method is invoked within `checkHashValues()` to determine whether a JAR verifier appears in the call stack. If it does, hash checking is skipped. The `checkHashValues()` method itself is only called inside the static initializer of the `sun.security.jca.Providers` class, it runs just once for the JVM. So, We don’t think checking the entire call stack of all threads will have a significant impact compared to checking only part of the caller’s stack.

For checking the policy sunset, I try to call `LocalDate.now()` as early as possible—at the beginning of the static block in the `RestrictedSecurity` class. However, this still did not fix the issue, the Deadlock test continues to fail. Additionally, we don’t think we can just simply replace `desc.sunsetDate` with a boolean flag such as `desc.isSunset`. We cannot expect clients to manually set it to false, nor can we just provide a patch that update the value to false once the profile’s sunset date has passed.

keithc-ca:
> call `LocalDate.now()` as early as possible

Some of the issues might have to do with resolving `Clock.systemDefaultZone()`; we can use `LocalDate.now(Clock.systemUTC())` instead (I think it makes sense that `sunsetDate` is independent of the system's default zone anyway).

taoliult:
> > call `LocalDate.now()` as early as possible
> 
> Some of the issues might have to do with resolving `Clock.systemDefaultZone()`; we can use `LocalDate.now(Clock.systemUTC())` instead (I think it makes sense that `sunsetDate` is independent of the system's default zone anyway).

Yes, I think UTC is a good way to keep consistent date globally, rather than relying on the user’s or system’s local time. @jasonkatonica FYI.

And I tested, replacing `LocalDate.now()` with `LocalDate.now(Clock.systemUTC())` did work — it avoids triggering the loading of time zone data (tzdb) from JAR files, so no deadlock occurs.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -135,15 +136,22 @@ private RestrictedSecurity() {
...
-                                .anyMatch(isJarVerifier);
-
-        return StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).walk(matcher);
+        final String targetClass = "java.util.jar.JarVerifier";
+        final String targetModule = "java.base";
+
+        for (java.util.Map.Entry<Thread, StackTraceElement[]> e : Thread.getAllStackTraces().entrySet()) {
+            StackTraceElement[] stack = e.getValue();
+            if (stack == null)
+                continue;
```
keithc-ca: I think this part of the change should be proposed separately (see the second point in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1097#issuecomment-3303975488).

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: Perhaps I was mistaken. Looking again, having any thread executing anything in `JarVerifier`, just delays the hash checking.

I still don't like the fact that it checks _all_ stack frames of _all_ threads, but I suppose there are _likely_ to be fewer threads before hash checking is complete, but it's not guaranteed.

pshipton:
fyi the backport PRs.

PR #1102, opened 2025-09-24
Remove categories and sort FIPS 140-3 excludes

jasonkatonica:
This update eliminates test bucket identifiers like sanity.openjdk and applies sorting to all entries within the FIPS 140-3 exclusion files. Since exclusion categories are not utilized on a per-test-bucket basis and the files are large, this change streamlines file maintenance and improves readability.


jasonkatonica:
> Which one is that?

The tool is called Alphabetize and is available in vs code. We are ok using any vscode plugin that can alphabetize lines.

keithc-ca:
VSCode is a non-starter for me. Can you not tell VSCode to use an external tool (e.g. the one I wrote)?

jasonkatonica:
OK I updated the order for those two tests that are in different order in commit https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1102/commits/9a8c22175da4ab34963d82e416f7ad5f34c34fff

keithc-ca (approved):

PR #1108, opened 2025-10-06
Exclude ReplayCacheTestProc from FIPS 140-3

jasonkatonica:
The test ReplayCacheTestProc is excluded in various other FIPS profiles and should also be excluded here in the Strongly-Enforced profile.


pshipton (approved):

jasonkatonica:
Backported this update to Java 11. Other Java releases already had this exclusion.

PR #1115, opened 2025-10-17
Update exclude list for OpenJCEPlus FIPS 140-3 Strongly Enforced profile

taoliult:
Add the following test cases across all JDK versions, as the customized security policy prevents the OpenJCEPlus library from loading, causing SHA256 MessageDigest to be unavailable during checkHashValues().

```
jdk/dynalink/BeanLinkerTest.java
jdk/dynalink/TrustedDynamicLinkerFactoryTest.java
jdk/dynalink/UntrustedDynamicLinkerFactoryTest.java
sun/security/mscapi/AccessKeyStore.java
```


keithc-ca:
Why does only one of the six OpenJCEPlus exclude lists need to change?

taoliult:
> Why does only one of the six OpenJCEPlus exclude lists need to change?

This update applies only to the Strongly Enforced profile exclusion list; the Strict and Weakly Enforced profiles already include these tests. OpenJCEPlus at first of provider list exclusion list "ProblemList-OpenJCEPlus.txt" no need to exclude these tests. The two other OpenJCEPlus TLS exclusion lists are placeholder files and are not in use yet.

keithc-ca (approved):

PR #1122, opened 2025-10-28
Fix deadlock in RestrictedSecurity caused by JAR verification and hash check

taoliult:
This PR updates isJarVerifierInStackTrace() to inspect all thread stacks, resolving the deadlock that occurred in ProfileParser.checkHashValues().

In the failing scenario, one thread (Thread-1) calls SecureRandom.getInstance(), which triggers checkHashValues() in RestrictedSecurity mode. This method attempts to obtain a MessageDigest from provider OpenJCEPlusFIPS, requiring the corresponding module to be loaded. However, another thread (Thread-2) is already holding the loader lock while performing signature verification for a signed class. Thread-1 is then blocked waiting for that module to load, while Thread-2 waits for the MessageDigest initialization to complete, resulting in a circular deadlock.

Previously, isJarVerifierInStackTrace() only checked the current thread’s stack to decide whether to skip check hash. It has now been updated to scan the stack traces of all threads, but except for the VM service threads, allowing it to detect active JAR verification in other non-VM threads and prevent potential cross-thread deadlocks.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -43,6 +43,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -82,6 +83,7 @@ public final class RestrictedSecurity {
 
     private static RestrictedSecurityProperties restricts;
 
+    private static final AtomicInteger hashPauseCount = new AtomicInteger(0);
     private static final Set<String> unmodifiableProperties = new HashSet<>();
 
     static {
@@ -164,11 +166,19 @@ public final class RestrictedSecurity {
             if (fromProviders) {
                 enableCheckHashes = true;
             }
-            if (enableCheckHashes && !isJarVerifierInStackTrace()) {
+            if (enableCheckHashes && !isJarVerifierInStackTrace() && (hashPauseCount.get() == 0)) {
                 profileParser = null;
                 parser.checkHashValues();
             }
         }
+    }
+
+    public static void pauseHashCheck() {
+        hashPauseCount.incrementAndGet();
+    }
+
+    public static void resumeHashCheck() {
+        hashPauseCount.decrementAndGet();
     }
 
     /**
--- a/src/java.base/share/classes/java/util/jar/JarFile.java
+++ b/src/java.base/share/classes/java/util/jar/JarFile.java
@@ -23,11 +23,18 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2025, 2025 All Rights Reserved
+ * ===========================================================================
+ */
+
 package java.util.jar;
 
 import jdk.internal.access.SharedSecrets;
 import jdk.internal.access.JavaUtilZipFileAccess;
 import jdk.internal.misc.ThreadTracker;
+import openj9.internal.security.RestrictedSecurity;
 import sun.security.util.ManifestEntryVerifier;
 import sun.security.util.SignatureFileVerifier;
 
@@ -1045,10 +1052,12 @@ public class JarFile extends ZipFile {
         }
         if (jv != null && !jvInitialized) {
             Object key = beginInit();
+            RestrictedSecurity.pauseHashCheck();
             try {
                 initializeVerifier();
                 jvInitialized = true;
             } finally {
+                RestrictedSecurity.resumeHashCheck();
                 endInit(key);
             }
         }
--- a/test/jdk/ProblemList-FIPS140_3_OpenJcePlus.txt
+++ b/test/jdk/ProblemList-FIPS140_3_OpenJcePlus.txt
@@ -309,7 +309,6 @@ java/security/SecureRandom/Serialize.java https://github.com/eclipse-openj9/open
 java/security/SecureRandom/SerializedSeedTest.java https://github.com/eclipse-openj9/openj9/issues/20978 generic-all
 java/security/SecureRandom/ThreadSafe.java https://github.com/eclipse-openj9/openj9/issues/20978 generic-all
 java/security/Security/CaseInsensitiveAlgNames.java https://github.com/eclipse-openj9/openj9/issues/20978 generic-all
-java/security/Security/ClassLoaderDeadlock/Deadlock.java https://github.com/eclipse-openj9/openj9/issues/21919 generic-all
 java/security/Security/ConfigFileTest.java https://github.com/eclipse-openj9/openj9/issues/20978 generic-all
 java/security/Security/ProviderFiltering.java https://github.com/eclipse-openj9/openj9/issues/20978 generic-all
 java/security/Security/removing/RemoveProviderByIdentity.java https://github.com/eclipse-openj9/openj9/issues/20978 generic-all


taoliult:
I explored alternative approaches other than scanning all threads but couldn’t find any feasible solution. I still believe that scanning all thread stacks is the most reliable way to resolve this deadlock.

The call to parser.checkHashValues() occurs only once, during the static initialization of the Providers class at JVM startup. At that early stage, the JVM typically has only a small number of active threads.

I also added an improvement to skip scanning VM service threads using the following checks:

```
Thread th = e.getKey();
boolean isVmThread = th.isDaemon() || th.getPriority() > 7;
if (isVmThread) continue; // Skip VM service threads.
```

In my test, the total number of scanned threads decreased from 18 (all threads) to 3 (after skipping VM service threads).

Additionally, the scan terminates immediately after finding the first matching, so it doesn’t traverse entire stacks unnecessarily.

So, I don’t believe scanning all thread stacks (skip scanning VM service threads) will have a noticeable performance impact compared to checking only the current thread’s stack.

keithc-ca:
Yes, I agreed that a call to `parser.checkHashValues()` will occur at most once.

However, you haven't addressed my concern that the number of threads present at the time that call occurs (if it occurs) is unbounded. I wrote a simple test that creates an arbitrary number of threads (e.g. 1000) before triggering loading the security infrastructure: tracing showed that 1019 threads stacks needed to be considered.

Perhaps we need to revisit how we manage the interaction between `JarVerifier` and `RestrictedSecurity`.

taoliult:
> Yes, I agreed that a call to `parser.checkHashValues()` will occur at most once.
> 
> However, you haven't addressed my concern that the number of threads present at the time that call occurs (if it occurs) is unbounded. I wrote a simple test that creates an arbitrary number of threads (e.g. 1000) before triggering loading the security infrastructure: tracing showed that 1019 threads stacks needed to be considered.
> 
> Perhaps we need to revisit how we manage the interaction between `JarVerifier` and `RestrictedSecurity`.

To check the potential impact of scanning all thread stacks, I created a simple test program that spawns 1,000 active threads, each with 40 stack frames before invoking the same scanning logic used in `isJarVerifierInStackTrace()`. Under this scenario, the total scan completed in roughly 11–23 ms using fyre redhat VM (average is 16 ms) with all 1,000 threads traversed.

And also, since `Thread.getAllStackTraces()` performs an in-memory snapshot without blocking or I/O, and this check runs only once during JVM initialization. So, the scan adds only a few milliseconds, which I don’t think it will have a real impact on performance. Please advise.

keithc-ca:
I was thinking something like this:
```java
    private static void checkHashValues(boolean fromProviders) {
        if (parser != null) {
            if (fromProviders) {
                enableCheckHashes = true;
            }
            if (enableCheckHashes && (hashPauseCount.get() == 0)) {
                profileParser = null;
                parser.checkHashValues();
            }
        }
    }

    private static final AtomicInteger hashPauseCount = new AtomicInteger(0);

    public static void pauseHashCheck() {
        if (profileParser != null) {
            hashPauseCount.incrementAndGet();
        }
    }

    public static void resumeHashCheck() {
        if (profileParser != null) {
            hashPauseCount.decrementAndGet();
        }
    }
```
In each method of `JarVerifier`, that are entry points to starting jar verification, we would add a pair of calls `RestrictedSecurity.pauseHashCheck();` and `RestrictedSecurity.resumeHashCheck();` (using `try / finally` as necessary). I count only ten methods in `JarVerifier` - not all of them will need to change (for example, the constructor doesn't need to change).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -164,13 +166,21 @@ private static void checkHashValues(boolean fromProviders) {
...
             }
         }
     }
 
+    public static void pauseHashCheck() {
+            hashPauseCount.incrementAndGet();
+    }
+
+    public static void resumeHashCheck() {
+            hashPauseCount.decrementAndGet();
```
keithc-ca: After the hash check is complete, there's no need to update `hashPauseCount`:
```java
        if (profileParser != null) {
            hashPauseCount.incrementAndGet();
        }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -254,7 +264,6 @@ public static boolean isServiceAllowed(Service service) {
      */
     public static boolean canServiceBeRegistered(Service service) {
         if (securityEnabled) {
-            checkHashValues(false);
```
keithc-ca: Why should this be removed?

src/java.base/share/classes/java/util/jar/JarFile.java
```diff
@@ -1046,9 +1053,11 @@ synchronized void ensureInitialization() {
         if (jv != null && !jvInitialized) {
             Object key = beginInit();
             try {
+                RestrictedSecurity.pauseHashCheck();
```
keithc-ca: This call should be made before entering the `try` statement.

I don't expect this is sufficient: I also see uses of `JarVerifier` in `JarInputStream` and `Manifest`.

keithc-ca (changes requested):

src/java.base/share/classes/java/util/jar/JarFile.java
keithc-ca: I didn't disagree with the change to `JarFile`, but you haven't convinced me that there are no scenarios involving `JarInputStream` or `Manifest` that could have one or more methods of `JarVerifier` on the call stack that were detected via `isJarVerifierInStackTrace()` prior to this change.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: I added the `profileParser != null` check as suggested, but the Deadlock test still fails for the same reason.

I believe the cause is that when the `MainThread` enters `pauseHashCheck()`, the `profileParser` has not yet finished initializing, so it’s still `null` and the pause counter isn’t incremented. Meanwhile, `Thread-1` enters `checkHashValues()` just after the `profileParser` finishes initializing. Since the pause count is still 0, it proceeds to call `parser.checkHashValues()` without skipping, which eventually triggers the deadlock.

I suggest two possible solutions:

1. Keep the current way without the `profileParser != null` check. When not in RestrictedSecurity mode, the pause counter will still be incremented and decremented in JarFile, but hash checking won’t actually occur since `profileParser` is always null in non-RestrictedSecurity mode.

2. Add a `RestrictedSecurity.isEnabled()` check in JarFile before `RestrictedSecurity.pauseHashCheck();` and `RestrictedSecurity.resumeHashCheck();` to make sure the pause counter won't changed in non-RestrictedSecurity mode.

src/java.base/share/classes/java/util/jar/JarFile.java
taoliult: Check the PR that originally introduced `isJarVerifierInStackTrace()`. It was designed to address a different issue — when a JAR verification process starts before the RestrictedSecurity profile is loaded, the hash calculation is triggered as part of that verification, causing a nested JAR verification and subsequent errors. To prevent this, the hash calculation for a profile was skipped if it was initiated by a JAR verification process and deferred until later in the loading sequence. 

Since that issue is unrelated to the current PR, I’ve reintroduced `isJarVerifierInStackTrace()` to maintain its original behavior.

For this PR, it addresses the deadlock issue. The lock isn’t on `JarVerifier` — it’s on the `JarFile`, which is what causes the deadlock. Therefore, I added the pause counter inside the synchronized `JarFile.ensureInitialization()` method to pause the hash check whenever the `JarFile` is locked.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -164,13 +166,21 @@ private static void checkHashValues(boolean fromProviders) {
             if (fromProviders) {
                 enableCheckHashes = true;
             }
-            if (enableCheckHashes && !isJarVerifierInStackTrace()) {
+            if (enableCheckHashes && !isJarVerifierInStackTrace() && (hashPauseCount.get() == 0)) {
```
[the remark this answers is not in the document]
keithc-ca: Why is it back?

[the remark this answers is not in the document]
taoliult: I added it back since the `isJarVerifierInStackTrace()` method has been restored.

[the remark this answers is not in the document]
keithc-ca: `isJarVerifierInStackTrace()` should not be restored; please use pause/resume instead.

taoliult:
> @taoliult Please share the plan for this change.

Yes, I am still working on it and will update shortly.

PR #1124, opened 2025-10-30

JasonFengJ9:

Conflicts:
```
```


src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -1578,7 +1578,7 @@ void exit() {
      * will report it via {@link #interrupted()} and {@link #isInterrupted()}.
      */
     public void interrupt() {
-        // Setting the interrupt status must be done before reading nioBlocker.
+        // Setting the interrupted status must be done before reading nioBlocker.
```
keithc-ca: Please also change "interrupt status" to "interrupted status" in lines 1659-1662.

PR #1125, opened 2025-10-30
Implement sunset policy handling with suppress and ignore options

taoliult:
Added logic to handle restricted security profile sunset behavior using -Dsemeru.restrictedsecurity.suppresssunsetwarning and -Dsemeru.restrictedsecurity.ignoresunsetexpiration system properties.

When suppresssunsetwarning is true, all sunset warning messages are suppressed; if the profile is expired and
ignoresunsetexpiration is false, the JVM exits silently with status 1.

When suppresssunsetwarning is false. If the profile has expired and ignoresunsetexpiration is false, a fatal error is printed and the JVM terminates. If expired and ignoresunsetexpiration is true, a warning message is printed indicating uncertified cryptography may be active. If the profile will expire within six months, a generic warning message is printed.

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -30,6 +30,7 @@ import java.security.Provider;
 import java.security.Provider.Service;
 import java.time.Clock;
 import java.time.LocalDate;
+import java.time.Period;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
@@ -63,6 +64,8 @@ public final class RestrictedSecurity {
     private static boolean isFIPSEnabled;
 
     private static final boolean allowSetProperties;
+    private static final boolean suppressSunsetWarning;
+    private static final boolean ignoreSunsetExpiration;
 
     private static final boolean isNSSSupported;
     private static final boolean isOpenJCEPlusCertifiedPlatform;
@@ -104,6 +107,8 @@ public final class RestrictedSecurity {
 
         userEnabledFIPS = Boolean.getBoolean("semeru.fips");
         allowSetProperties = Boolean.getBoolean("semeru.fips.allowsetproperties");
+        suppressSunsetWarning = getBooleanPropertyTreatEmptyAsTrue("semeru.restrictedsecurity.suppresssunsetwarning");
+        ignoreSunsetExpiration = getBooleanPropertyTreatEmptyAsTrue("semeru.restrictedsecurity.ignoresunsetexpiration");
 
         if (userEnabledFIPS) {
             if (isFIPSSupported) {
@@ -133,6 +138,19 @@ public final class RestrictedSecurity {
 
     private RestrictedSecurity() {
         super();
+    }
+
+    private static boolean getBooleanPropertyTreatEmptyAsTrue(String name) {
+        String systemProp = System.getProperty(name);
+        if (systemProp == null) {
+            return false;
+        }
+        if (systemProp.isEmpty()) {
+            // If set without a value (e.g., "-Dsemeru.restrictedsecurity.suppresssunsetwarning"),
+            // treat it as true and suppress the warning.
+            return true;
+        }
+        return Boolean.parseBoolean(systemProp);
     }
 
     private static boolean isJarVerifierInStackTrace() {
@@ -612,9 +630,34 @@ public final class RestrictedSecurity {
             printStackTraceAndExit("Restricted security property is null.");
         }
 
-        // Check if the SunsetDate expired.
-        if (isPolicySunset(restricts.descSunsetDate)) {
-            printStackTraceAndExit("Restricted security policy expired.");
+        int expireMonths = monthsToPolicySunset(restricts.descSunsetDate);
+
+        if (suppressSunsetWarning) {
+            if ((expireMonths <= 0) && !ignoreSunsetExpiration) {
+                System.exit(1);
+            }
+        } else if (expireMonths <= 0) { // Check if the SunsetDate expired.
+            if (ignoreSunsetExpiration) {
+                System.err.format("The requested restricted security profile %s expired on %s"
+                        + ": Certified cryptography use cannot be guaranteed.%n"
+                        + "Use -Dsemeru.restrictedsecurity.suppresssunsetwarning to stop displaying this message.%n"
+                        + "The -Dsemeru.restrictedsecurity.ignoresunsetexpiration option has been specified.%n"
+                        + "WARNING: Java will start with the requested restricted security profile but uncertified"
+                        + " cryptography may be active.%n",
+                        restricts.profileID, restricts.descSunsetDate);
+            } else {
+                printStackTraceAndExit(String.format("The requested restricted security profile %s expired on %s"
+                        + ": Java will stop because certified cryptography use cannot be guaranteed.%n"
+                        + "Use -Dsemeru.restrictedsecurity.suppresssunsetwarning to stop displaying this message.%n"
+                        + "Use -Dsemeru.restrictedsecurity.ignoresunsetexpiration to allow Java to start while"
+                        + " possibly using uncertified cryptography.%n",
+                        restricts.profileID, restricts.descSunsetDate));
+            }
+        } else if (expireMonths <= 6) { // Check if the SunsetDate will expire within 6 months.
+            System.err.format("The restricted security profile %s will expire on %s,"
+                    + " after which Java will fail to start if this profile is specified.%n"
+                    + "The latest Semeru Runtimes release may include an updated security profile.%n",
+                    restricts.profileID, restricts.descSunsetDate);
         }
 
         // Check secure random settings.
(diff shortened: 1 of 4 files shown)


taoliult:
@keithc-ca Any review suggestions for this PR?

keithc-ca:
Sorry, I have been very busy. I may not get to this today.

One of the things I noted, though, was the time to sunset is truncated down to a number of months, so it will complain (almost) a month sooner than it should.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -612,9 +618,37 @@ private static void restrictsCheck() {
...
+        } else {
+            if (expireMonths == 0) { // Check if the SunsetDate expired.
+                if (ignoreSunsetExpiration) {
+                    System.err.println("The requested restricted security profile " + restricts.profileID
+                            + " expired on " + restricts.descSunsetDate
+                            + ": java will stop because certified cryptography use cannot be guaranteed."
+                            + " Use -Dsemeru.restrictedsecurity.suppresssunsetwarning to stop displaying this message."
+                            + " The -Dsemeru.restrictedsecurity.ignoresunsetexpiration option has been specified."
+                            + " WARNING: java will start with the requested restricted security profile but uncertified"
+                            + " cryptography may be active.");
```
keithc-ca: This message is not self-consistent. It says "java will stop" and "java will start" - it can't be both.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -612,9 +618,37 @@ private static void restrictsCheck() {
...
+                } else {
+                    printStackTraceAndExit(
+                            "The requested restricted security profile " + restricts.profileID
+                                    + " expired on " + restricts.descSunsetDate
```
keithc-ca: Please join lines 638 and 639 and indent like above.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -633,25 +667,34 @@ private static void restrictsCheck() {
...
+        // If no sunset date is specified, it will not sunset.
+        if (isNullOrBlank(descSunsetDate)) {
+            return Integer.MAX_VALUE;
         }
 
-        if (debug != null) {
-            debug.println("Restricted security policy is sunset: " + isSunset);
+        try {
+            LocalDate sunsetDate = LocalDate.parse(descSunsetDate, DateTimeFormatter.ofPattern("yyyy-MM-dd"));
+            LocalDate now = LocalDate.now(Clock.systemUTC());
```
keithc-ca: Suggest only the `parse()` call should be within the `try` block:
```java
        LocalDate sunsetDate;
        try {
            sunsetDate = LocalDate.parse(descSunsetDate, DateTimeFormatter.ofPattern("yyyy-MM-dd"));
        } catch (DateTimeParseException e) {
            printStackTraceAndExit(
                    "Restricted security policy sunset date is incorrect, the correct format is yyyy-MM-dd.");
            return 0;
        }

        LocalDate now = LocalDate.now(Clock.systemUTC());

        if (sunsetDate.isBefore(now)) {
            return 0; // already sunset
        }

        Period period = Period.between(now, sunsetDate);
        int months = (period.getYears() * 12) + period.getMonths();
        if (period.getDays() > 0) {
            months += 1;
        }
        return months;
```
Notice the added parentheses, and the use of `+=` instead of `++`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -612,9 +618,37 @@ private static void restrictsCheck() {
             printStackTraceAndExit("Restricted security property is null.");
         }
 
-        // Check if the SunsetDate expired.
-        if (isPolicySunset(restricts.descSunsetDate)) {
-            printStackTraceAndExit("Restricted security policy expired.");
+        int expireMonths = monthsToPolicySunset(restricts.descSunsetDate);
+
+        if (suppressSunsetWarning) {
+            if (expireMonths == 0 && !ignoreSunsetExpiration) {
```
keithc-ca: Missing parentheses:
```java
            if ((expireMonths == 0) && !ignoreSunsetExpiration) {
```

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: Updated.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
taoliult: Updated.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -612,9 +617,36 @@ private static void restrictsCheck() {
             printStackTraceAndExit("Restricted security property is null.");
         }
 
-        // Check if the SunsetDate expired.
-        if (isPolicySunset(restricts.descSunsetDate)) {
-            printStackTraceAndExit("Restricted security policy expired.");
+        int expireMonths = monthsToPolicySunset(restricts.descSunsetDate);
+
+        if (suppressSunsetWarning) {
+            if ((expireMonths == 0) && !ignoreSunsetExpiration) {
```
keithc-ca: Suggest this and line 627 should use `<=`:
```java
            if ((expireMonths <= 0) && !ignoreSunsetExpiration) {
```

taoliult:
> Please update the commit message; the summary (first) line should be no more than 70 characters long.

Updated.

keithc-ca:
More concretely, I was thinking something like this:
```java
        if ("ignore".equals(sunsetPolicy)) {
            return;
        }

        int expireMonths = monthsToPolicySunset(restricts.descSunsetDate);

        if (expireMonths <= 0) { // SunsetDate is in the past.
            if ("quiet".equals(sunsetPolicy)) {
                System.exit(1);
            }
            printStackTraceAndExit("The requested restricted security profile " + restricts.profileID
                    + " expired on " + restricts.descSunsetDate
                    + ": java will stop because certified cryptography use cannot be guaranteed."
                    + " Use -Dsemeru.restrictedsecurity.sunsetpolicy=quiet to stop displaying this message."
                    + " Use the -Dsemeru.restrictedsecurity.sunsetpolicy=ignore to allow java to start while"
                    + " possibly using uncertified cryptography.");
        } else if ((expireMonths <= 6) && !"quiet".equals(sunsetPolicy)) {
            System.err.println("The restricted security profile " + restricts.profileID
                    + " will expire on " + restricts.descSunsetDate
                    + ", after which java will fail to start if this profile is specified."
                    + " The latest Semeru Runtimes release may include an updated security profile.");
        }
```

taoliult:
> As I was reading the updated commit message, two things occur to me:
> 
> 1. The behavior is not as described - e.g. using `-Dsemeru.restrictedsecurity.suppresssunsetwarning` does nothing (one would have to specify `-Dsemeru.restrictedsecurity.suppresssunsetwarning=true` instead.
> 2. This might be clearer and simpler if a single property were used - it eliminates the need to explain the priority of one property over the other. What do you think about updating this to recognize `-Dsemeru.restrictedsecurity.sunsetpolicy=ignore` and `-Dsemeru.restrictedsecurity.sunsetpolicy=quiet`? If the property is unset, or set to some other value the current behavior would apply.

For these two options, if they are not set, they will both default to false, which is the same as explicitly setting them to false. This matches the current behavior when the options are not provided: the sunset warning message will not be suppressed, and an expired sunset date will not be ignored. This behavior is consistent with the document: https://github.ibm.com/runtimes/jit-crypto/issues/963#issuecomment-144354566

keithc-ca (changes requested):
Please also rebase to avoid implicit merges

closed/test/jdk/openj9/internal/security/TestPolicySunset.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @ParameterizedTest
+    @MethodSource("patternMatches_expectedExitValue0")
+    public void shouldContain_expectedExitValue0(String customprofile, String securityPropertyFile, String suppresssunsetwarning, String ignoresunsetexpiration, String expected) throws Exception {
+        OutputAnalyzer outputAnalyzer = ProcessTools.executeTestJava(
+                "-Dsemeru.fips=true",
+                "-Dsemeru.customprofile=" + customprofile,
+                "-Djava.security.properties=" + securityPropertyFile,
+                "-Dsemeru.restrictedsecurity.suppresssunsetwarning" + suppresssunsetwarning,
+                "-Dsemeru.restrictedsecurity.ignoresunsetexpiration" + ignoresunsetexpiration,
```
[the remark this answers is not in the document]
taoliult: We obtain both properties using the same method, `getBooleanPropertyTreatEmptyAsTrue()`, so they follow identical parsing behavior. The existing test that verifies the case where both properties are not specified is already sufficient—there is no need to test scenarios where one property is missing and the other is present, since those paths collapse to the same boolean results.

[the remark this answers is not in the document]
keithc-ca: Yes, the expectation is that `property=false` is handled the same as if `property` is not set, but that's what the test should verify.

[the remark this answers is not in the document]
taoliult: Thanks for the review. The existing test in `TestProperties` already covers the scenario where both sunset properties are unspecified. In that case, both properties are treated as false, and the behavior functions as expected. Because an unset property is interpreted the same as `property=false`, that code path is already being exercised.

Scenarios where the properties are explicitly set to `true` or `false` are covered in `TestPolicySunset`.

Given this, I’m unsure why we need an additional test where only one property is unset. Since the “unset” state is functionally equivalent to false—and the existing tests already verify both properties unset and explicit false cases—the situation when one property is unset, it still resolves to the same boolean value, so it doesn’t add any new behavior to test.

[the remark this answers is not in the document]
keithc-ca: Yes, the existing tests don't specify those properties, but they don't do any checking related to the effects of the new properties, so I don't think that's sufficient.

closed/test/jdk/openj9/internal/security/TestPolicySunset.java
```diff
@@ -0,0 +1,180 @@
...
+        args.add("-Djava.security.properties=" + securityPropertyFile);
+        if (suppresssunsetwarning != null) {
+            args.add("-Dsemeru.restrictedsecurity.suppresssunsetwarning" + suppresssunsetwarning);
+        }
+        if (ignoresunsetexpiration != null) {
+            args.add("-Dsemeru.restrictedsecurity.ignoresunsetexpiration" + ignoresunsetexpiration);
+        }
+        args.add("TestPolicySunset");
+
+        OutputAnalyzer outputAnalyzer = ProcessTools.executeTestJava(args.toArray(new String[0]));
```
keithc-ca: Please just use the list directly (via `executeTestJava(List<String>)`:
```java
        OutputAnalyzer outputAnalyzer = ProcessTools.executeTestJava(args);
```

taoliult: Updated.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -612,9 +630,34 @@ private static void restrictsCheck() {
...
+            }
+        } else if (expireMonths <= 0) { // Check if the SunsetDate expired.
+            if (ignoreSunsetExpiration) {
+                System.err.format("The requested restricted security profile " + restricts.profileID
+                        + " expired on " + restricts.descSunsetDate
+                        + ": Certified cryptography use cannot be guaranteed.%n"
+                        + " Use -Dsemeru.restrictedsecurity.suppresssunsetwarning to stop displaying this message.%n"
+                        + " The -Dsemeru.restrictedsecurity.ignoresunsetexpiration option has been specified.%n"
+                        + " WARNING: Java will start with the requested restricted security profile but uncertified"
+                        + " cryptography may be active.");
```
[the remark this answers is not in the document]
keithc-ca: Also, use `%s` for the two parameters to avoid misinterpreting those strings as formats.

PR #1128, opened 2025-11-05
Fix AddressComputationContractTest for OpenJ9

pshipton:
Update the test to expect the OpenJ9 result rather than what appears to be a Hotspot bug.

Depends on https://github.com/eclipse-openj9/openj9/pull/22890

Issue https://github.com/eclipse-openj9/openj9/issues/22862

Tested via https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/56054


keithc-ca:
This need not stay in draft mode. I think this change is safe to merge now, the test just won't pass until https://github.com/eclipse-openj9/openj9/pull/22890 is merged.

PR #1134, opened 2025-11-10
Bundle zlib on Linux except s390x

pshipton:
This addresses https://github.com/ibmruntimes/Semeru-Runtimes/issues/114.

### DIFF
--- a/make/autoconf/lib-bundled.m4
+++ b/make/autoconf/lib-bundled.m4
@@ -24,7 +24,7 @@
 #
 
 # ===========================================================================
-# (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+# (c) Copyright IBM Corp. 2021, 2025 All Rights Reserved
 # ===========================================================================
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


pshipton:
@zl-wang We can leave the system zlib enabled for AIX if that is your preference.

zl-wang:
@pshipton maybe just go ahead with this bundling for the time being (even for AIX), since zlibNX bug is still outstanding on older 7.2 & 7.3 TLs (I was told it has been fixed in the latest 7.3 TL at least).

joransiu:
The bundled zlib source in this repo does not have the s390x NXU hardware acceleration support. The distros have patched their respective system zlib with the acceleration. Given the dramatic performance difference, I suspect we still prefer to use system zlib for Linux on Z by default.

pshipton:
For now I haven't changed the behaviour on AIX, as I don't think we have issues with the default zlib and using NX requires adding an option.

keithc-ca:
This is also unrelated to https://github.com/eclipse-openj9/openj9/issues/21054 which is about Linux on s390x which is unaffected by this change.

PR #1135, opened 2025-11-12
Add tests for property file list and RestrictedSecurity service checks

taoliult:
This PR adds test coverage for 2 PRs:

1. https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1069

2. https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1078

### DIFF
--- a/closed/test/jdk/openj9/internal/security/TestProperties.java
+++ b/closed/test/jdk/openj9/internal/security/TestProperties.java
@@ -33,6 +33,7 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
+import java.io.File;
 import java.security.Provider;
 import java.security.Security;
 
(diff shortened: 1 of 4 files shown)


closed/test/jdk/openj9/internal/security/TestProperties.java
```diff
@@ -43,6 +43,8 @@
 
 public class TestProperties {
 
+    static String fileSep = java.io.File.pathSeparator;
```
keithc-ca: This is a misleading name; suggest just using `File.pathSeparator` directly instead.

closed/test/jdk/openj9/internal/security/TestProperties.java
```diff
@@ -164,6 +166,28 @@ private static Stream<Arguments> patternMatches_expectedExitValue1() {
         );
     }
 
+    private static Stream<Arguments> patternMatches_PropertiesList() {
+        return Stream.of(
+                Arguments.of("Test-Profile-Property-List.A",
+                        System.getProperty("test.src") + "/propertyListA-java.security",
+                        "(?s)(?=.*OpenJCEPlusFIPS)(?=.*Sun)(?=.*SunJSSE)(?=.*SunEC)",
```
keithc-ca: Some comments would be welcome explaining patterns such as this - it's not clear (to me anyway) what this is intended to match.

closed/test/jdk/openj9/internal/security/TestProperties.java
```diff
@@ -190,6 +214,19 @@ public void shouldContain_expectedExitValue1(String customprofile, String securi
         outputAnalyzer.shouldHaveExitValue(1).shouldMatch(expected);
     }
 
+    @ParameterizedTest
+    @MethodSource("patternMatches_PropertiesList")
+    public void shouldContain_PropertiesList(String customprofile, String securityPropertyFileList,
```
keithc-ca: For consistency, I suggest a lower-case "p" should be used:
```java
    @MethodSource("patternMatches_propertiesList")
    public void shouldContain_propertiesList(String customprofile, String securityPropertyFileList,
```

keithc-ca (changes requested):

closed/test/jdk/openj9/internal/security/TestProperties.java
keithc-ca: I suggest the comments need not repeat `"-Djava.security.propertiesList=" test` and should follow the style of existing comments. So, for example, this would become:
```java
                // 1 - The profile in propertyListA-java.security extends the profile
                // in the main java.security file, which lists 4 providers.
```
Notice also the comment starts with a capital and ends with a period (again like pre-existing comments).

keithc-ca (changes requested):

closed/test/jdk/openj9/internal/security/TestProperties.java
taoliult: Updated.

keithc-ca (changes requested):

keithc-ca:

taoliult:
Still working on JDK8 PR, will update shortly.

PR #1137, opened 2025-11-13
Support shutdown of the BufferCleaner thread

pshipton:
Fixes https://github.com/eclipse-openj9/openj9/issues/22931

Depends on https://github.com/eclipse-openj9/openj9/pull/22939

### DIFF
--- a/src/java.base/share/classes/java/nio/BufferCleaner.java
+++ b/src/java.base/share/classes/java/nio/BufferCleaner.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2025, 2025 All Rights Reserved
+ * ===========================================================================
+ */
+
 package java.nio;
 
 import java.lang.ref.PhantomReference;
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/nio/BufferCleaner.java
```diff
@@ -212,6 +220,9 @@ public void run() {
                     Cleaner c = (Cleaner) queue.remove();
                     c.clean();
                 } catch (InterruptedException e) {
+                    if (CleanerShutdown.shuttingDown()) {
+                        break;
+                    }
```
keithc-ca: I think this should go after the comment (not that it makes any functional difference).

keithc-ca (changes requested):

src/java.base/share/classes/java/nio/BufferCleaner.java
keithc-ca: Actually, I think the comment should just go: We're not ignoring the interrupt, but instead deciding whether we're done.

Please correct the indentation of line 226.

keithc-ca (changes requested):

PR #1149, opened 2026-01-02
Exclude TestTLS12 test in FIPS140-3 profiles

sabrinajlee:
This test fails as follows :
```
Execution failed: `main' threw exception: java.lang.SecurityException:
Property 'jdk.tls.disabledAlgorithms' cannot be set programmatically
when in FIPS mode
```

This test should be excluded accordingly.

Signed-off-by: Sabrina Lee <sabrinalee@ibm.com>


keithc-ca (approved):

sabrinajlee:
@keithc-ca Updated.

PR #1151, opened 2026-01-05
Change provider supporting TLS algorithms in FIPS strict mode

sabrinajlee:
These algorithms are being no longer being supported in OpenJCEPlus due to the deprecation of their internal spec classes. The Sun provider implementation will be used instead in the FIPS profile.

Signed-off-by: Sabrina Lee <sabrinalee@ibm.com>


keithc-ca (changes requested):

keithc-ca:
Please fix the typo in the commit message and the summary here (support -> supported).

keithc-ca:
I find the title somewhat misleading: it doesn't "remove" any algorithms, but instead changes the provider of some algorithms. Please update the commit message and the text here to better reflect the change.

keithc-ca:
Update hash verified locally.

keithc-ca (approved):

PR #1156, opened 2026-01-16
Add build flag for compact layouts

theresa-m:
Related:
https://github.com/eclipse-openj9/openj9/issues/21251
https://github.com/eclipse-openj9/openj9/pull/23206

Port from https://github.com/ibmruntimes/openj9-openjdk-jdk.valuetypes/pull/25


theresa-m:
> This is safe to merge ahead of uses in OpenJ9; the only downside appears to be a new warning:
> 
> ```
> CMake Warning:
>   Manually-specified variables were not used by the project:
> 
>     J9VM_OPT_VALHALLA_COMPACT_LAYOUTS
> ```
> 
> This group of flags is either all ON or all OFF: Do we anticipate a time when only a (strict) subset will be ON?

Yes they will most likely be enabled for the main project at different times.

PR #1157, opened 2026-01-16
Filter brainpool signature schemes by default

jasonkatonica:
Semeru supports the brainpool `ecdsa_brainpoolP512r1tls13_sha512`
signature scheme. The expectation is that users must explicitly enable
this feature in order to enable this signature scheme. This
must be done using either the `jdk.tls.client.SignatureSchemes` or
`jdk.tls.server.SignatureSchemes` properties. By default no brainpool
signature schemes are enabled.

Fixes: https://github.com/eclipse-openj9/openj9/issues/23177

### DIFF
--- a/src/java.base/share/classes/sun/security/ssl/SignatureScheme.java
+++ b/src/java.base/share/classes/sun/security/ssl/SignatureScheme.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * (c) Copyright IBM Corp. 2024, 2026 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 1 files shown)


test/jdk/sun/security/ssl/SignatureScheme/DefaultSSLConfigSignatureSchemes.java
```diff
@@ -58,6 +58,7 @@ public class DefaultSSLConfigSignatureSchemes extends SSLEngineTemplate {
                     "ecdsa_secp256r1_sha256",
                     "ecdsa_secp384r1_sha384",
                     "ecdsa_secp521r1_sha512",
+                    "ecdsa_brainpoolP512r1tls13_sha512",
```
keithc-ca: This scheme is excluded unless explicitly specified; see https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/src/java.base/share/classes/sun/security/ssl/SignatureScheme.java#L432. Unless that changes, I don't see how this test will not start failing.

jasonkatonica: Likely this would involve writting similar blocking code in https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/src/java.base/share/classes/javax/net/ssl/SSLParameters.java , I dont see any updates similar to what you sent in that code.

Also the user facing doc we sent to the client would be located here: https://www.ibm.com/support/pages/node/7171766

Its not specific enough to help us decide, but I would be OK adding blocking code if we think its the right thing to do to fix the test.

keithc-ca: Ah, I had confused the set of default signature schemes with the defaults for SSLParameters.

This change should add an IBM copyright notice to the modified file.

I'm not sure it makes sense to omit brainpool from the defaults for SSLParameters; perhaps you can remind us why it was excluded in https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/0370f64ebb0f25316caba556ed7ec52376c2bee3?

keithc-ca (approved):

PR #1177, opened 2026-02-17
Update CheckPBEKeySize test to use the correct provider

dev-koan:
Ensure the Cipher instance is obtained from SunJCE provider, as the class PKCS12PBECipherCore$PBEWithSHA1AndDESede is specific to it.

Signed-off-by: Dev Agarwal dev.agarwal@ibm.com

### DIFF
--- a/test/jdk/com/sun/crypto/provider/Cipher/PBE/CheckPBEKeySize.java
+++ b/test/jdk/com/sun/crypto/provider/Cipher/PBE/CheckPBEKeySize.java
@@ -21,6 +21,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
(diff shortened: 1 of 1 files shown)


keithc-ca:
Is this because the provider might be OpenJCEPlus? Why does it [answer 168 for the key size](https://github.com/ibmruntimes/OpenJCEPlus/blob/semeru-main/src/main/java/com/ibm/crypto/plus/provider/DESedeCipher.java#L116)? Doesn't DESede only involve two 56-bit keys, making the correct (effective) key size 112? Wouldn't a better solution be to correct OpenJCEPlus?

keithc-ca:
Jenkins compile plinux jdknext

keithc-ca (approved):

PR #1183, opened 2026-03-02
Update exclude list for FIPS 140-3 profile testing

taoliult:
Add multiple tests to the exclusion list for FIPS 140-3 profile testing across all JDK versions.

The excluded tests depend on functionality that is restricted or unavailable under FIPS 140-3 enforcement, including:

- Programmatic modification of restricted security properties
- Use of non-approved or unsupported algorithms
- Unsupported keystore types in FIPS profiles
- PKCS11 provider unavailability in 140-3 mode

These failures are expected under FIPS 140-3 enforcement and the affected tests have been added to the exclusion list.

### DIFF
--- a/test/jdk/ProblemList-FIPS140_3_OpenJCEPlusFIPS.FIPS140-3.txt
+++ b/test/jdk/ProblemList-FIPS140_3_OpenJCEPlusFIPS.FIPS140-3.txt
@@ -70,6 +70,7 @@ com/sun/crypto/provider/Cipher/ChaCha20/ChaCha20Poly1305ParamTest.java https://g
 com/sun/crypto/provider/Cipher/ChaCha20/OutputSizeTest.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
 com/sun/crypto/provider/Cipher/ChaCha20/unittest/ChaCha20CipherUnitTest.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
 com/sun/crypto/provider/Cipher/ChaCha20/unittest/ChaCha20Poly1305ParametersUnitTest.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
+com/sun/crypto/provider/Cipher/ChaCha20/UpdateAADTest.java https://github.com/eclipse-openj9/openj9/issues/23424 generic-all
 com/sun/crypto/provider/Cipher/CTR/CounterMode.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
 com/sun/crypto/provider/Cipher/CTS/CTSMode.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
 com/sun/crypto/provider/Cipher/DES/DesAPITest.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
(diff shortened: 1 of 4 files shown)


test/jdk/ProblemList-FIPS140_3_OpenJCEPlusFIPS.FIPS140-3.txt
```diff
@@ -540,7 +550,8 @@ javax/net/ssl/FixingJavadocs/ImplicitHandshake.java https://github.com/eclipse-o
 javax/net/ssl/FixingJavadocs/KMTMGetNothing.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
 javax/net/ssl/FixingJavadocs/SSLSessionNulls.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
 javax/net/ssl/FixingJavadocs/SSLSocketInherit.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
-javax/net/ssl/HttpsURLConnection/CriticalSubjectAltName.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
+javax/net/ssl/HttpsURLConnection/CriticalSubjectAltName.java#tls12 https://github.com/eclipse-openj9/openj9/issues/23424 generic-all
+javax/net/ssl/HttpsURLConnection/CriticalSubjectAltName.java#tls13 https://github.com/eclipse-openj9/openj9/issues/23424 generic-all
```
keithc-ca: I don't think we want this change. In Java version 25 and earlier, neither `tls12` nor `tls13` test exists so this will have the effect of un-excluding the test. In Java version 26 and later this has no effect (excluding all sub-tests is the same as excluding them all). Updating the issue URL is reasonable and welcome.

Same comment applies to all modified exclusion lists.

test/jdk/ProblemList-FIPS140_3_OpenJCEPlusFIPS.FIPS140-3.txt
```diff
@@ -1215,7 +1246,11 @@ sun/security/x509/EDIPartyName/NullName.java https://github.com/eclipse-openj9/o
 sun/security/x509/Extensions/IllegalExtensions.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
 sun/security/x509/OtherName/Parse.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
 sun/security/x509/URICertStore/AIACertTimeout.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
-sun/security/x509/URICertStore/CRLReadTimeout.java https://github.com/eclipse-openj9/openj9/issues/20343 generic-all
+sun/security/x509/URICertStore/CRLReadTimeout.java#0 https://github.com/eclipse-openj9/openj9/issues/23424 generic-all
+sun/security/x509/URICertStore/CRLReadTimeout.java#1 https://github.com/eclipse-openj9/openj9/issues/23424 generic-all
+sun/security/x509/URICertStore/CRLReadTimeout.java#2 https://github.com/eclipse-openj9/openj9/issues/23424 generic-all
+sun/security/x509/URICertStore/CRLReadTimeout.java#3 https://github.com/eclipse-openj9/openj9/issues/23424 generic-all
+sun/security/x509/URICertStore/CRLReadTimeout.java#4 https://github.com/eclipse-openj9/openj9/issues/23424 generic-all
```
keithc-ca: Other than the change to the issue URL, this seems to have no effect. Why should list the specific sub-tests?

taoliult: Since the exclude list references these specific subtest names when excluding them, so I updated this section to include the corresponding subtest names as well.

test/jdk/ProblemList-FIPS140_3_OpenJCEPlusFIPS.FIPS140-3.txt
taoliult: Line 543 for excluding test "`javax/net/ssl/HttpsURLConnection/CriticalSubjectAltName.java`" must stay, since it excludes this test for JDK25 and earlier versions. 

Line 553 and 554 also must stay for JDK26 and later versions, since once the test is split into subtests, the exclude list must reference the specific subtest names when excluding them.

PR #1197, opened 2026-03-30
Check services in legacy security providers

KostasTsiounis:
Legacy security providers do not use the more modern getService() and putService() methods, but rather other more map based methods.

Further checks are added to restrict through RestrictedSecurity the registration and retrieval of services through legacy providers.

Tests are, also, added to verify these checks.

### DIFF
--- a/closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
+++ b/closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
@@ -1,0 +1,248 @@
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2026, 2026 All Rights Reserved
+ * ===========================================================================
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * IBM designates this particular file as subject to the "Classpath" exception
+ * as provided by IBM in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * ===========================================================================
+ */
+
+/*
+ * @test
+ * @summary Test Restricted Security Mode with Legacy Providers
+ * @library /test/lib
+ * @run junit TestLegacyProviders
+ */
+import org.junit.jupiter.api.Test;
+
+import java.security.Provider;
+import java.security.Security;
+import java.util.Collection;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import jdk.test.lib.LegacyProvider;
+
+import jdk.test.lib.process.OutputAnalyzer;
+import jdk.test.lib.process.ProcessTools;
+
+public class TestLegacyProviders {
+    private static Provider legacyProvider;
+
+    private static void useLegacyPaths() throws Exception {
+        // Check get operation.
+        String response = (String) legacyProvider.get("Signature.Test");
+        if (!"example.class.MyClass".equals(response)) {
+            throw new RuntimeException("Incorrect response for get()");
+        }
+
+        response = (String) legacyProvider.get("Signature.Test2");
+        if (!"example.class.MyClass2".equals(response)) {
+            throw new RuntimeException("Incorrect response for get()");
+        }
+
+        response = (String) legacyProvider.get("Signature.Test3");
+        if (response != null) {
+            throw new RuntimeException("Incorrect response for get()");
+        }
+
+        response = (String) legacyProvider.get("Signature.Test4");
+        if (!"example.class.MyClass4".equals(response)) {
+            throw new RuntimeException("Incorrect response for get()");
+        }
+
+        response = (String) legacyProvider.get("Signature.Test5");
+        if (!"example.class.MyClass5".equals(response)) {
+            throw new RuntimeException("Incorrect response for get()");
+        }
+
+        String[] acceptedValues = {"example.class.MyClass",
+                                   "example.class.MyClass2",
+                                   "example.class.MyClass4",
+                                   "example.class.MyClass5"};
+
+        // Check entrySet operation.
+        Set<Map.Entry<Object, Object>> es = legacyProvider.entrySet();
+        for (Map.Entry<Object, Object> entry : es) {
+            String stringValue = (String) entry.getValue();
+            boolean found = false;
+            for (String acceptedValue : acceptedValues) {
+                if (stringValue.equals(acceptedValue)
+                    || "1".equals(stringValue)                                      // Version from constructor
+                    || "jdk.test.lib.LegacyProvider".equals(stringValue)            // Classname from constructor
+                    || "LegacyProvider".equals(stringValue)                         // Provider name from constructor
+                    || "Test provider".equals(stringValue)                          // Info from constructor
+                ) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found) {
+                throw new RuntimeException("Incorrect values returned from entrySet()");
+            }
+        }
+
+        // Check values operation.
+        Collection<Object> values = legacyProvider.values();
+        for (Object value : values) {
+            String stringValue = (String) value;
+            boolean found = false;
+            for (String acceptedValue : acceptedValues) {
+                if (stringValue.equals(acceptedValue)
+                    || "1".equals(stringValue)                                      // Version from constructor
+                    || "jdk.test.lib.LegacyProvider".equals(stringValue)            // Classname from constructor
+                    || "LegacyProvider".equals(stringValue)                         // Provider name from constructor
+                    || "Test provider".equals(stringValue)                          // Info from constructor
+                ) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found) {
+                throw new RuntimeException("Incorrect values returned from values()");
+            }
+        }
+
+        // Check elements operation.
+        Enumeration<Object> elements = legacyProvider.elements();
+        while (elements.hasMoreElements()) {
+            String stringElement = (String) elements.nextElement();
+            boolean found = false;
+            for (String acceptedValue : acceptedValues) {
+                if (stringElement.equals(acceptedValue)
+                    || "1".equals(stringElement)                                    // Version from constructor
+                    || "jdk.test.lib.LegacyProvider".equals(stringElement)          // Classname from constructor
+                    || "LegacyProvider".equals(stringElement)                       // Provider name from constructor
+                    || "Test provider".equals(stringElement)                        // Info from constructor
+                ) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found) {
+                throw new RuntimeException("Incorrect values returned from elements()");
+            }
+        }
+
+        // Check getProperty operation.
+        response = (String) legacyProvider.getProperty("Signature.Test");
+        if (!"example.class.MyClass".equals(response)) {
+            throw new RuntimeException("Incorrect response for getProperty()");
+        }
+
+        response = (String) legacyProvider.getProperty("Signature.Test2");
+        if (!"example.class.MyClass2".equals(response)) {
+            throw new RuntimeException("Incorrect response for getProperty()");
+        }
+
+        response = (String) legacyProvider.getProperty("Signature.Test3");
+        if (response != null) {
+            throw new RuntimeException("Incorrect response for getProperty()");
+        }
+
+        response = (String) legacyProvider.getProperty("Signature.Test4");
+        if (!"example.class.MyClass4".equals(response)) {
+            throw new RuntimeException("Incorrect response for getProperty()");
+        }
+
+        response = (String) legacyProvider.getProperty("Signature.Test5");
+        if (!"example.class.MyClass5".equals(response)) {
+            throw new RuntimeException("Incorrect response for getProperty()");
+        }
+
+        // Check compute operation.
+        response = (String) legacyProvider.compute("Signature.Test", (a, b) -> (b + "New"));
+        if (!"example.class.MyClassNew".equals(response)) {
+            throw new RuntimeException("Incorrect response for compute()");
+        }
+
+        response = (String) legacyProvider.compute("Signature.Test3", (a, b) -> (b + "New"));
+        if (response != null) {
+            throw new RuntimeException("Incorrect response for compute()");
+        }
+
+        // Check computeIfAbsent operation.
+        legacyProvider.remove("Signature.Test");
+        response = (String) legacyProvider.computeIfAbsent("Signature.Test", a -> "example.class.MyClass");
+        if (!"example.class.MyClass".equals(response)) {
+            throw new RuntimeException("Incorrect response for computeIfAbsent()");
+        }
+
+        response = (String) legacyProvider.computeIfAbsent("Signature.Test3", a -> "example.class.MyClass3");
+        if (response != null) {
+            throw new RuntimeException("Incorrect response for computeIfAbsent()");
+        }
+
+        // Check computeIfPresent operation.
+        response = (String) legacyProvider.computeIfPresent("Signature.Test", (a, b) -> (b + "New"));
+        if (!"example.class.MyClassNew".equals(response)) {
+            throw new RuntimeException("Incorrect response for computeIfPresent()");
+        }
+
+        response = (String) legacyProvider.computeIfPresent("Signature.Test3", (a, b) -> (b + "New"));
+        if (response != null) {
+            throw new RuntimeException("Incorrect response for computeIfPresent()");
+        }
+
+        String[] acceptedKeys = {"Signature.Test",
+                                 "Signature.Test2",
+                                 "Signature.Test4",
+                                 "Signature.Test5"};
+
+        // Check keys operation.
+        Enumeration<Object> keys = legacyProvider.keys();
+        while (keys.hasMoreElements()) {
+            String stringKey = (String) keys.nextElement();
+            boolean found = false;
+            for (String acceptedKey : acceptedKeys) {
+                if (stringKey.equals(acceptedKey)
+                    || "Provider.id version".equals(stringKey)      // Version from constructor
+                    || "Provider.id className".equals(stringKey)    // Provider classname from constructor
+                    || "Provider.id info".equals(stringKey)         // Info from constructor
+                    || "Provider.id name".equals(stringKey)         // Provider name from constructor
+                ) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found) {
+                throw new RuntimeException("Incorrect values returned from keys()");
+            }
+        }
+    }
+
+    @Test
+    public void runWithConstraints() throws Exception {
+        OutputAnalyzer outputAnalyzer = ProcessTools.executeTestJava(
+                "-Dsemeru.customprofile=TestLegacyProviders.Version",
+                "-Djava.security.properties=" + System.getProperty("test.src") + "/legacy-java.security",
+                "TestLegacyProviders"
+        );
+        outputAnalyzer.reportDiagnosticSummary();
+        outputAnalyzer.shouldHaveExitValue(0);
+    }
+
+    public static void main(String[] args) throws Exception {
+        Security.getProviders();
+        Security.addProvider(new LegacyProvider());
+        legacyProvider = Security.getProvider("LegacyProvider");
+        useLegacyPaths();
+    }
+}
--- a/closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
+++ b/closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
@@ -73,71 +73,39 @@ public class TestLegacyProviders {
             throw new RuntimeException("Incorrect response for get()");
         }
 
-        String[] acceptedValues = {"example.class.MyClass",
-                                   "example.class.MyClass2",
-                                   "example.class.MyClass4",
-                                   "example.class.MyClass5"};
+        Set<String> acceptedValues = Set.of(
+                "example.class.MyClass",
+                "example.class.MyClass2",
+                "example.class.MyClass4",
+                "example.class.MyClass5",
+                "1",                                                         // version from constructor
+                "jdk.test.lib.LegacyProvider",                               // class name from constructor
+                "LegacyProvider",                                            // provider name from constructor
+                "Test provider");                                            // info from constructor
 
         // Check entrySet operation.
         Set<Map.Entry<Object, Object>> es = legacyProvider.entrySet();
         for (Map.Entry<Object, Object> entry : es) {
-            String stringValue = (String) entry.getValue();
-            boolean found = false;
-            for (String acceptedValue : acceptedValues) {
-                if (stringValue.equals(acceptedValue)
-                    || "1".equals(stringValue)                                      // Version from constructor
-                    || "jdk.test.lib.LegacyProvider".equals(stringValue)            // Classname from constructor
-                    || "LegacyProvider".equals(stringValue)                         // Provider name from constructor
-                    || "Test provider".equals(stringValue)                          // Info from constructor
-                ) {
-                    found = true;
-                    break;
-                }
-            }
-            if (!found) {
-                throw new RuntimeException("Incorrect values returned from entrySet()");
+            Object value = entry.getValue();
+            if (!acceptedValues.contains(value)) {
+                throw new RuntimeException("Incorrect value returned from entrySet()");
             }
         }
 
         // Check values operation.
         Collection<Object> values = legacyProvider.values();
         for (Object value : values) {
-            String stringValue = (String) value;
-            boolean found = false;
-            for (String acceptedValue : acceptedValues) {
-                if (stringValue.equals(acceptedValue)
-                    || "1".equals(stringValue)                                      // Version from constructor
-                    || "jdk.test.lib.LegacyProvider".equals(stringValue)            // Classname from constructor
-                    || "LegacyProvider".equals(stringValue)                         // Provider name from constructor
-                    || "Test provider".equals(stringValue)                          // Info from constructor
-                ) {
-                    found = true;
-                    break;
-                }
-            }
-            if (!found) {
-                throw new RuntimeException("Incorrect values returned from values()");
+            if (!acceptedValues.contains(value)) {
+                throw new RuntimeException("Incorrect value returned from values()");
             }
         }
 
         // Check elements operation.
         Enumeration<Object> elements = legacyProvider.elements();
         while (elements.hasMoreElements()) {
-            String stringElement = (String) elements.nextElement();
-            boolean found = false;
-            for (String acceptedValue : acceptedValues) {
-                if (stringElement.equals(acceptedValue)
-                    || "1".equals(stringElement)                                    // Version from constructor
-                    || "jdk.test.lib.LegacyProvider".equals(stringElement)          // Classname from constructor
-                    || "LegacyProvider".equals(stringElement)                       // Provider name from constructor
-                    || "Test provider".equals(stringElement)                        // Info from constructor
-                ) {
-                    found = true;
-                    break;
-                }
-            }
-            if (!found) {
-                throw new RuntimeException("Incorrect values returned from elements()");
+            Object value = elements.nextElement();
+            if (!acceptedValues.contains(value)) {
+                throw new RuntimeException("Incorrect value returned from elements()");
             }
         }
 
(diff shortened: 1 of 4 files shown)


keithc-ca (changes requested):

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -428,6 +434,27 @@ public synchronized Set<Map.Entry<Object,Object>> entrySet() {
         if (entrySetCallCount != 2)
             throw new RuntimeException("Internal error.");
 
+        if (RestrictedSecurity.isEnabled()) {
+            Set<Map.Entry<Object,Object>> toRemove = new HashSet<>();
+            for (Map.Entry<Object,Object> entry : entrySet) {
+                if (isProviderInfoKey(entry.getKey())) {
+                    // This is a value pertaining to Provider info.
```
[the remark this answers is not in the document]
KostasTsiounis: Changed to lowercase.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -452,7 +480,25 @@ public Set<Object> keySet() {
...
+        Collection<Object> values = super.values();
+
+        if (RestrictedSecurity.isEnabled()) {
+            Set<Map.Entry<Object, Object>> entrySet = entrySet();
+            for (Map.Entry<Object, Object> entry : entrySet) {
+                if (isProviderInfoKey(entry.getKey())) {
+                    // This is a value pertaining to provider info.
+                    continue;
+                }
+                Service service = createServiceFromKey(entry.getKey());
```
keithc-ca: Suggest a local to avoid calling `getKey()` twice:
```java
                Object key = entry.getKey();
                if (isProviderInfoKey(key)) {
                    // This is a value pertaining to provider info.
                    continue;
                }
                Service service = createServiceFromKey(key);
```

keithc-ca (changes requested):

closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
```diff
@@ -0,0 +1,248 @@
...
+                    || "Test provider".equals(stringValue)                          // Info from constructor
+                ) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found) {
+                throw new RuntimeException("Incorrect values returned from entrySet()");
+            }
+        }
```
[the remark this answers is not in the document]
KostasTsiounis: Used suggested way in `entrySet()`, `values()`, `elements()` and `keys()`.

The point is to check that only the appropriate entries are allowed to be retrieved, so checking either of the key or value is enough, given that we know what the combinations are.

closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
```diff
@@ -0,0 +1,248 @@
...
+                "-Dsemeru.customprofile=TestLegacyProviders.Version",
+                "-Djava.security.properties=" + System.getProperty("test.src") + "/legacy-java.security",
+                "TestLegacyProviders"
+        );
+        outputAnalyzer.reportDiagnosticSummary();
+        outputAnalyzer.shouldHaveExitValue(0);
+    }
+
+    public static void main(String[] args) throws Exception {
+        Security.getProviders();
```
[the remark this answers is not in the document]
KostasTsiounis: Removed. It was left from a previous iteration of the test where I was instantiating and using the provider directly.

test/lib/jdk/test/lib/LegacyProvider.java
```diff
@@ -0,0 +1,51 @@
...
+        putIfAbsent("Signature.Test4", "example.class.MyClass4");
+
+        replace("Signature.Test2", "example.class.MyClass2New");
+        replace("Signature.Test2", "example.class.MyClass2New", "example.class.MyClass2");
+
+        replace("Signature.Test3", "example.class.MyClass3New");
+        replace("Signature.Test3", "example.class.MyClass3", "example.class.MyClass3New");
+
+        merge("Signature.Test5", "example.class.MyClass5", (a, b) -> b);
+
```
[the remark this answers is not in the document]
KostasTsiounis: Removed.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -162,6 +162,12 @@ public abstract class Provider extends Properties {
 
     private static final Object[] EMPTY = new Object[0];
 
+    private static final List<String> providerInfoKeys = new ArrayList<>(Arrays.asList(
+                                                                    "Provider.id name",
+                                                                    "Provider.id version",
+                                                                    "Provider.id info",
+                                                                    "Provider.id className"));
```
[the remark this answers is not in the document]
KostasTsiounis: Changed to suggestion.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Changed to suggestion.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -428,6 +434,27 @@ public synchronized Set<Map.Entry<Object,Object>> entrySet() {
...
+                if ((s != null) && !RestrictedSecurity.isServiceAllowed(s)) {
+                    // We're in restricted security mode which does not allow this service,
+                    // mark it to be removed from returned set.
+                    toRemove.add(entry);
+                }
+            }
+            if (!toRemove.isEmpty()) {
+                Set<Map.Entry<Object,Object>> modifiableSet = new HashSet<>(entrySet);
+                modifiableSet.removeAll(toRemove);
+                entrySet = Collections.unmodifiableSet(modifiableSet);
```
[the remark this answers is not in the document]
keithc-ca: The call to `entrySet()` on line 424 makes an unmodifiable set from the value returned on (what was) line 454.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -677,12 +744,42 @@ public Enumeration<Object> keys() {
...
-        return super.elements();
+        Enumeration<Object> elements;
+
+        if (RestrictedSecurity.isEnabled()) {
+            List<Object> list = new ArrayList<>();
+            Set<Map.Entry<Object, Object>> entrySet = entrySet();
+            for (Map.Entry<Object, Object> entry : entrySet) {
+                Service service = createServiceFromKey(entry.getKey());
+                if ((service == null) || isProviderInfoKey(entry.getKey())
+                        || RestrictedSecurity.isServiceAllowed(service)) {
```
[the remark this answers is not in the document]
KostasTsiounis: Changed to suggestion.

closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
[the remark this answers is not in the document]
keithc-ca: A test shouldn't _assume_ good behavior of the API it is testing: it should _verify_ it.

closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
```diff
@@ -0,0 +1,208 @@
...
+
+        Set<String> acceptedValues = Set.of(
+                "example.class.MyClass",
+                "example.class.MyClass2",
+                "example.class.MyClass4",
+                "example.class.MyClass5",
+                "1",                                                         // version from constructor
+                "jdk.test.lib.LegacyProvider",                               // class name from constructor
+                "LegacyProvider",                                            // provider name from constructor
+                "Test provider");                                            // info from constructor
```
keithc-ca: Suggest `acceptedKeys` should go here, so it can be used when validating `entrySet()` behavior. The comments need not be separated by quite that much whitespace.
```java
        Set<String> acceptedKeys = Set.of(
                "Signature.Test",
                "Signature.Test2",
                "Signature.Test4",
                "Signature.Test5",
                "Provider.id version",          // version from constructor
                "Provider.id className",        // provider class name from constructor
                "Provider.id info",             // info from constructor
                "Provider.id name");            // provider name from constructor

        Set<String> acceptedValues = Set.of(
                "example.class.MyClass",
                "example.class.MyClass2",
                "example.class.MyClass4",
                "example.class.MyClass5",
                "1",                            // version from constructor
                "jdk.test.lib.LegacyProvider",  // class name from constructor
                "LegacyProvider",               // provider name from constructor
                "Test provider");               // info from constructor
```

closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
```diff
@@ -0,0 +1,208 @@
...
+
+        // Check entrySet operation.
+        Set<Map.Entry<Object, Object>> es = legacyProvider.entrySet();
+        for (Map.Entry<Object, Object> entry : es) {
+            Object value = entry.getValue();
```
keithc-ca, line 104: This should also validate keys:
```java
            Object key = entry.getValue();
            if (!acceptedKeys.contains(key)) {
                throw new RuntimeException("Incorrect key returned from entrySet()");
            }
```

keithc-ca (changes requested):

closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
KostasTsiounis: Moved.

closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
KostasTsiounis: Added key validation in `entrySet()` check.

closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
```diff
@@ -0,0 +1,213 @@
...
+            throw new RuntimeException("Incorrect response for get()");
+        }
+
+        Set<String> acceptedKeys = Set.of(
+                "Signature.Test",
+                "Signature.Test2",
+                "Signature.Test4",
+                "Signature.Test5",
+                "Provider.id version",          // version from constructor
+                "Provider.id className",        // provider classname from constructor
```
keithc-ca: Please spell "class name" as two words (like line 92).

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -428,6 +434,28 @@ public synchronized Set<Map.Entry<Object,Object>> entrySet() {
...
+                    // mark it to be removed from returned set.
+                    toRemove.add(entry);
+                }
+            }
+            if (!toRemove.isEmpty()) {
+                Set<Map.Entry<Object, Object>> modifiableSet = new HashSet<>(entrySet);
+                modifiableSet.removeAll(toRemove);
+                entrySet = Collections.unmodifiableSet(modifiableSet);
+            }
+        }
```
keithc-ca: I think this belongs in the else block of line 425.
Why does it not use `checkRestrictedSecurityKey()`?

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -945,6 +1097,51 @@ private void implClear() {
...
+     */
+    private static boolean isProviderInfoKey(Object key) {
+        if (key instanceof String sk) {
+            return providerInfoKeys.contains(sk);
+        }
+        return false;
+    }
+
+    /*
+     * Creates a Service instance based on the provided key.
```
keithc-ca: Please don't capitalize service here (compare with line 1129).

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -945,6 +1097,51 @@ private void implClear() {
...
+
+        return null;
+    }
+
+    /*
+     * Checks if the provided key is one of the following:
+     *  - One that cannot be used to create a service
+     *  - Pertaining to information about the provider
+     *  - Corresponding to a service that is allowed
+     *    by the active RestrictedSecurity profile
```
keithc-ca: The points should not begin with capitals; suggest
```java
     *  - one that cannot be used to create a service
     *  - pertains to information about the provider
     *  - corresponds to a service that is allowed
```

keithc-ca (changes requested):

closed/test/jdk/openj9/internal/security/TestLegacyProviders.java
KostasTsiounis: Changed.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Changed to lowercase.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Changed to suggestion.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: I can't add this to the else block in line 425, because it will only be executed once. We, however, want this to be executed every time someone calls this method, since the allowed services depend on who's making the call (`RestrictedSecurity` allows constraints that indicate a specific class that is allowed to use a service - the stack trace is checked).

I did see that I had a mistake, though, where I was changing the underlying `entrySet`, which I shouldn't. So, I updated that to return a copy of it with the disallowed services removed.

KostasTsiounis: I also changed it to call `checkRestrictedSecurityKey()`. I missed that spot before.

src/java.base/share/classes/java/security/Provider.java
keithc-ca: I didn't mean to suggest removal of line 1132. The final bullet could be
```java
     *  - corresponds to a service that is allowed
     *    by the active RestrictedSecurity profile
```
I'll let you decide if the qualifier is redundant in this context.

src/java.base/share/classes/java/security/Provider.java
keithc-ca: I had forgotten to consider call-specific behavior, however, this still modifies `entrySet` when it should not.

In restricted security mode, I don't think the field can be field can be set (ever). My new suggestion is that this method begin like this:
```java
        checkInitialized();

        if (RestrictedSecurity.isEnabled()) {
            return super.entrySet().stream()
                        .filter(entry -> checkRestrictedSecurityKey(entry.getKey()))
                        .collect(Collectors.toSet());
        }

        if (entrySet == null) {
```
and the remainder of the method would remain the same as upstream.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -452,6 +474,20 @@ public Set<Object> keySet() {
...
+            Set<Map.Entry<Object, Object>> entrySet = entrySet();
+            Collection<Object> valuesToReturn = new ArrayList<>(entrySet.size());
+            for (Map.Entry<Object, Object> entry : entrySet) {
+                if (checkRestrictedSecurityKey(entry.getKey())) {
+                    // We're in restricted security mode which allows this service
+                    // or provider info, so add it to list of values to be returned.
+                    valuesToReturn.add(entry.getValue());
+                }
+            }
+            return Collections.unmodifiableCollection(valuesToReturn);
```
keithc-ca: The set should have already been filtered by `entrySet()`; this only needs to collect the values:
```java
            return entrySet().stream().map(entry -> entry.getValue()).toList();
```

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -645,6 +681,16 @@ public synchronized Object merge(Object key, Object value,
...
+
+        if (RestrictedSecurity.isEnabled() && !isProviderInfoKey(key)) {
+            // This is not a value pertaining to provider info.
+            Service service = createServiceFromKey(key);
+            if ((service != null) && !RestrictedSecurity.isServiceAllowed(service)) {
+                // We're in restricted security mode which does not allow this service,
+                // return null.
+                return null;
+            }
+        }
```
keithc-ca: Can this not use `checkRestrictedSecurityKey()`?
```java
        if (RestrictedSecurity.isEnabled() && !checkRestrictedSecurityKey(key)) {
            // We're in restricted security mode which does not allow this service.
            return null;
        }
```
Likewise for `getOrDefault()`, but it should return the default value instead of null.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -677,12 +734,37 @@ public Enumeration<Object> keys() {
...
+            Set<Map.Entry<Object, Object>> entrySet = entrySet();
+            List<Object> list = new ArrayList<>(entrySet.size());
+            for (Map.Entry<Object, Object> entry : entrySet) {
+                if (checkRestrictedSecurityKey(entry.getKey())) {
+                    // We're in restricted security mode which allows this service
+                    // or provider info, so add it to list to be returned.
+                    list.add(entry.getValue());
+                }
+            }
+            return Collections.enumeration(list);
```
keithc-ca: This can reuse `values()`
```java
            return Collections.enumeration(values());
```

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -677,12 +734,37 @@ public Enumeration<Object> keys() {
...
+
+        if (RestrictedSecurity.isEnabled() && !isProviderInfoKey(key)) {
+            // This is not a value pertaining to provider info.
+            Service service = createServiceFromKey(key);
+            if ((service != null) && !RestrictedSecurity.isServiceAllowed(service)) {
+                // We're in restricted security mode which does not allow this service,
+                // return null.
+                return null;
+            }
+        }
```
keithc-ca: Reuse `checkRestrictedSecurityKey()` here as well?

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -850,6 +946,14 @@ private void implReplaceAll(BiFunction<? super Object, ? super Object,
                 if (!checkLegacy(sk)) {
                     continue;
                 }
+
+                Service service = createServiceFromKey(key);
+                if ((service != null) && !RestrictedSecurity.canServiceBeRegistered(service)) {
+                    // We're in restricted security mode which does not allow this service,
+                    // return without replacing anything.
+                    continue;
```
keithc-ca, line 910: "return" is not the right word here.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -808,6 +890,13 @@ private boolean implRemove(Object key, Object value) {
     private boolean implReplace(Object key, Object oldValue, Object newValue) {
         if (!checkLegacy(key)) return false;
 
+        Service service = createServiceFromKey(key);
+        if ((service != null) && !RestrictedSecurity.canServiceBeRegistered(service)) {
```
keithc-ca: I suggest a helper method for this recurring pattern.

keithc-ca (changes requested):

test/lib/jdk/test/lib/LegacyProvider.java
```diff
@@ -0,0 +1,50 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * ===========================================================================
+ */
+
+/**
+ * A provider created to test the restriction of legacy paths through
+ * the RestrictedSecurity mode.
+ */
```
keithc-ca: Javadoc must precede the element it describes: This belongs before line 33.

test/lib/jdk/test/lib/LegacyProvider.java
```diff
@@ -0,0 +1,50 @@
...
+package jdk.test.lib;
+
+import java.security.Provider;
+
+public class LegacyProvider extends Provider {
```
keithc-ca, line 33:
```
test/lib/jdk/test/lib/LegacyProvider.java:33: warning: [serial] serializable class LegacyProvider has no definition of serialVersionUID
public class LegacyProvider extends Provider {
       ^
```

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: I can do that. I'll just have to add another step to make the returned set unmodifiable, since this is what this method is supposed to return.

keithc-ca: `Stream.toList()` documents that the returned list is unmodifiable; I expected the same for `Collectors.toSet()`, but you're right it doesn't promise immutability. Luckily there's `Collectors.toUnmodifiableSet()`.

KostasTsiounis: Changed to suggestion.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Added the line back. I think it's important to mention that this is in a `RestrictedSecurity` setting.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Changed to suggestion mostly. I changed the last part to be `collect(Collectors.toUnmodifiableList())` instead of `toList()` since that's what the method is supposed to return.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Changed in both places.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Changed it to use `values()`.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Added it there.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Updated.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Created helper methods for `canServiceBeRegistered` and `isServiceAllowed`.

test/lib/jdk/test/lib/LegacyProvider.java
KostasTsiounis: Moved Javadoc.

test/lib/jdk/test/lib/LegacyProvider.java
KostasTsiounis: I didn't see any of the other helper classes have anything to imitate. How do you think I should handle this? Just assign `1L` since it won't be actually serialized?

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -44,6 +44,7 @@
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.stream.Collectors;
```
keithc-ca, line 47: Please move this after line 48 (which is out of place, but from upstream).

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -452,6 +466,11 @@ public Set<Object> keySet() {
     @Override
     public Collection<Object> values() {
         checkInitialized();
+
+        if (RestrictedSecurity.isEnabled()) {
+            return entrySet().stream().map(entry -> entry.getValue()).collect(Collectors.toUnmodifiableList());
```
keithc-ca, line 471: Please use the `toList()` convenience method (which promises an unmodifiable list):
```java
            return entrySet().stream().map(entry -> entry.getValue()).toList();
```

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -645,6 +664,11 @@ public synchronized Object merge(Object key, Object value,
     @Override
     public Object get(Object key) {
         checkInitialized();
+
+        if (RestrictedSecurity.isEnabled() && !checkRestrictedSecurityKey(key)) {
+            // We're in restricted security mode which does not allow this service.
+            return null;
+        }
         return super.get(key);
```
keithc-ca, line 672: Please add a blank line between lines 671 and 672, like other updated methods.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -916,6 +1000,12 @@ private Object implComputeIfPresent(Object key, BiFunction<? super Object,
     private Object implPut(Object key, Object value) {
         if (!checkLegacy(key)) return null;
 
+        if (!canRestrictedSecurityServiceBeRegistered(key)) {
```
keithc-ca, line 1003: Please explain the rules regarding which of `isRestrictedSecurityServiceAllowed()` and `canRestrictedSecurityServiceBeRegistered()` should be used. `implCompute()` uses the former, while `implPut()`, `implReplace()` and `implReplaceAll()`  use the latter. This doesn't appear to be consistent.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -945,6 +1041,61 @@ private void implClear() {
...
+                return new Service(this, type, algorithm);
+            }
+        }
+
+        return null;
+    }
+
+    private boolean canRestrictedSecurityServiceBeRegistered(Object key) {
+        Service service = createServiceFromKey(key);
+        return ((service == null) || RestrictedSecurity.canServiceBeRegistered(service));
```
keithc-ca, line 1073: Please remove the outermost parentheses, here and on line 1078.

test/lib/jdk/test/lib/LegacyProvider.java
```diff
@@ -0,0 +1,50 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * ===========================================================================
+ */
+
```
keithc-ca, line 24: Please omit this blank line.

test/lib/jdk/test/lib/LegacyProvider.java
keithc-ca: You can pick a UID at random; eclipse suggested this:
```java
    private static final long serialVersionUID = 5997431841873045817L;
```

test/lib/jdk/test/lib/LegacyProvider.java
```diff
@@ -0,0 +1,50 @@
...
+package jdk.test.lib;
+
+import java.security.Provider;
+
+/**
+ * A provider created to test the restriction of legacy paths through
+ * the RestrictedSecurity mode.
+ */
+public class LegacyProvider extends Provider {
+    public LegacyProvider() {
```
keithc-ca, line 34: This triggers "this-escape" warnings; making the class `final` addresses that.

keithc-ca (changes requested):

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: The idea is that when doing `put()` or `replace()` you already know the value to be inserted, so you are not actually retrieving any information and should thus be allowed to be used from any place of your code (i.e., `canRestrictedSecurityServiceBeRegistered()`). But with `compute()` and all of its variants, one can use it to just retrieve an existing value, which should only be allowed from the specific class defined in the constraint (i.e., `isRestrictedSecurityServiceAllowed()`).

What I just noticed though is that `merge()` does a thing very similar to `compute()`, so I might have to switch that to behave like it.

keithc-ca: I don't think the result is consistent yet; examples include:
* `keys()` and `keySet()` may yield a different elements than the keys in `entrySet()`
* `forEach()` exposes entries that would not be included in `entrySet()`
* `merge()` and `put()` allow storing entries that would not be permitted via `compute*()`

KostasTsiounis: > I don't think the result is consistent yet; examples include:
> 
> * `keys()` and `keySet()` may yield a different elements than the keys in `entrySet()`
> * `forEach()` exposes entries that would not be included in `entrySet()`
> * `merge()` and `put()` allow storing entries that would not be permitted via `compute*()`

Regarding each of the points:
* The whole idea is to restrict getting services that potentially should not be allowed in a specific setting. Getting the keys doesn't give you access to any such services, thus the decision to not restrict `keys()` and `keySet()`. I could restrict that too, I just think it might be defeating the purpose. I would also need to add a debug path for cases where we want to see what is available through the provider.
* I actually missed that. I will update `forEach()`.
* Like I mentioned before, I will update `merge()` to match `compute()`, but I think there should be a difference between them and `put()`, because we need to allow users to register from wherever but only retrieve from explicitly specified classes. I know that I'm hindering the functionality of `compute()` and `merge()` that could be used to just update/register new values, unless I manually check whether the supplied BiFunction is a tautology (i.e., just returns the same value) and it's just used to get a value. Do you think that would make sense to do?

keithc-ca: > The whole idea is to restrict getting services that potentially should not be allowed in a specific setting

Perhaps I'm missing something. How does restricting access to keys or values affect access to services offered by a given provider?

KostasTsiounis: > Perhaps I'm missing something. How does restricting access to keys or values affect access to services offered by a given provider?

This whole change was inspired by a use case of someone using a legacy Kafka provider, where services were inserted using `put()` and then retrieved using `get()` and subsequently being initialized through reflection. In these cases `RestrictedSecurity` wasn't able to enforce constraints from the loaded profile. Of course, we couldn't just check these two methods.

As far as `keys()` goes, my initial thought process was that it could not possibly affect the aforementioned scenario. I will, however, restrict it too for consistency reasons.

keithc-ca: If one has access to a service to be able to insert it via `put()` it's too late to prevent distribution of that service to others.

It's not clear that it's possible to completely restrict access to services by participants that employ reflection. The most direct route might be to reflect `Properties.map` and operate directly on that object, bypassing anything we might do the implementation of `Provider`.

How much effort should we expend on this when it can always be circumvented?

keithc-ca:
How does this, and `RestrictedSecurity` in general, interact with https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/eec176970e1a740358b11dab22a7be56e0c0e4f4?

PR #1205, opened 2026-04-15
Generate JFR headers during build

thallium:
As first part of JFR event metadata loader support, this invokes the JFR
header generation from hotspot. Then post-processing commands will
extract information that can be used by OpenJ9.

Related: https://github.com/eclipse-openj9/openj9/issues/23688

Co-written by IBM Bob


closed/custom/Main-pre.gmk
```diff
@@ -46,7 +46,15 @@ ifeq (,$(filter clean dist-clean, $(SEQUENTIAL_TARGETS)))
   create-main-targets-include java.base-gensrc : generate-j9jcl-sources
 endif
 
-j9vm-build : buildtools-langtools
+# JFR header generation target - invokes wrapper that includes upstream GensrcJfr.gmk
```
keithc-ca: Please finish each sentence with a period.

closed/custom/Main-pre.gmk
```diff
@@ -46,7 +46,15 @@ ifeq (,$(filter clean dist-clean, $(SEQUENTIAL_TARGETS)))
...
+.PHONY : hotspot-jfr-gensrc
```
keithc-ca: Please instead add to the dependency after line 26.

closed/custom/Main-pre.gmk
```diff
@@ -46,7 +46,15 @@ ifeq (,$(filter clean dist-clean, $(SEQUENTIAL_TARGETS)))
...
+
+hotspot-jfr-gensrc : buildtools-hotspot
+  ifeq (true,$(OPENJ9_ENABLE_JFR))
+	@+($(MAKE) $(MAKE_ARGS) -f $(TOPDIR)/closed/make/hotspot/gensrc/GensrcJfrWrapper.gmk)
+  endif
```
keithc-ca: Please follow the existing conventions,
There's no need for a subshell: Please remove the outermost parentheses.
Please add a comment on `endif` (see line 60).
```
  ifeq ($(OPENJ9_ENABLE_JFR), true)
	@+$(MAKE) $(MAKE_ARGS) -f $(TOPDIR)/closed/make/hotspot/gensrc/GensrcJfrWrapper.gmk
  endif # OPENJ9_ENABLE_JFR
```

closed/custom/Main-pre.gmk
```diff
@@ -46,7 +46,15 @@ ifeq (,$(filter clean dist-clean, $(SEQUENTIAL_TARGETS)))
...
+
+j9vm-build : buildtools-langtools hotspot-jfr-gensrc
```
keithc-ca: Please omit the redundant dependency:
```
j9vm-build : hotspot-jfr-gensrc
```

closed/make/hotspot/gensrc/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,44 @@
...
+# version 2 for more details (a copy is included in the LICENSE file that
+# accompanied this code).
+#
+# You should have received a copy of the GNU General Public License version
+# 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+# ===========================================================================
+
+################################################################################
+# Minimal wrapper to invoke upstream GensrcJfr.gmk with proper dependencies
+# This wrapper ensures all required makefiles are included before GensrcJfr.gmk
```
keithc-ca: Please finish each sentence with a period.

closed/make/hotspot/gensrc/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,44 @@
...
+################################################################################
+
+default: all
```
keithc-ca: Just `all` will suffice:
```
all:
```
Please remove `default` from the `.PHONY` rule.

closed/make/hotspot/gensrc/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,44 @@
...
+
+include $(SPEC)
+include MakeBase.gmk
+include Execute.gmk
+
+# Set JVM_VARIANT_OUTPUTDIR to OpenJ9's build directory
+JVM_VARIANT_OUTPUTDIR := $(OPENJ9_VM_BUILD_DIR)
+
+# Include ToolsHotspot.gmk to get TOOL_JFR_GEN and BUILD_TOOLS_HOTSPOT
+include $(TOPDIR)/make/ToolsHotspot.gmk
```
keithc-ca: This is redundant (see `GensrcJfr.gmk`); please remove.

closed/make/hotspot/gensrc/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,44 @@
...
+
+# Now include the upstream GensrcJfr.gmk which will use our JVM_VARIANT_OUTPUTDIR
+include $(TOPDIR)/make/hotspot/gensrc/GensrcJfr.gmk
+
+# The SetupExecute macro creates jfr_gen_headers variable with the target
+all: $(jfr_gen_headers_TARGET)
```
keithc-ca: This should make fewer assumptions, use `TARGETS` instead:
```
# GensrcJfr.gmk appends to TARGETS.
all: $(TARGETS)
```

keithc-ca (changes requested):
At least some of the generated header files depend on Hotspot header files which we won't have: What's the plan to deal with that?

thallium:
> At least some of the generated header files depend on Hotspot header files which we won't have: What's the plan to deal with that?

@tajila Will the headers be included or we're just parsing information from the headers?

tajila:
> @tajila Will the headers be included or we're just parsing information from the headers?

Ideally, we would just include the headers. But if thats not possible as they are, perhaps we can do some post-processing to allow us to include the headers, or parse the info we need from them.

closed/custom/Main-pre.gmk
```diff
@@ -26,6 +26,7 @@ CLEAN_DIRS += vm
 	debug-image \
 	j9vm-build \
 	test-image-openj9 \
+	hotspot-jfr-gensrc \
```
keithc-ca: I did say "after line 26" so the names remain sorted.

closed/make/hotspot/gensrc/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,41 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# version 2 for more details (a copy is included in the LICENSE file that
+# accompanied this code).
+#
+# You should have received a copy of the GNU General Public License version
+# 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+# ===========================================================================
+
+################################################################################
+# Minimal wrapper to invoke upstream GensrcJfr.gmk with proper dependencies
```
keithc-ca: Missing period.

keithc-ca (changes requested):

closed/GensrcJfrWrapper.gmk
keithc-ca, line 1: I don't think we need two new folders for this file, suggest it should be in `$(TOPDIR)/closed` (preferred) or `$(TOPDIR)/closed/make`.

keithc-ca:
In the commit message, I think the co-written line should immediately precede the signed-off-by line and relocate the hyphen:
```
Co-written-by: IBM Bob
Signed-off-by: Gengchen Tuo <gengchen.tuo@ibm.com>
```

thallium:
Updating generation code so generated header can be used by VM.

make/src/classes/build/tools/jfr/GenerateJfrFiles.java
```diff
@@ -702,7 +707,62 @@ private static void printJfrEventIdsHpp(Metadata metadata, File outputFile) thro
...
+                String javaName = "jdk." + t.name;
+                out.write("  {\"" + javaName.replace('.', '/') + "\", " + t.id + "},");
+            }
+            out.write("};");
+            out.write("");
+            out.write("static const int JFR_EVENT_MAPPING_COUNT = sizeof(jfrEventMappings) / sizeof(JfrEventMapping);");
+            out.write("");
+            out.write("// Inline function to populate J9HashTable with event IDs");
+            out.write("// Returns true on success, false on failure");
+            out.write("inline bool addEventIds(J9HashTable *table, J9JavaVM *vm) {");
```
tajila: I would refrain from implementing this in generated code, because if we want to make any changes we need to change the generator. Instead, we can just call a function that exists elsewhere.

tajila: Basically, you need to generate a function that can be called from the VM that looks something like:

```
inline bool addEventIds(J9JavaVM *vm) {
      addEvent(vm, "eventName1", eventID1, true);
      addEvent(vm, "eventName2", eventID2, true);
      addEvent(vm, "eventName3", eventID3, true);
      ...
      addEvent(vm, "eventNameN", eventIDN, true);
}
```

where `addEvent(J9JavaVM *vm, const char * eventOrTypeName, UDATA id, bool isEvent)` is implemented in J9 code.

Keeping as much of the VM internals out of generated code as possible.

The same can be below for JFRTypes, we only need a bool to track if its event or type. Currently, we dont make a distinction between the two in the VM, but to future proof we should track that info.

tajila: It might be easier if the function is defined as a C call

closed/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,41 @@
...
+################################################################################
+
+all:
+
+include $(SPEC)
+include MakeBase.gmk
+include Execute.gmk
+
+# Set JVM_VARIANT_OUTPUTDIR to OpenJ9's build directory.
+JVM_VARIANT_OUTPUTDIR := $(OPENJ9_VM_BUILD_DIR)
```
tajila: i think this needs to be $(OPENJ9_VM_BUILD_DIR)/include

make/src/classes/build/tools/jfr/GenerateJfrFiles.java
```diff
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2026, 2025 All Rights Reserved
```
keithc-ca: Both years should be 2026.

I'm not thrilled about modifying this code: Can we not derive what we need from something this already generates?

keithc-ca (changes requested):

make/src/classes/build/tools/jfr/GenerateJfrFiles.java
thallium: If we don't modify this file, we might have to hard-code 200+ lines of `addEvent(vm, "eventNameN", eventIDN, true);`, which can be hard to keep synchronized with the upstream metadata.

keithc-ca: It seems we really only need the list of event names and ids. The rest can be manually written.

make/src/classes/build/tools/jfr/GenerateJfrFiles.java
keithc-ca: I don't think we even need a function, just the enum (maybe) and the body of a table:
```c
  { "jdk/Duration", 2 },
  { "jdk/Instant", 3 },
  { "jdk/Value", 4 },
  { "jdk/Text", 5 },
  { "jdk/ZThreadDebug", 6 }, 
  { "jdk/ResidentSetSize", 7 },
  { "jdk/ThreadStart", 8 },
...
  { "jdk/JavaAgent", 161 },
  { "jdk/NativeAgent", 162 },
  { "jdk/DeprecatedInvocation", 163 },
```
I expect we can achieve that with `$(SED) < $(JVM_VARIANT_OUTPUTDIR)/gensrc/jfrfiles/jfrEventIds.hpp`.

make/src/classes/build/tools/jfr/GenerateJfrFiles.java
thallium: Yes, but we can't easily get java names of events and types from the original generated headers.

make/src/classes/build/tools/jfr/GenerateJfrFiles.java
keithc-ca: More concretely:
```
$(SED) -E -n -e 's|^ *Jfr([^ ]*) = ([0-9]*),|{ "jdk/\1", \2 },|p'
```

make/src/classes/build/tools/jfr/GenerateJfrFiles.java
keithc-ca: Please share one or more examples where a suggestion like in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1205#discussion_r3111854678 doesn't work.

thallium: I think that would work for events. However, for types, not everything is under `jdk/types` (although still can be solved by hard-coding the exceptions). What are the concerns of modifying this file?

keithc-ca: Every new file we modify increases the cost of merging changes from upstream. It doesn't appear necessary in this case.

I didn't see anything in your change for anything other than events.

thallium: > Every new file we modify increases the cost of merging changes from upstream.

Is creating a new file a better option then?

> I didn't see anything in your change for anything other than events.

Please refer to line 793-796.

keithc-ca: The concern is only with source files tracked in the repository. New (generated) files are not a concern.

Extracting the type identifiers from `enum JfrTypeId` should be similarly straight-forward. The special names dealt with by `name_to_id()` should be simple to handle directly.

thallium:
Added a post-processing script to extract JFR event and type mappings.

make/scripts/extract-jfr-mappings.sh
```diff
@@ -0,0 +1,153 @@
...
+
+EVENT_IDS_FILE="$1"
+TYPE_IDS_FILE="$2"
+
+# Output file in same directory as source headers
+OUTPUT_DIR=$(dirname "$EVENT_IDS_FILE")
+OUTPUT_FILE="$OUTPUT_DIR/jfrMappings.hpp"
+
+# Extract event enum
+EVENT_ENUM=$(sed -n '/enum JfrEventId {/,/^};/p' "$EVENT_IDS_FILE")
```
keithc-ca: This should not be in a shell script, where it does not have access to `AWK`, `GREP`, `SED`, etc.

We don't need anything this complicated anyway. The essential tables can be produced by simple commands in makefiles (which do have access to `SED`, etc.). Those tables can be included in source files that are included in the openj9 repository.

thallium: Just realized it might be impossible to produce the type mapping using post-processing commands because the type names are capitalized (e.g. `TYPE_VIRTUALSPACE`) so we can't reconstruct the Java type name `jdk/types/VirtualSpace`.

keithc-ca: Where would the name `"jdk/types/VirtualSpace"` be needed?

keithc-ca: The generated file `jfrTypes.hpp` contains
```c
  static jlong name_to_id(const char* type_name) {
    if (strcmp(type_name, "java.lang.Class") == 0) {
      return TYPE_CLASS;
    }
    if (strcmp(type_name, "java.lang.Thread") == 0) {
      return TYPE_THREAD;
    }
    if (strcmp(type_name, "long") == 0) {
      return TYPE_LONG;
    }
    if (strcmp(type_name, "int") == 0) {
      return TYPE_INT;
    }
    if (strcmp(type_name, "short") == 0) {
      return TYPE_SHORT;
    }
    if (strcmp(type_name, "byte") == 0) {
      return TYPE_BYTE;
    }
    if (strcmp(type_name, "double") == 0) {
      return TYPE_DOUBLE;
    }
    if (strcmp(type_name, "float") == 0) {
      return TYPE_FLOAT;
    }
    if (strcmp(type_name, "boolean") == 0) {
      return TYPE_BOOLEAN;
    }
    if (strcmp(type_name, "char") == 0) {
      return TYPE_CHAR;
    }
    if (strcmp(type_name, "java.lang.String") == 0) {
      return TYPE_STRING;
    }
    if (strcmp(type_name, "jdk.types.StackTrace") == 0) {
      return TYPE_STACKTRACE;
    }
    return -1;
  }
```
You'll notice it only handles a few, well known, type names. There's no mention of `VirtualSpace`.

thallium: > Where would the name "jdk/types/VirtualSpace" be needed?

The Java type name is used as key in https://github.com/eclipse-openj9/openj9/blob/master/runtime/vm/jfr.cpp#L1462-L1463. The issue is not about  `name_to_id`, my point was we wouldn't be able to reconstruct the type name `jdk/types/VirtualSpace` with correct cases as `TYPE_VIRTUALSPACE` is in upper case.

keithc-ca: So the plan is to replace the hashtable there with a (generated) hard-coded map from type name to type id to be used in the implementation of `getTypeIdImpl()`? If so, I suggest we want a new generator specific to openj9 that produces the header files we want (like `GenerateJfrFiles --mode headers` does for Hotspot).

tajila: The table is not going away, we need it for event IDs (getTypeIdImpl()) but we also need it for user events as well as storing metadata associated with the events. Secondly, the JCL uses the generator as well so if we create a new one it has to work exactly the same (generate the same IDs).

`name_to_id` is an optimization, if this is an issue we can just do without.

tajila: The most important thing we need from the headers is, the name of the class, the id, if its a type or an event.

keithc-ca: The Hotspot generator does not produce anything that I can see that connects `TYPE_VIRTUALSPACE` and `jdk/types/VirtualSpace`, for example, thus we need a new generator.

tajila: Does `jdk/types/VirtualSpace` actually exists? Its used in the GC heap summary event but I cant find anywhere in the JCL where this is referenced.

I suspect this may be one of the types that are only used in the RI VM.

thallium: [This](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/f93fa420824c51c07bffafc4dbf0e8faad8a7865/make/src/classes/build/tools/jfr/GenerateJfrFiles.java#L191) is what we need. If we create a new generator, most of the code will come from the old generator. Not sure about the maintainability of this approach vs modifying the old generator.

keithc-ca: Ok, we'd have to copy more code than I would like. Instead of modifying the behavior of the existing modes of that class, I suggest we introduce a new mode (e.g. `--mode openj9` or similar) that would use new methods to produce the content we want without modifying any existing code (besides recognizing and acting on the new `OutputMode`). Keeping our additions well isolated should minimize future conflicts.

thallium:
After discussion with Tobi, we decided that only type mappings in `name_to_id()` are needed, so postprocessing could still work. I've moved the post-processing commands to the makefile.

closed/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,95 @@
...
+	$(ECHO) '// Standalone header with no external dependencies.' >> $@
+	$(ECHO) '' >> $@
+	$(ECHO) '#ifndef JFR_EXTRACTED_MAPPINGS_HPP' >> $@
+	$(ECHO) '#define JFR_EXTRACTED_MAPPINGS_HPP' >> $@
+	$(ECHO) '' >> $@
+	$(ECHO) '// ============================================================================' >> $@
+	$(ECHO) '// Event Enum' >> $@
+	$(ECHO) '// ============================================================================' >> $@
+	$(ECHO) '' >> $@
+	$(SED) -n '/enum JfrEventId {/,/^};/p' $(JFR_OUTPUTDIR)/jfrEventIds.hpp >> $@
```
keithc-ca: Please keep only the two `SED` commands capturing the event ids and type names in separate files. The rest can be written manually in openj9.

closed/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,95 @@
...
+include Execute.gmk
+
+# Set JVM_VARIANT_OUTPUTDIR to OpenJ9's build directory.
+JVM_VARIANT_OUTPUTDIR := $(OPENJ9_VM_BUILD_DIR)/include
+
+# Now include the upstream GensrcJfr.gmk which will use our JVM_VARIANT_OUTPUTDIR.
+include $(TOPDIR)/make/hotspot/gensrc/GensrcJfr.gmk
+
+# Extract JFR mappings after headers are generated
+extract_jfr_mappings: $(jfr_gen_headers_TARGET)
```
keithc-ca: `extract_jfr_mappings` should be listed as a `.PHONY` target.

closed/GensrcJfrWrapper.gmk
thallium: Mind elaborating? I think all the `SED` commands are necessary.

keithc-ca: Based on https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1205#issuecomment-4305893988, it appears that only the type ids are required, so I'll focus on that.

I'm suggesting that the new wrapper makefile contain (among other things) this sequence:
```
EVENT_IDS   := $(JVM_VARIANT_OUTPUTDIR)/gensrc/jfrfiles/jfrEventIds.hpp
EVENT_ENUM  := $(JVM_VARIANT_OUTPUTDIR)/gensrc/jfrfiles/jfrEventEnum.hpp

$(EVENT_ENUM): $(EVENT_IDS)
	$(SED) -n '/enum JfrEventId {/,/^};/p' < $< > $@
```
This yields `jfrEventEnum.hpp` which looks like this (the middle is elided):
```c
enum JfrEventId {
  JfrMetadataEvent = 0,
  JfrCheckpointEvent = 1, 
  JfrDurationEvent = 2,
  JfrInstantEvent = 3,
// etc.
  JfrFinalizerStatisticsEvent = 160, 
  JfrJavaAgentEvent = 161,
  JfrNativeAgentEvent = 162,
  JfrDeprecatedInvocationEvent = 163,
};
```
Then, in openj9, we can include `jfrEventEnum.hpp` in `JfrEvents.h` like this
```c
#ifndef JfrEvents_h
#define JfrEvents_h

#include "jfrEventEnum.hpp"

typedef enum JfrEventId JfrEventId;

#endif /* JfrEvents_h */
```
If need be, other declarations can be added to `JfrEvents.h`.

thallium: I get the general idea. What I meant is that we still need the type mappings and event mappings:
```cpp
static const TypeMapping TYPE_MAPPINGS[] = {
  { "java/lang/Class", TYPE_CLASS },
  { "java/lang/Thread", TYPE_THREAD },
  { "long", TYPE_LONG },
  { "int", TYPE_INT },
  { "short", TYPE_SHORT },
  { "byte", TYPE_BYTE },
  { "double", TYPE_DOUBLE },
  { "float", TYPE_FLOAT },
  { "boolean", TYPE_BOOLEAN },
  { "char", TYPE_CHAR },
  { "java/lang/String", TYPE_STRING },
  { "jdk/types/StackTrace", TYPE_STACKTRACE },
};

static const EventMapping EVENT_MAPPINGS[] = {
  { "jdk/MetadataEvent", 0 },
  { "jdk/CheckpointEvent", 1 },
  //...
};
```

Unless you have other suggestion on how this should be generated.

keithc-ca: `TYPE_MAPPINGS` can be written exactly as quoted above in the appropriate source file in openj9. That table doesn't appear likely to change frequently, if at all.

The body of `EVENT_MAPPINGS` can be written to a separate file, say `jfrEventTable.hpp`, via the `SED` command you have:
```
    $(SED) -E -n 's|^ *Jfr([^ ]*) = ([0-9]*),|  { "jdk/\1", \2 },|p'
```
The openj9 source files would declare `EventMapping` and say:
```c
static const EventMapping EVENT_MAPPINGS[] = {
#include "jfrEventTable.hpp"
};
```

closed/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,58 @@
...
+JVM_VARIANT_OUTPUTDIR := $(OPENJ9_VM_BUILD_DIR)/include
+
+# Now include the upstream GensrcJfr.gmk which will use our JVM_VARIANT_OUTPUTDIR.
+include $(TOPDIR)/make/hotspot/gensrc/GensrcJfr.gmk
+
+EVENT_IDS      := $(JFR_OUTPUTDIR)/jfrEventIds.hpp
+EVENT_ENUM     := $(JFR_OUTPUTDIR)/jfrEventEnum.hpp
+EVENT_MAPPINGS := $(JFR_OUTPUTDIR)/jfrEventMapping.hpp
+TYPE_IDS       := $(JFR_OUTPUTDIR)/jfrTypes.hpp
+TYPE_ENUM      := $(JFR_OUTPUTDIR)/jfrTypeEnum.hpp
```
keithc-ca: Suggest these be grouped as inputs and outputs of this makefile:
```
# Files generated by GensrcJfr.gmk:
EVENT_IDS      := $(JFR_OUTPUTDIR)/jfrEventIds.hpp
TYPE_IDS       := $(JFR_OUTPUTDIR)/jfrTypes.hpp

# Additional files for OpenJ9:
EVENT_ENUM     := $(JVM_VARIANT_OUTPUTDIR)/jfrEventEnum.hpp
EVENT_MAPPINGS := $(JVM_VARIANT_OUTPUTDIR)/jfrEventMapping.hpp
TYPE_ENUM      := $(JVM_VARIANT_OUTPUTDIR)/jfrTypeEnum.hpp
```
Notice the outputs go directly in `$(OPENJ9_VM_BUILD_DIR)/include` instead of two directory levels deeper so code can say
```c
#include "jfrEventEnum.hpp"
```
instead of
```c
#include "gensrc/jfrfiles/jfrEventEnum.hpp"
```

closed/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,58 @@
...
+
+$(EVENT_ENUM): $(jfr_gen_headers_TARGET)
```
keithc-ca: Please use `TARGETS` instead of `jfr_gen_headers_TARGET`.

closed/GensrcJfrWrapper.gmk
```diff
@@ -0,0 +1,58 @@
...
+$(EVENT_MAPPINGS): $(jfr_gen_headers_TARGET)
+	$(SED) -E -n 's|^ *Jfr([^ ]*) = ([0-9]*),|  { "jdk/\1", \2 },|p' $(EVENT_IDS) > $@
+
+$(TYPE_ENUM): $(jfr_gen_headers_TARGET)
+	$(SED) -n '/enum JfrTypeId {/,/^};/p' $(TYPE_IDS) > $@
+
+TARGETS += $(EVENT_ENUM) $(EVENT_MAPPINGS) $(TYPE_ENUM)
+
+# GensrcJfr.gmk appends to TARGETS.
+all: $(TARGETS)
```
keithc-ca: Line 53 is not needed; line 56 can be
```
all: $(EVENT_ENUM) $(EVENT_MAPPINGS) $(TYPE_ENUM)
```
Line 55, if we keep, it belongs either after line 36, or before line 44.

keithc-ca (changes requested):

keithc-ca:
The commit message should be updated so it does not refer to a non-existent "post-processing script"; and "IBM Bob" should not be hyphenated.

thallium:
Sorry I just realized I was updating the wrong repo. Just updated this PR.

keithc-ca:
Please update the commit message; see https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1205#issuecomment-4254664777:
```
Co-written-by: IBM Bob
Signed-off-by: Gengchen Tuo <gengchen.tuo@ibm.com>
```

tajila (approved):

keithc-ca:
The commit message has not been updated as requested; see https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1205#issuecomment-4353116450.

> Co-written by

should be

> Co-written-by:

keithc-ca:

thallium:
Backports:
- JDK26: https://github.com/ibmruntimes/openj9-openjdk-jdk26/pull/57
- JDK25: https://github.com/ibmruntimes/openj9-openjdk-jdk25/pull/131
- JDK21: https://github.com/ibmruntimes/openj9-openjdk-jdk21/pull/435
- JDK17: https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/610

JDK11 is different, will modify the changes and open a PR later.

keithc-ca:
JFR is enabled by default for Linux x64 in jdk08; we'll want a change for that version as well.

PR #1211, opened 2026-05-07
Remove duplicated entries from the FIPS 140-2 ProblemList files

taoliult:
According to the issue https://github.com/eclipse-openj9/openj9/issues/23838. Several test exclusions were listed more than once for the same test path. This change keeps a single exclusion entry for each test and removes the duplicates to make the lists easier to maintain.


keithc-ca (approved):

taoliult:
I did not see any duplicate entries in the JDK8 FIPS 140-2 problem list, so no update is needed for JDK8.

PR #1217, opened 2026-05-26
Update KeytoolOpensslInteropTest to use FIPS Compliant pkcs12 file

dev-koan:
When in FIPS mode KeytoolOpensslInteropTest is updated to use a FIPS compliant pkcs12 file instead of skipping the test.

Signed-off-by: Dev Agarwal [dev.agarwal@ibm.com](mailto:dev.agarwal@ibm.com)

---------


dev-koan:
The copyright check is failing because it is expecting a copyright tag in the new PKCS12 file but I don't think we can add it there

keithc-ca:
What are the key properties of `os6FIPSCompliant` that makes it FIPS-compliant and `os6` not?

dev-koan:
> What are the key properties of `os6FIPSCompliant` that makes it FIPS-compliant and `os6` not?

There are two: 

- The salt length, os6 has a salt length of 8 bytes while os6FIPSCompliant has a salt length of 16 bytes
- The password length, os6 password ("changeit") is less 10 characters while os6FIPSCompliant password ("changeitlongversion") is 19 characters

Another note:- Since PKCS12 key store is not supported in the strict profile the test will remain excluded there. But for strongly-enforced the test will actually run now (as before we were just skipping it).

keithc-ca (approved):

PR #1220, opened 2026-05-29
Add test property `vm.cds.nocoh.archive.available=false`

keithc-ca:
For consistency with upstream change:
* [8385437: Provide test property vm.cds.nocoh.archive.available and use it in CompressedCPUSpecificClassSpaceReservation](https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/6160ce254a09ddb1bcd2653af8d2aafd54db5f8b)

---------

### DIFF
--- a/closed/test/jtreg-ext/requires/OpenJ9PropsExt.java
+++ b/closed/test/jtreg-ext/requires/OpenJ9PropsExt.java
@@ -72,6 +72,7 @@ public class OpenJ9PropsExt implements Callable<Map<String, String>> {
         map.put("vm.asan", "false");
         map.put("vm.bits", this::vmBits);
         map.put("vm.cds", "false");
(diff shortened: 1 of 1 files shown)


keithc-ca:
I think the copyright check failure is an error. The file path matches `*/test/*` so `copyrightCheckDir.sh` would not expect the classpath exception; it seems `copyrightCheck` should be updated to match that behavior.

pshipton (approved):

pshipton:
Not sure we should correct the copyright check in this case. Since the file is under closed, it expects an IBM license instead of an Oracle license, but I assume we've copied the file from OpenJDK and modified it, keeping the Oracle license. We could change the file location.

keithc-ca:
> Not sure we should correct the copyright check in this case

The two scripts should be consistent, but as a committer you have the discretion to ignore the complaint.

PR #1223, opened 2026-06-04
Update expectedValueTest by dynamically fetching locale display name

Amrutha-Kanhirathingal:
Updated the test scripts to fetch the locale display name dynamically instead of expecting locale name in English.

The existing test script is expecting locale display name in English in all of the mbcs machines but in Chinese & Japanese machines the toString() method uses Locale.getDisplayName() which returns names in the JVM's default locale, it may not be English always.

The fix we will fetch the default locale display name of each machines across all machine and pass to assertTrue for comparison.

 Related : https://github.ibm.com/runtimes/automation/issues/881

---------

### DIFF
--- a/test/jdk/java/text/Format/CompactNumberFormat/ToStringTest.java
+++ b/test/jdk/java/text/Format/CompactNumberFormat/ToStringTest.java
@@ -22,6 +22,12 @@
  */
 
 /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2026, 2026 All Rights Reserved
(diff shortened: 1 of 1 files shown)


pshipton:
The IBM copyright is missing.

When the display name is not "English (Canada)", what is it? Is there an issue being fixed by this?

Amrutha-Kanhirathingal:
Added IBM Policy ,Updated the related ticket information & Description

pshipton (approved):

keithc-ca (approved):

PR #1227, opened 2026-06-18
Set jdk.tls.namedGroups property through RestrictedSecurity profiles

KostasTsiounis:
The `jdk.tls.namedGroups` system property can now be set through a RestrictedSecurity profile, instead of being specified explicitly through the command line.

The `TestProperties.java` test was updated to verify this new feature.

---------

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -632,6 +632,17 @@ public final class RestrictedSecurity {
             // SSL property "javax.net.ssl.keyStore" set at the JVM level via system properties.
             System.setProperty("javax.net.ssl.keyStore", keyStore);
         }
+        String jdkTlsNamedGroups = restricts.jdkTlsNamedGroups;
+        if (!isNullOrBlank(jdkTlsNamedGroups)) {
+            String namedGroups = System.getProperty("jdk.tls.namedGroups");
+            if (namedGroups == null) {
+                // TLS property "jdk.tls.namedGroups" set at the JVM level via system properties.
+                System.setProperty("jdk.tls.namedGroups", jdkTlsNamedGroups);
+            } else {
+                printStackTraceAndExit("System property jdk.tls.namedGroups cannot be set"
+                        + " when defined in a RestrictedSecurity profile");
+            }
+        }
     }
 
     /**
@@ -777,6 +788,7 @@ public final class RestrictedSecurity {
         private final String jdkTlsDisabledAlgorithms;
         private final String jdkTlsEphemeralDHKeySize;
         private final String jdkTlsLegacyAlgorithms;
+        private final String jdkTlsNamedGroups;
         private final String jdkCertpathDisabledAlgorithms;
         private final String jdkSecurityLegacyAlgorithms;
         private final String keyStoreType;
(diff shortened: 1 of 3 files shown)


keithc-ca:
What guarantees the system property is set early enough to be available elsewhere (e.g. in `sun.security.ssl.NamedGroup`)?

KostasTsiounis:
> What guarantees the system property is set early enough to be available elsewhere (e.g. in `sun.security.ssl.NamedGroup`)?

Nothing guarantees it. It's the same issue as we've had with other properties, like `ibm.com.fips.mode`. But in this case, given that the class `NamedGroup` is used for TLS connections, something else is needed before it's referenced, like `SSLContext.getInstance()` or `Security.getProperty()`, at least in the paths I've considered.

Do you have any other path in mind that might need `NamedGroup` before anything `Security` related is needed (thus loading the `Security` class and in turn the `RestrictedSecurity` class and associated profile)?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -632,6 +632,16 @@ private static void setProperties(Properties props) {
...
             System.setProperty("javax.net.ssl.keyStore", keyStore);
         }
+        String jdkTlsNamedGroups = restricts.jdkTlsNamedGroups;
+        if (!isNullOrBlank(jdkTlsNamedGroups)) {
+            String namedGroups = System.getProperty("jdk.tls.namedGroups");
+            if (namedGroups == null) {
+                // TLS property "jdk.tls.namedGroups" set at the JVM level via system properties.
+                System.setProperty("jdk.tls.namedGroups", jdkTlsNamedGroups);
+            } else {
+                printStackTraceAndExit("Property jdk.tls.namedGroups cannot be set when defined in a RestrictedSecurity profile");
```
keithc-ca: Suggest this message should be more specific:
```java
                printStackTraceAndExit("System property jdk.tls.namedGroups cannot be set"
                        + " when defined in a RestrictedSecurity profile");
```

closed/test/jdk/openj9/internal/security/TestProperties.java
```diff
@@ -338,6 +366,20 @@ private static void testStrongAlgorithms() {
         }
     }
 
+    private static void testSystemProperties() {
+        // Test javax.net.ssl.keyStore system property
```
keithc-ca: Please finish each sentence with a period (here and line 376).

keithc-ca (changes requested):
I did convince myself that the static initializer of `NamedGroup.SupportedGroups` (the only use of `jdk.tls.namedGroups` I could find) does trigger loading `RestricedSecurity` so the system property will be copied from the security profile when appropriate.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/test/jdk/openj9/internal/security/TestProperties.java
KostasTsiounis: Added in both places.

keithc-ca:
Please squash, ensuring that the commit message uses proper punctuation.

KostasTsiounis:
> Please squash, ensuring that the commit message uses proper punctuation.

Done.

keithc-ca:

keithc-ca:
The sanity.openjdk test failures are expected; the sanity.functional failures are known - see https://github.com/eclipse-openj9/openj9/issues/20835.

keithc-ca (approved):

PR #1232, opened 2026-06-30
Handle Windows SecureRandom strongAlgorithms defaults in test

Mohit-Rajbhar100698:
Update the test regular expression to match the Windows default
value of the `securerandom.strongAlgorithms` security property.

The default value on Windows differs from Linux and macOS, causing
the existing regular expression to fail even though the property is
correctly restored after all `RestrictedSecurity` settings are
removed.

Update the regular expression to match both the Windows and
non-Windows default values.

Signed-off-by: Mohit Rajbhar <mohit.rajbhar@ibm.com>

### DIFF
--- a/closed/test/jdk/openj9/internal/security/TestProperties.java
+++ b/closed/test/jdk/openj9/internal/security/TestProperties.java
@@ -223,7 +223,7 @@ public class TestProperties {
         // 4 - Test property - securerandom.strongAlgorithms remove algorithm in extended profile.
         tests.add(Arguments.of("Test-Profile-strongAlgorithms-Extension_2",
                 System.getProperty("test.src") + "/property-java.security",
-                "securerandom\\.strongAlgorithms: (?=.*NativePRNGBlocking:SUN)(?=.*DRBG:SUN)",
+                "securerandom\\.strongAlgorithms: (NativePRNGBlocking:SUN|Windows-PRNG:SunMSCAPI),DRBG:SUN",
                 0));
         // 5 - Test property - securerandom.strongAlgorithms invalid algorithm.
         tests.add(Arguments.of("Test-Profile-strongAlgorithms-InvalidFormat",


closed/test/jdk/openj9/internal/security/TestProperties.java
```diff
@@ -223,7 +223,7 @@ private static Stream<Arguments> patternMatches_strongAlgorithms() {
         // 4 - Test property - securerandom.strongAlgorithms remove algorithm in extended profile.
         tests.add(Arguments.of("Test-Profile-strongAlgorithms-Extension_2",
                 System.getProperty("test.src") + "/property-java.security",
-                "securerandom\\.strongAlgorithms: (?=.*NativePRNGBlocking:SUN)(?=.*DRBG:SUN)",
+                "securerandom\\.strongAlgorithms: (?=.*(?:NativePRNG(?:Blocking)?:SUN|Windows-PRNG:SunMSCAPI))(?=.*DRBG:SUN)",
```
keithc-ca: Please explain the pattern. It's not clear that it needs to be as complicated as it appears.
* `(?:X)` could instead be just `(X)` - I don't think we care if there are any matching groups
* why is lookahead necessary (`(?=X)`)?

Mohit-Rajbhar100698: > * `(?:X)` could instead be just `(X)` - I don't think we care if there are any matching groups

Yeah, you're right. I used (?:x) unnecessarily here. Since we don't use any matching groups in this test, I'll update it.

For the lookaheads, my intent was to verify that the property contains both the platform-specific algorithm (NativePRNG... or Windows-PRNG) and DRBG:SUN without relying on their order.  I will fix it remove lookahead

Mohit-Rajbhar100698: Updated.

keithc-ca (approved):

PR #1237, opened 2026-07-15
Ensure JFR periodic task exits promptly during JVM shutdown

tomal-majumder:
When running with -Xcheck:memory, the JFR periodic task thread not terminating at shutdown causes terminateRemainingThreads to return non-zero, which bypasses freeJavaVM and results in a large number of reported unfreed memory blocks.

The thread was blocked in takeNap() on JVM.CHUNK_ROTATION_MONITOR with no mechanism to wake promptly when shutdown began. Although PlatformRecorder had an existing inShutdown flag set by the JFR shutdown hook, the periodic task loop did not act on it in time.

Two changes are made:
- setInShutDown() notifies JVM.CHUNK_ROTATION_MONITOR to wake the periodic task immediately if suspended in takeNap().
- periodicTask() checks inShutdown after each work iteration and exits the loop before entering takeNap() if shutdown is requested.

Related Issue: https://github.com/eclipse-openj9/openj9/issues/23980

---------

### DIFF
--- a/src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2026, 2026 All Rights Reserved
+ * ===========================================================================
+ */
+
 package jdk.jfr.internal;
 
 import static jdk.jfr.internal.LogLevel.INFO;
@@ -163,6 +169,9 @@ public final class PlatformRecorder {
 
     static void setInShutDown() {
         inShutdown = true;
+        synchronized (JVM.CHUNK_ROTATION_MONITOR) {
+            JVM.CHUNK_ROTATION_MONITOR.notifyAll();
+        }
     }
 
     static boolean isInShutDown() {
@@ -541,6 +550,9 @@ public final class PlatformRecorder {
                 // Catch everything and log, but don't allow it to end the periodic task
                 Logger.log(JFR_SYSTEM, WARN, "Error in Periodic task: " + t.getMessage());
             } finally {
+                if (inShutdown) {
+                    break;
+                }
                 takeNap(wait);
             }
         }


tomal-majumder:
FYI @tajila

tajila:
jenkins test sanity alinux64 jdknext

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
```diff
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2021, 2026 All Rights Reserved
```
keithc-ca: The first year should also be 2026.

keithc-ca:
The thread that runs `periodicTask()` is a daemon thread: it should not block shutdown.

tomal-majumder:
> The thread that runs `periodicTask()` is a daemon thread: it should not block shutdown.

The following command:
```shell
./build/linux-x86_64-server-release/images/jdk/bin/java -XX:+EnableOpenJ9ExperimentalFlightRecording -Xcheck:memory -Dibm.java9.forceCommonCleanerShutdown=true -Xint -XX:StartFlightRecording  -verbose:shutdown  -version 
``` 
calls [ProtectedDestroyJavaVM](https://github.com/eclipse-openj9/openj9/blob/db67b2fc6e057f38e9dbe324ea6652b6864c25ec/runtime/vm/jniinv.c#L317), which calls `terminateRemainingThreads` at [jniinv.c:424](https://github.com/eclipse-openj9/openj9/blob/db67b2fc6e057f38e9dbe324ea6652b6864c25ec/runtime/vm/jniinv.c#L424.). The `terminateRemainingThreads` returns non-zero value, as it detects one thread still alive called "JFR Periodic Task" which skips the calling `freeJavaVm` at [jniinv.c:499](https://github.com/eclipse-openj9/openj9/blob/db67b2fc6e057f38e9dbe324ea6652b6864c25ec/runtime/vm/jniinv.c#L499).

The `freeJavaVM` method is doing a lot of memory free/cleanups. That is why skipping this reports lots of unfreed blocks during `Xcheck:memory`.

keithc-ca:
Deamon threads should not inhibit shutdown. See https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java#L465.

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
```diff
@@ -540,9 +549,11 @@ private void periodicTask() {
             } catch (Throwable t) {
                 // Catch everything and log, but don't allow it to end the periodic task
                 Logger.log(JFR_SYSTEM, WARN, "Error in Periodic task: " + t.getMessage());
-            } finally {
-                takeNap(wait);
             }
+            if (inShutdown) {
+                break;
+            }
+            takeNap(wait);
```
keithc-ca: Why should this not stay in the `finally` block?

tajila:
> Deamon threads should not inhibit shutdown. 

Daemon threads do not inhibit shutdown, but they do inhibit a clean shutdown where are resources are freed. This makes it impossible to use -Xcheck:memory to find unfreed blocks. We have addressed this in the past by explicitly shutting down threads, see https://github.com/eclipse-openj9/openj9/blob/db67b2fc6e057f38e9dbe324ea6652b6864c25ec/jcl/src/java.base/share/classes/java/lang/J9VMInternals.java#L108.

tajila:
@tomal-majumder Can you please take a look a the test failures and see if they are related to these changes.

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
tomal-majumder: Actually this can stay inside `finally`. I split it out for minor style preferences. I will move that inside `finally`.

tomal-majumder:
`sanity.functional` test: https://openj9-jenkins.osuosl.org/job/Test_openjdknext_j9_sanity.functional_aarch64_linux_Personal/63/
`sanity.openjdk` test: https://openj9-jenkins.osuosl.org/job/Test_openjdknext_j9_sanity.openjdk_aarch64_linux_Personal/21/

tomal-majumder:
> @tomal-majumder Can you please take a look a the test failures and see if they are related to these changes.

**`sanity.functional` failures** are known: https://github.com/eclipse-openj9/openj9/issues/20835.

**`sanity.openjdk` failures**: I am guessing they are environmental failures, unrelated to the change in `PlatformRecoreder.java`

- `java/lang/IO/IO.java` (all 4 testlists): fails due to `/usr/bin/expect` missing on the test agent.
- `ServiceLoader/LinkageErrorsTest.java`, `java/foreign/*`, `AsyncShutdownNowInvokeAny.java`: different subsystems, no overlap with JFR.
- `jdk_*vector*_j9_0 groups` — show passed: 1, no actual failed case; looks like a platform artifact.

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
tomal-majumder: Updated!

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
tomal-majumder: Updated!

tajila:
jenkins test sanity alinux64 jdknext

keithc-ca (approved):

tajila:
All the failures are known failures.

tajila (approved):

tajila:
@tomal-majumder Please open PRs for the other repos

tomal-majumder:
> @tomal-majumder Please open PRs for the other repos

Added the backports. FYI: @tajila

keithc-ca:
I don't see a link to a change for jdk26 (which may see one more release).

tomal-majumder:
> I don't see a link to a change for jdk26 (which may see one more release).

Changes were backported to jdk26.

PR #1241, opened 2026-08-07
Allow aliases in RestrictedSecurity constraints

KostasTsiounis:
When defining constraints for specific services allowed by specific providers, one can now use an alias instead of the original algorithm name.

A test case has, also, been added to verify said functionality.

---------

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -270,7 +270,7 @@ public final class RestrictedSecurity {
     public static boolean isServiceAllowed(Service service) {
         if (securityEnabled) {
             checkHashValues(false);
-            return restricts.isRestrictedServiceAllowed(service, true);
+            return restricts.isRestrictedServiceAllowed(service, true, null);
         }
         return true;
     }
@@ -279,12 +279,13 @@ public final class RestrictedSecurity {
      * Check if the service is allowed to be registered in restricted security mode.
      *
      * @param service the service to check
+     * @param aliases a list of aliases for the algorithm name of the service
      * @return true if the service is allowed to be registered
      */
-    public static boolean canServiceBeRegistered(Service service) {
+    public static boolean canServiceBeRegistered(Service service, List<String> aliases) {
         if (securityEnabled) {
             checkHashValues(false);
-            return restricts.isRestrictedServiceAllowed(service, false);
+            return restricts.isRestrictedServiceAllowed(service, false, aliases);
         }
         return true;
     }
@@ -858,9 +859,11 @@ public final class RestrictedSecurity {
          *
          * @param service   the Service to check
          * @param checkUse  should its attempted use be checked against the accepted
+         * @param aliases   a list of aliases for the algorithm name of the service
+         *                  (could be null if no aliases are available)
          * @return true if the Service is allowed
          */
-        boolean isRestrictedServiceAllowed(Service service, boolean checkUse) {
+        boolean isRestrictedServiceAllowed(Service service, boolean checkUse, List<String> aliases) {
             Provider provider = service.getProvider();
             String providerClassName = provider.getClass().getName();
 
@@ -914,11 +917,22 @@ public final class RestrictedSecurity {
                     continue constraints;
                 }
                 if (!isAsterisk(cAlgorithm) && !algorithm.equalsIgnoreCase(cAlgorithm)) {
-                    // The constraint doesn't apply to the service algorithm.
-                    if (debug != null) {
-                        debug.println("The constraint doesn't apply to the service algorithm.");
+                    // Check if the algorithm name in the constraint is an alias (case-insensitive).
+                    if ((aliases != null)
+                            && aliases.stream().anyMatch(cAlgorithm::equalsIgnoreCase)
+                    ) {
+                        // Fix the constraint to have the registered name.
+                        constraint.setAlgorithm(algorithm);
+                        if (debug != null) {
+                            debug.println("Constraint had an alias ('" + cAlgorithm + "'). Switching to registered ('" + algorithm + "').");
+                        }
+                    } else {
+                        // The constraint doesn't apply to the service algorithm.
+                        if (debug != null) {
+                            debug.println("The constraint doesn't apply to the service algorithm.");
+                        }
+                        continue constraints;
                     }
-                    continue constraints;
                 }
 
                 // For type and algorithm match, and attribute is not *.
@@ -2031,7 +2045,7 @@ public final class RestrictedSecurity {
      */
     private static final class Constraint {
         final String type;
-        final String algorithm;
+        String algorithm;
         final String attributes;
         final String acceptedUses;
 
@@ -2041,6 +2055,10 @@ public final class RestrictedSecurity {
             this.algorithm = algorithm;
             this.attributes = attributes;
             this.acceptedUses = acceptedUses;
+        }
+
+        void setAlgorithm(String algorithm) {
+            this.algorithm = algorithm;
         }
 
         @Override
--- a/closed/test/jdk/openj9/internal/security/TestConstraintsSuccess.java
+++ b/closed/test/jdk/openj9/internal/security/TestConstraintsSuccess.java
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2025, 2025 All Rights Reserved
+ * (c) Copyright IBM Corp. 2025, 2026 All Rights Reserved
  * ===========================================================================
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -87,6 +87,12 @@ public class TestConstraintsSuccess {
         // middle one having the correct attributes, successfully getting the
         // algorithm verifies that all constraints are checked.
         Signature.getInstance("SHA512withECDSA");
+
+        // Request algorithm whose alias has been specified in the constraint.
+        // Ask for it using both its registered name and an alias (with some
+        // changes in capitalization to check case-insensitivity).
+        MessageDigest.getInstance("SHA-384");
+        MessageDigest.getInstance("Sha384");
     }
 
     @Test
--- a/src/java.base/share/classes/java/security/Provider.java
+++ b/src/java.base/share/classes/java/security/Provider.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2025 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2026 All Rights Reserved
  * ===========================================================================
  */
 
@@ -1254,7 +1254,7 @@ public abstract class Provider extends Properties {
             throw new IllegalArgumentException
                     ("service.getProvider() must match this Provider object");
         }
-        if (!RestrictedSecurity.canServiceBeRegistered(s)) {
+        if (!RestrictedSecurity.canServiceBeRegistered(s, s.getAliases())) {
             // We're in restricted security mode which does not allow this service,
             // return without registering.
             return;
--- a/closed/test/jdk/openj9/internal/security/constraints-java.security
+++ b/closed/test/jdk/openj9/internal/security/constraints-java.security
@@ -21,7 +21,7 @@
 RestrictedSecurity.TestConstraints.Version.desc.name = Test Base Profile
 RestrictedSecurity.TestConstraints.Version.desc.default = false
 RestrictedSecurity.TestConstraints.Version.desc.fips = false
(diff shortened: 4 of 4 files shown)


src/java.base/share/classes/java/security/Provider.java
```diff
@@ -1254,10 +1254,13 @@ protected void putService(Service s) {
...
-            // We're in restricted security mode which does not allow this service,
-            // return without registering.
-            return;
+        if (RestrictedSecurity.isEnabled()) {
+            List<String> aliases = s.getAliases();
+            if (!RestrictedSecurity.canServiceBeRegistered(s, aliases)) {
+                // We're in restricted security mode which does not allow this service,
+                // return without registering.
+                return;
+            }
```
keithc-ca: It's not clear this change is needed; `canServiceBeRegistered()` can call `getAliases()` if need be.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -2043,6 +2052,10 @@ private static final class Constraint {
             this.acceptedUses = acceptedUses;
         }
 
+        public void setAlgorithm(String algorithm) {
```
keithc-ca: This method need not be public.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -860,7 +860,7 @@ private RestrictedSecurityProperties(String profileID, ProfileParser parser) {
          * @param checkUse  should its attempted use be checked against the accepted
          * @return true if the Service is allowed
          */
-        boolean isRestrictedServiceAllowed(Service service, boolean checkUse) {
+        boolean isRestrictedServiceAllowed(Service service, boolean checkUse, List<String> aliases) {
```
keithc-ca, line 866: Rather than expect callers to supply aliases, this method can retrieve them if necessary.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -914,11 +914,20 @@ boolean isRestrictedServiceAllowed(Service service, boolean checkUse) {
                     continue constraints;
                 }
                 if (!isAsterisk(cAlgorithm) && !algorithm.equalsIgnoreCase(cAlgorithm)) {
-                    // The constraint doesn't apply to the service algorithm.
-                    if (debug != null) {
-                        debug.println("The constraint doesn't apply to the service algorithm.");
+                    // Check if the algorithm name in the constraint is an alias.
+                    if ((aliases != null) && (aliases.contains(cAlgorithm))) {
```
keithc-ca: The containment check must be case-insensitive.

keithc-ca (changes requested):

src/java.base/share/classes/java/security/Provider.java
keithc-ca: Sorry, I see `getAliases()` is private.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -270,7 +270,7 @@ public static boolean isFIPS1402Enabled() {
     public static boolean isServiceAllowed(Service service) {
         if (securityEnabled) {
             checkHashValues(false);
-            return restricts.isRestrictedServiceAllowed(service, true);
+            return restricts.isRestrictedServiceAllowed(service, true, null);
```
keithc-ca, line 273: Aliases should be handled here as well.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -281,10 +281,10 @@ public static boolean isServiceAllowed(Service service) {
      * @param service the service to check
      * @return true if the service is allowed to be registered
      */
-    public static boolean canServiceBeRegistered(Service service) {
+    public static boolean canServiceBeRegistered(Service service, List<String> aliases) {
```
keithc-ca, line 285: Please ensure javadoc is updated for all methods with new signatures.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: After an alias is discovered in the registration, the algorithm name in the constraint is changed to the original. So, any subsequent checks (like this one) should not require or look for aliases.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Do you think we should allow all aliases regardless of case sensitivity, or precisely the ones registered with the service?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Changed to `private`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Updated for all signature changes.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: What guarantees that update occurs before this call?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: Why should aliases have to match exactly while the primary name is checked with `equalsIgnoreCase()`?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: When a provider is initialized and services are registered, the `canServiceBeRegistered()` that will fix the alias issue is called. The `isServiceAllowed()` will only be called when someone requests an algorithm and thus a specific service.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: That is a good point. I will change it to case-insensitive to match the primary name lookup.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: I was thinking package-private:
```java
        void setAlgorithm(String algorithm) {
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -914,11 +916,23 @@ boolean isRestrictedServiceAllowed(Service service, boolean checkUse) {
                     continue constraints;
                 }
                 if (!isAsterisk(cAlgorithm) && !algorithm.equalsIgnoreCase(cAlgorithm)) {
-                    // The constraint doesn't apply to the service algorithm.
-                    if (debug != null) {
-                        debug.println("The constraint doesn't apply to the service algorithm.");
+                    // Check if the algorithm name in the constraint is an alias (case-insensitive).
+                    if ((aliases != null)
+                        && aliases.stream().anyMatch(e -> e.equalsIgnoreCase(cAlgorithm))
+                    ) {
```
keithc-ca: No trailing spaces, please. Line 921 should be indented four more spaces and can be simplified:
```java
                    if ((aliases != null)
                            && aliases.stream().anyMatch(cAlgorithm::equalsIgnoreCase)
                    ) {
```

closed/test/jdk/openj9/internal/security/TestConstraintsSuccess.java
```diff
@@ -87,6 +87,11 @@ private static void getInstances() throws Exception {
         // middle one having the correct attributes, successfully getting the
         // algorithm verifies that all constraints are checked.
         Signature.getInstance("SHA512withECDSA");
+
+        // Request algorithm whose alias has been specified in the constraint.
+        // Ask for it using both its registered name and the alias.
+        MessageDigest.getInstance("SHA-384");
+        MessageDigest.getInstance("SHA384");
```
keithc-ca: SHA348 is only one of the aliases (the others are OID.2.16.840.1.101.3.4.2.2 and 2.16.840.1.101.3.4.2.2): "the alias" should say "an alias".

Suggest both the constraint and this test should verify that matching is case-insensitive; say, in the profile:
```
    {MessageDigest, sha384, *, FullClassName:TestConstraintsSuccess}, \
```
and here
```
        MessageDigest.getInstance("Sha384");
```
(or some other pair of names that differ in case from each other and any aliases).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
keithc-ca: Ok, I think I've convinced myself that `isServiceAllowed()` can only be called after the constraints have been updated by calls to `canServiceBeRegistered()`.

src/java.base/share/classes/java/security/Provider.java
keithc-ca: I don't think the call to `isEnabled()` is worthwhile.
```java
        if (!RestrictedSecurity.canServiceBeRegistered(s, s.getAliases())) {
            // We're in restricted security mode which does not allow this service,
            // return without registering.
            return;
        }
```

keithc-ca (changes requested):

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Removed additional check.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java

closed/test/jdk/openj9/internal/security/TestConstraintsSuccess.java
KostasTsiounis: Changed both constraint and test to suggestion.

keithc-ca:
Please address the failure of the line endings check.

KostasTsiounis:
> Please address the failure of the line endings check.

Oh, sorry. I missed that. Fixed now.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -858,9 +859,10 @@ private RestrictedSecurityProperties(String profileID, ProfileParser parser) {
          *
          * @param service   the Service to check
          * @param checkUse  should its attempted use be checked against the accepted
+         * @param aliases   a list of aliases for the algorithm name of the service
```
keithc-ca, line 862: nit: This ought to clarify that `null` is permitted for the `aliases` parameter.

keithc-ca (changes requested):
Also, please squash.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Added disclaimer that the parameter could be `null`.

KostasTsiounis:
> Also, please squash.

Done.

keithc-ca:

keithc-ca:
Testing in https://openj9-jenkins.osuosl.org/job/Grinder/5337 encountered https://github.com/eclipse-openj9/openj9/issues/24525:
```
        Caused by: java.lang.UnsatisfiedLinkError: jdk/internal/misc/Unsafe.isFlatField(Ljava/lang/reflect/Field;)Z
        	at java.base/jdk.internal.misc.Unsafe.fieldLayout(Unsafe.java:6494)
        	at java.base/java.io.ObjectStreamClass$FieldReflector.<init>(ObjectStreamClass.java:1955)
        	at java.base/java.io.ObjectStreamClass.getReflector(ObjectStreamClass.java:2138)
        	at java.base/java.io.ObjectStreamClass.<init>(ObjectStreamClass.java:431)
```
Merging despite this.

keithc-ca (approved):

PR #1250, opened 2026-08-20
Fix java/lang/IO/IO.java failure on OpenJ9

babsingh:
Skip the RI-specific `AOT` and `CDS` logging options when running
`printlnNoParamsTest` on OpenJ9. OpenJ9 reports the unsupported
`-Xlog` options to `stderr`, causing the test's empty `stderr`
assertion to fail.

Related: https://github.com/eclipse-openj9/openj9/issues/23292

JDK27 backport: https://github.com/ibmruntimes/openj9-openjdk-jdk27/pull/25

---------

### DIFF
--- a/test/jdk/java/lang/IO/IO.java
+++ b/test/jdk/java/lang/IO/IO.java
@@ -21,6 +21,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2026, 2026 All Rights Reserved
+ * ===========================================================================
+ */
+
 import java.io.Writer;
 import java.lang.reflect.Method;
 import java.nio.file.Files;
@@ -29,6 +35,7 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.stream.Stream;
 
+import jdk.test.lib.Platform;
 import jdk.test.lib.process.OutputAnalyzer;
 import jdk.test.lib.process.ProcessTools;
 import org.junit.jupiter.api.Assumptions;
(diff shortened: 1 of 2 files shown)


babsingh:
25x grinder passed: https://openj9-jenkins.osuosl.org/job/Grinder/5349.

@keithc-ca Requesting your review.

test/jdk/java/lang/IO/IO.java
```diff
@@ -149,7 +155,15 @@ void main() {
                     }
                     """);
         }
-        var pb = ProcessTools.createTestJavaProcessBuilder("-Xlog:aot=off", "-Xlog:cds=off", file.toString());
+
+        ProcessBuilder pb;
+        if (System.getProperty("java.vm.name").contains("OpenJ9")) {
+            /* The AOT/CDS -Xlog options are unsupported by OpenJ9 and are reported to stderr. */
```
keithc-ca, line 162: Wouldn't it be better to update OpenJ9 to ignore (or understand) these options?

babsingh: I considered that, but these `-Xlog` AOT/CDS options are RI-specific, while OpenJ9 has separate AOT and SCC (CDS-equivalent) options. OpenJ9 also documents [non-GC `-Xlog` options as unsupported](https://eclipse.dev/openj9/docs/xlog). Silently accepting them could imply functionality OpenJ9 doesn't provide, with little compatibility benefit outside this test. I think keeping the adaptation in the test makes that distinction clearer.

keithc-ca: It's not clear either solution is ideal, but I can accept modifying the test.

test/jdk/java/lang/IO/IO.java
```diff
@@ -149,7 +155,15 @@ void main() {
                     }
                     """);
         }
-        var pb = ProcessTools.createTestJavaProcessBuilder("-Xlog:aot=off", "-Xlog:cds=off", file.toString());
+
+        ProcessBuilder pb;
+        if (System.getProperty("java.vm.name").contains("OpenJ9")) {
```
keithc-ca: Please update `Platform.isJ9()` so it is `public` allowing it to be used here rather than duplicating that condition.

babsingh: Updated.

keithc-ca:
The grinder linked above was for jdk27, but the changes are the same as in https://github.com/ibmruntimes/openj9-openjdk-jdk27/pull/25.

keithc-ca (approved):

PR #1251, opened 2026-08-20
Add OpenJCEPlusSemeruDefaults as the first security provider

taoliult:
Introduce `com.ibm.crypto.plus.provider.OpenJCEPlusSemeruDefaults` as the first provider in the provider list.

The provider uses a runtime-generated name in the form `OpenJCEPlusSemeruDefaults-XXXXXX`. The generated suffix discourages applications from coding directly to the provider name while still allowing the provider to be inserted early in the provider list.

This provider supports a staged transition toward broader OpenJCEPlus enablement by default, while keeping the initial change limited in scope and easier to validate.

### DIFF
--- a/src/java.base/share/conf/security/java.security
+++ b/src/java.base/share/conf/security/java.security
@@ -93,6 +93,7 @@
 #
 # List of providers and their preference orders (see above):
 #
+security.provider.tbd=com.ibm.crypto.plus.provider.OpenJCEPlusSemeruDefaults
 security.provider.tbd=SUN
 security.provider.tbd=SunRsaSign
 security.provider.tbd=SunEC


keithc-ca:
Please elaborate on the initialization sequence. I see that `OpenJCEPlusSemeruDefaults` uses `UUID.randomUUID()` which uses `java.security.SecureRandom` which, in turn, requires the security providers list. With restricted security, the hash values must be checked. `OpenJCEPlusSemeruDefaults` does not appear to supply hash algorithms, so which provider will offer SHA-256, for example? How does this work?

taoliult:
> Please elaborate on the initialization sequence. I see that `OpenJCEPlusSemeruDefaults` uses `UUID.randomUUID()` which uses `java.security.SecureRandom` which, in turn, requires the security providers list. With restricted security, the hash values must be checked. `OpenJCEPlusSemeruDefaults` does not appear to supply hash algorithms, so which provider will offer SHA-256, for example? How does this work?

@keithc-ca There was a `SecureRandom` implementation in `OpenJCEPlusSemeruDefaults` when I originally introduced `UUID.randomUUID()` to generate the provider-name suffix, but that `SecureRandom` implementation was later removed.

With the current implementation, `UUID.randomUUID()` will cause `SecureRandom` to keep walking the registered provider list and select the first provider that supplies a SecureRandom service.

Since the UUID is only used to generate a runtime-unique provider-name suffix, I will open a separate PR in the OpenJCEPlus repository to replace `UUID.randomUUID()` with the approach below to generate a six-character random suffix without involve any security providers.

```
    private static String generateRandomSuffix() {
        // XOR a AtomicLong counter with System.nanoTime() to produce a suffix that
        // is both guaranteed unique (counter) and unpredictable (nanoTime).
        // No JCA provider lookup involved.
        long value = suffixCounter.getAndIncrement() ^ System.nanoTime();
        String hex = Long.toHexString(value).toUpperCase();
        // Left-pad with zeros if shorter than 6, truncate to last 6 chars if longer.
        if (hex.length() < 6) {
            hex = "000000".substring(hex.length()) + hex;
        } else if (hex.length() > 6) {
            hex = hex.substring(hex.length() - 6);
        }
        return hex;
    }
```

keithc-ca:
Getting exactly six hex digits can be much simpler:
```java
        return String.format("%06X", value & 0xFFFFFFL);
```

taoliult:
@keithc-ca So, something like this:

```
private static String generateRandomSuffix() {
        // XOR a AtomicLong counter with System.nanoTime() to produce a suffix that
        // no two calls return the same suffix. No JCA provider lookup involved.
        long value = suffixCounter.getAndIncrement() ^ System.nanoTime();
        return String.format("%06X", value & 0xFFFFFFL);
    }
```

keithc-ca:
I think this should wait until the change to OpenJCEPlus discussed above is available.

taoliult:
> I think this should wait until the change to OpenJCEPlus discussed above is available.

Ok, I will create the PR in the OpenJCEPlus repository for this `generateRandomSuffix()` change.

taoliult:
@keithc-ca The OpenJCEPlus repo PR (https://github.com/ibmruntimes/OpenJCEPlus/pull/38) for updating generateRandomSuffix() has been merged. Please continue reviewing this PR.

keithc-ca:

keithc-ca:
Please create an equivalent change for jdk27 which can be used to help decide which of the test failures are specific to jdk28.

taoliult:
> Please create an equivalent change for jdk27 which can be used to help decide which of the test failures are specific to jdk28.

@keithc-ca The equivalent change for jdk27  created: https://github.com/ibmruntimes/openj9-openjdk-jdk27/pull/28

keithc-ca (approved):
Approved based on the test results for https://github.com/ibmruntimes/openj9-openjdk-jdk27/pull/28.

PR #1258, opened 2026-08-28
Update exclude list for OpenJCEPlusFIPS Profiles

Mohit-Rajbhar100698:
This change excludes the ML-KEM-768 test from the OpenJCEPlusFIPS profiles because the algorithm is currently not supported.

---------


keithc-ca:
Should these files be updated as well?
* ProblemList-OpenJCEPlusFIPS.FIPS140-3-Test-TLS-JKS.txt
* ProblemList-OpenJCEPlusFIPS.FIPS140-3-Test-TLS-PKCS12.txt

Or, are those files unused (and safe to remove)?

Mohit-Rajbhar100698:
> * ProblemList-OpenJCEPlusFIPS.FIPS140-3-Test-TLS-PKCS12.txt

Hi Keithc, 
we’re not sure what these files are used for. Longyou, who created them, has left, so we don’t have much context about them.

These files are also empty and don’t seem to have ever been updated with the FIPS problems list. So, we’re not sure if they are still needed or if we can delete them.

keithc-ca (approved):

keithc-ca:
Please create back-ports for jdk21, jdk17 and jdk11 so the lists remain identical across versions (even though the test does not exist in those versions).

PR #1267, opened 2026-09-04
Enable OpenJ9 JFR V2 testing in TestModularImage

tomal-majumder:
TestModularImage spawns each jlinked runtime via ProcessBuilder. -`XX:+EnableOpenJ9ExperimentalFlightRecording` must be passed explicitly to each spawned java invocation for JFR V2 to actually be exercised.

Issue: https://github.com/eclipse-openj9/openj9/issues/23708
Required to test this PR: https://github.com/eclipse-openj9/openj9/pull/24714

---------

### DIFF
--- a/test/jdk/jdk/jfr/jvm/TestModularImage.java
+++ b/test/jdk/jdk/jfr/jvm/TestModularImage.java
@@ -20,6 +20,13 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2026, 2026 All Rights Reserved
+ * ===========================================================================
+ */
+
 package jdk.jfr.jvm;
 
 import java.io.File;
@@ -89,7 +96,7 @@ public class TestModularImage {
     }
 
     private static void testCommandLineWithJFR(Path binPath) throws Exception {
-        var result = java(binPath, "-XX:StartFlightRecording", "--module", "hello.world/hello.Main");
+        var result = java(binPath, "-XX:+EnableOpenJ9ExperimentalFlightRecording", "-XX:StartFlightRecording", "--module", "hello.world/hello.Main");
         result.shouldNotContain(ERROR_LINE1);
         result.shouldNotContain(ERROR_LINE2);
         result.shouldContain(HELLO_WORLD);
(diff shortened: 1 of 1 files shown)


keithc-ca:
How does this make sense when support for JVF V2 is limited to Java 17 (see https://github.com/eclipse-openj9/openj9/blob/master/runtime/vm/jvminit.c#L4660)?

tomal-majumder:
> How does this make sense when support for JVF V2 is limited to Java 17 (see https://github.com/eclipse-openj9/openj9/blob/master/runtime/vm/jvminit.c#L4660)?

This test is currently gated to run only on JDK 17 in aqa-tests jdk_jfr_j9 group. That said, your concern is still
valid.

I can gate it behind a runtime version check instead, so the flag `-XX:+EnableOpenJ9ExperimentalFlightRecording` is only passed when actually running on JDK 17. Would that be reasonable way to address this?

```
    private static final boolean JFR_V2_SUPPORTED =
            Runtime.version().feature() == 17;
```

keithc-ca:
If you want to change the test for JDK 17, the place for that is https://github.com/ibmruntimes/openj9-openjdk-jdk17.

test/jdk/jdk/jfr/jvm/TestModularImage.java
```diff
@@ -89,7 +96,7 @@ public static void main(String[] args) throws Exception {
     }
 
     private static void testCommandLineWithJFR(Path binPath) throws Exception {
-        var result = java(binPath, "-XX:StartFlightRecording", "--module", "hello.world/hello.Main");
+        var result = java(binPath, "-XX:+EnableOpenJ9ExperimentalFlightRecording", "-XX:StartFlightRecording", "--module", "hello.world/hello.Main");
```
JasonFengJ9, line 99: This isn't necessary, JFRV2 test requirement (`-XX:+EnableOpenJ9ExperimentalFlightRecording -Xshare:off`) is enabled via https://github.com/adoptium/aqa-tests/blob/f08048fbb7e41f54a9016e0784427d41338277e8/openjdk/playlist.xml#L2332 and https://github.com/adoptium/aqa-tests/blob/f08048fbb7e41f54a9016e0784427d41338277e8/openjdk/playlist.xml#L2360

keithc-ca: The options in that `playlist.xml` file only apply to the test driver process, not the sub-processes that are launched by the test.

JasonFengJ9: `test/jdk/jdk/jfr/jvm/TestModularImage.java` has been excluded.
However other tests in same jfr folder run with `-XX:+EnableOpenJ9ExperimentalFlightRecording -Xshare:off` at https://openj9-jenkins.osuosl.org/job/Grinder/5399/consoleFull, those tests don't need change like this PR.

tomal-majumder:
> If you want to change the test for JDK 17, the place for that is https://github.com/ibmruntimes/openj9-openjdk-jdk17.

Makes sense. Closing this PR and I will land it in https://github.com/ibmruntimes/openj9-openjdk-jdk17.

PR #1225, opened 2026-06-17
Remove provider-specific condition from SecureRandom strong test

Mohit-Rajbhar100698:
This change removes the provider-specific condition from the
SecureRandom strong algorithm test. The test should run with
any security provider to avoid failures when a different
provider is configured.

Signed-off-by: Mohit Rajbhar <mohit.rajbhar@ibm.com>

PR #1268, opened 2026-09-08
Fix StrongAlgorithm Tests When OpenJCEPlus Is Unavailable

Mohit-Rajbhar100698:
Fix StrongAlgorithm test failures when tests run with a JDK that does not include the OpenJCEPlus module.

All StrongAlgorithm tests are now guarded with an
OpenJCEPlusFIPS availability check, since the FIPS SecureRandom strong algorithm is not available when OpenJCEPlus is absent.

Also added an additional test to verify the default SecureRandom value when OpenJCEPlus is not available. This also prevents the stream argument from being null.

Signed-off-by: Mohit Rajbhar [mohit.rajbhar@ibm.com](mailto:mohit.rajbhar@ibm.com)

---------

### DIFF
--- a/closed/test/jdk/openj9/internal/security/property-java.security
+++ b/closed/test/jdk/openj9/internal/security/property-java.security
@@ -649,3 +649,18 @@ RestrictedSecurity.Test-Profile-strongAlgorithms-MisspelledPropertyName.jce.prov
 RestrictedSecurity.Test-Profile-strongAlgorithms-MisspelledPropertyName.securerandom.provider = OpenJCEPlusFIPS
 RestrictedSecurity.Test-Profile-strongAlgorithms-MisspelledPropertyName.securerandom.algorithm = SHA512DRBG
 RestrictedSecurity.Test-Profile-strongAlgorithms-MisspelledPropertyName.securerandom.strongAlgorithmsWrong = SHA512DRBG:OpenJCEPlusFIPS
+
+#
+# Test-Profile-strongAlgorithms-Default
+# Test constraint - default JDK value of securerandom.strongAlgorithms
+#
+RestrictedSecurity.Test-Profile-strongAlgorithms-Default.desc.name = Test-Profile-strongAlgorithms-Default
+RestrictedSecurity.Test-Profile-strongAlgorithms-Default.desc.default = true
+RestrictedSecurity.Test-Profile-strongAlgorithms-Default.desc.fips = true
+RestrictedSecurity.Test-Profile-strongAlgorithms-Default.desc.hash = SHA256:26f0026f0806b09495eafcffa37140cc11e8a5c087bfdf546227cc2e4d2f393c
+RestrictedSecurity.Test-Profile-strongAlgorithms-Default.fips.mode = 140-3
+RestrictedSecurity.Test-Profile-strongAlgorithms-Default.jce.provider.1 = sun.security.provider.Sun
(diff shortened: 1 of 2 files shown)


closed/test/jdk/openj9/internal/security/property-java.security
```diff
@@ -649,3 +649,18 @@ RestrictedSecurity.Test-Profile-strongAlgorithms-MisspelledPropertyName.jce.prov
 RestrictedSecurity.Test-Profile-strongAlgorithms-MisspelledPropertyName.securerandom.provider = OpenJCEPlusFIPS
 RestrictedSecurity.Test-Profile-strongAlgorithms-MisspelledPropertyName.securerandom.algorithm = SHA512DRBG
 RestrictedSecurity.Test-Profile-strongAlgorithms-MisspelledPropertyName.securerandom.strongAlgorithmsWrong = SHA512DRBG:OpenJCEPlusFIPS
+
+#
+# Test-Profile-strongAlgorithms-Default
+# Test constraint - securerandom.strongAlgorithms default JDK value when OpenJCEPlus is absent.
```
keithc-ca: This comment is misleading: The test uses it regardless of whether `OpenJCEPlus` is present.

keithc-ca (changes requested):

closed/test/jdk/openj9/internal/security/property-java.security
Mohit-Rajbhar100698: Hi @keithc-ca 
Updated, can u review once.?

keithc-ca:
> prevents the stream argument from being null

Please explain this comment.

Mohit-Rajbhar100698:
> > prevents the stream argument from being null

Hi @keithc-ca , 

If all the Strong Algorithm tests are guarded with an `OpenJCEPlusFIPS` check, then when OpenJCEPlusFIPS is not present, the argument method https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/5b6b1a93d74131ccd5e0e0fb933ffeb97d8bc241/closed/test/jdk/openj9/internal/security/TestProperties.java#L205 will return null, which will eventually cause `TestProperties` to fail. 


In the first commit for this feature, I guarded all the tests with the OpenJCEPlusFIPS condition. However, we ran into this issue, so we removed the condition in this PR: [PR #1225](https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1225). 



On the other hand, running the tests without any condition causes failures when they are run on a JDK build where `OpenJCEPlus` is not present. 
To handle both cases, I added a test to print the default SecureRandom value. This avoids the failure when OpenJCEPlus is absent. 

In simple terms, I have guarded only the tests that specifically verify the FIPS-certified SecureRandom Strong Algorithm with an `OpenJCEPlusFIPS` check, similar to what is already done here: https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/5b6b1a93d74131ccd5e0e0fb933ffeb97d8bc241/closed/test/jdk/openj9/internal/security/TestProperties.java#L171 
 So this is how this prevent the null argument.

keithc-ca:
`Stream.builder().build()` returns an empty stream, not `null`. Where does the `null` originate?

Mohit-Rajbhar100698:
> `Stream.builder().build()` returns an empty stream, not `null`. Where does the `null` originate?

Yes, @keithc-ca  I mean the empty stream. I mistakenly thought it was null. 

When an empty stream is returned, the TestProperties test fails because it expects a non-empty stream.

keithc-ca:
Out of curiosity, do you know where the empty stream is mapped to null?

keithc-ca (approved):

PR #1272, opened 2026-09-09
Update CheckSecurityProvider tests

taoliult:
Update `CheckSecurityProvider` test to account for the new `OpenJCEPlusSemeruDefaults` provider added to java.security.

- `CheckSecurityProvider.java`: add the expected class name "`com.ibm.crypto.plus.provider.OpenJCEPlusSemeruDefaults`" before SUN in the ordered provider list, matching the position defined in java.security.

### DIFF
--- a/test/jdk/java/lang/SecurityManager/CheckSecurityProvider.java
+++ b/test/jdk/java/lang/SecurityManager/CheckSecurityProvider.java
@@ -61,6 +61,8 @@ public class CheckSecurityProvider {
         List<String> expected = new ArrayList<>();
 
         // NOTE: the ordering must match what's defined inside java.security
+        layer.findModule("openjceplus")
+            .ifPresent(m -> expected.add("com.ibm.crypto.plus.provider.OpenJCEPlusSemeruDefaults"));
         expected.add("sun.security.provider.Sun");
         expected.add("sun.security.rsa.SunRsaSign");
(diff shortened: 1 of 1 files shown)


keithc-ca (approved):

keithc-ca:
The description is misleading: This doesn't change `UnsupportedProvider` (which is only present in jdk11).

taoliult:
> The description is misleading: This doesn't change `UnsupportedProvider` (which is only present in jdk11).

Yes, `UnsupportedProvider.java` is only present in JDK 11. So, when this PR is backported to JDK 11, the commit will include the changes to `UnsupportedProvider.java`. That is why I kept the description of this change here as well.

Do you think we should remove this description from the current PR?

keithc-ca:
Only the back-port to jdk11 should mention that other test. To be clear, the commit message here is fine, but the summary line (which was copied into the merge commit message) is what is misleading. I should have noticed that before merging, but alas I did not.

taoliult:
> Only the back-port to jdk11 should mention that other test. To be clear, the commit message here is fine, but the summary line (which was copied into the merge commit message) is what is misleading. I should have noticed that before merging, but alas I did not.

Okay, I’ve updated the description. Does the PR title also need to remove UnsupportedProvider?

keithc-ca:
The title could be updated, but it's too late to have any effect on the commit history.

PR #1274, opened 2026-09-10
Port Vahalla Fix GetStaticFieldID call ordering in InlineWithJni test

JasonFengJ9:
Port Vahalla Fix GetStaticFieldID call ordering in InlineWithJni test

* https://github.com/ibmruntimes/openj9-openjdk-jdk.valuetypes/pull/44

This is expected to be merged back to Vahalla extension repo to restore the change lost during recent merging.

---------

### DIFF
--- a/test/hotspot/jtreg/runtime/valhalla/valuetypes/libValueWithJni.c
+++ b/test/hotspot/jtreg/runtime/valhalla/valuetypes/libValueWithJni.c
@@ -21,13 +21,19 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2026, 2026 All Rights Reserved
+ * ===========================================================================
+ */
+
 #include <jni.h>
 
 JNIEXPORT void JNICALL
 Java_runtime_valhalla_valuetypes_ValueWithJni_doJniMonitorEnter(JNIEnv *env, jobject obj) {
-    int ret = (*env)->MonitorEnter(env, obj);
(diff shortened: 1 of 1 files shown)


JasonFengJ9:
FYI @hangshao0

keithc-ca:
I never fully understood why https://github.com/ibmruntimes/openj9-openjdk-jdk.valuetypes/pull/44 was appropriate. Please help.

JasonFengJ9:
As per https://github.com/eclipse-openj9/openj9/issues/24013#issuecomment-4671578546

_According to my understanding this is a control-flow issue where MonitorEnter correctly sets a pending IdentityException for value types, but the native code continues execution without handling the exception. As a result, subsequent JNI calls like GetStaticFieldID return NULL due to the pending exception state, which prevents the flow from safely proceeding and from updating returnValue to -1._

_With value type, MonitorEnter() could result in an exception. The native code should really check for the return value/exception before moving on. Unfortunately I don't see the test code doing that._

_In the_ [spec](https://docs.oracle.com/en/java/javase/26/docs/specs/jni/design.html#exception-handling);
_After an exception has been raised, the native code must first clear the exception before making other JNI calls. I don't see the following calls like GetStaticFieldID()/SetStaticIntField)() listed as safe JNIs to call when there is an exception._

This change is a workaroud for Valhalla  `MonitorEnter()` that could result an exception.

@hangshao0 correct me if my understand is wrong.

keithc-ca (approved):
Thank you for the explanation.

PR #1275, opened 2026-09-18
Adjust StrictFieldNoEarlyLarval test error to match OpenJ9

theresa-m:
This message has existed in OpenJ9 prior to the existance of strict fields, adjust the test rather then changing OpenJ9.

Related: https://github.com/eclipse-openj9/openj9/issues/24731

---------
- [X] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).


theresa-m:
@hangshao0 can you please review?

test/hotspot/jtreg/runtime/valhalla/valuetypes/verifier/StrictInstanceFieldsTest.java
```diff
@@ -193,7 +193,7 @@ public static void main(String[] args) throws Exception {
 
         negativeTest(LateOrdinaryStrictFrame.class, "Inconsistent stackmap frames at branch target");
 
-        negativeTest(StrictFieldNoEarlyLarval.class, "Constructor must call super() or this() before return");
+        negativeTest(StrictFieldNoEarlyLarval.class, "<init> does not call this <init> or super <init>");
```
keithc-ca, line 196: This change should be accompanied by an IBM copyright notice.

The message in OpenJ9 could be version-specific.

keithc-ca (changes requested):

test/hotspot/jtreg/runtime/valhalla/valuetypes/verifier/StrictInstanceFieldsTest.java
theresa-m: There is already an IBM copyright header in this file.

theresa-m: I'm not sure if I understand what you mean by the second comment but this error message has been in OpenJ9 since its initial contribution.

keithc-ca: Yes, I now see the IBM copyright notice.

If the message text is expected to change for Java 28, we could introduce `J9NLS_BCV_ERR_INIT_NOT_CALL_INIT28` which has the new text. 
```c
#if JAVA_SPEC_VERSION >= 28
					errorType = J9NLS_BCV_ERR_INIT_NOT_CALL_INIT28__ID;
#else /* JAVA_SPEC_VERSION >= 28 */
					errorType = J9NLS_BCV_ERR_INIT_NOT_CALL_INIT__ID;
#endif /* JAVA_SPEC_VERSION >= 28 */
```
I agree the OpenJ9 message is technically better (it isn't, and shouldn't be, specific to the Java language as it's a JVM-level issue), so I won't oppose adjusting the test, even though it increases the likelihood of future merge conflicts.

hangshao0 (approved):

test/hotspot/jtreg/runtime/valhalla/valuetypes/verifier/StrictInstanceFieldsTest.java
theresa-m: The message is not new to 28, OpenJ9 has had a different message for this scenario in previous versions. That's why I was thinking it made more sense to adjust the test instead of change an existing message.

PR #1277, opened 2026-09-22
Update the test expectation in NPEInPreviewTest to match the correct behaviour produced by OpenJ9

### LINKED ISSUE #24618: Valhalla runtime/valhalla/inlinetypes/NPEInPreviewTest.java
```
19:20:42  java.lang.RuntimeException: assertEquals expected: Cannot assign field "val" because "test" is null or "val" is a null restricted field and there's an attempt to store null in it but was: Cannot assign field "val" because "test" is null
19:20:42  	at jdk.test.lib.Asserts.fail(Asserts.java:715)
19:20:42  	at jdk.test.lib.Asserts.assertEquals(Asserts.java:208)
19:20:42  	at jdk.test.lib.Asserts.assertEquals(Asserts.java:195)
19:20:42  	at NPEInPreviewTest.testNullRestrictedFieldError(NPEInPreviewTest.java:86)
19:20:42  	at NPEInPreviewTest.main(NPEInPreviewTest.java:187)
19:20:42  	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
19:20:42  	at java.base/java.lang.reflect.Method.invoke(Method.java:589)
19:20:42  	at com.sun.javatest.regtest.agent.MainMethodHelper.executeModernMainClass(MainMethodHelper.java:56)
19:20:42  	at com.sun.javatest.regtest.agent.MainWrapper$MainTask.run(MainWrapper.java:140)
19:20:42  	at java.base/java.lang.Thread.run(Thread.java:1538)
```
https://hyc-runtimes-jenkins.swg-devops.com/view/Valhalla%20Tests/job/Test_JDKnext_valhalla_openjdk_x86-64_linux/209/

konark2010:
Update the test expectation to match the correct behaviour produced by OpenJ9 for c1 mode and ArrayStoreException and expected message for testNullRestrictedNotFlatArrayError().

This PR is dependent with the changes in this https://github.com/eclipse-openj9/openj9/pull/24789 so please merge them toghether.

Fixes: https://github.com/eclipse-openj9/openj9/issues/24618


konark2010:
@hangshao0 created a PR in this repo, as asked in https://github.com/ibmruntimes/openj9-openjdk-jdk.valuetypes/pull/55#issuecomment-5623505824

test/hotspot/jtreg/runtime/valhalla/valuetypes/NPEInPreviewTest.java
```diff
@@ -111,10 +117,11 @@ static void testActualNullFieldError() {
         } catch (NullPointerException npe) {
             String message = npe.getMessage();
             System.out.println("*** " + message);
+            //OpenJ9 does not have the c1 mode, so we can just check for the c1ExpectedMessage.
```
keithc-ca: Please follow the convention here with a space after `//`, and spell C1 like on line 110.

test/hotspot/jtreg/runtime/valhalla/valuetypes/NPEInPreviewTest.java
```diff
@@ -111,10 +117,11 @@ static void testActualNullFieldError() {
...
             if (c1Mode) {
                 Asserts.assertEquals(c1ExpectedMessage, message);
             } else {
-                Asserts.assertEquals(expectedMessage, message);
+                Asserts.assertEquals(c1ExpectedMessage, message);
             }
```
keithc-ca: Both branches of the `if` are the same:
```java
            Asserts.assertEquals(c1ExpectedMessage, message);
```

test/hotspot/jtreg/runtime/valhalla/valuetypes/NPEInPreviewTest.java
```diff
@@ -155,24 +162,24 @@ static void testNullRestrictedArrayError() {
         try {
             var a = ValueClass.newNullRestrictedAtomicArray(MyValue.class, 10, new MyValue());
             a[4] = null;
-        } catch (NullPointerException npe) {
-            String message = npe.getMessage();
+        } catch (NullPointerException | ArrayStoreException e) {
```
keithc-ca: Please list exceptions alphabetically:
```java
        } catch (ArrayStoreException | NullPointerException e) {
```

test/hotspot/jtreg/runtime/valhalla/valuetypes/NPEInPreviewTest.java
```diff
@@ -155,24 +162,24 @@ static void testNullRestrictedArrayError() {
...
+            String message = e.getMessage();
             System.out.println("*** " + message);
             if (c1Mode) {
-                Asserts.assertEquals(c1ExpectedMessage, message);
+                Asserts.assertEquals(expectedMessage, message);
             } else {
                 Asserts.assertEquals(expectedMessage, message);
             }
```
keithc-ca: Branches are the same:
```java
            Asserts.assertEquals(expectedMessage, message);
```

test/hotspot/jtreg/runtime/valhalla/valuetypes/NPEInPreviewTest.java
```diff
@@ -155,24 +162,24 @@ static void testNullRestrictedArrayError() {
...
 
     static void testNullRestrictedNotFlatArrayError() {
-        String expectedMessage = "Cannot store to object array because \"a\" is null or is a null-free array and there's an attempt to store null in it";
+        String expectedMessage = "Cannot store null in a null-restricted array";
         try {
             var a = ValueClass.newNullRestrictedAtomicArray(NotFlatValue.class, 10, new NotFlatValue());
             a[4] = null;
-        } catch (NullPointerException npe) {
-            String message = npe.getMessage();
+        } catch (NullPointerException | ArrayStoreException e) {
```
keithc-ca: Sort, please:
```java
        } catch (ArrayStoreException | NullPointerException e) {
```

keithc-ca (changes requested):
A smaller change would remove the `@run ... c1` line, eliminating the need for some other changes.

konark2010:
> A smaller change would remove the `@run ... c1` line, eliminating the need for some other changes.

shall I completely remove the c1 logic after that in all the tests, including the c1 boolean, and c1ExpectedMessage?

keithc-ca:
No, in the interest of keeping the change small relative to upstream, leave the c1 logic alone.

Looking again, it's the `c1` line we want to keep, as the messages from OpenJ9 match those; other modes should not be used.

test/hotspot/jtreg/runtime/valhalla/valuetypes/NPEInPreviewTest.java
```diff
@@ -111,11 +117,8 @@ static void testActualNullFieldError() {
         } catch (NullPointerException npe) {
             String message = npe.getMessage();
             System.out.println("*** " + message);
-            if (c1Mode) {
-                Asserts.assertEquals(c1ExpectedMessage, message);
-            } else {
-                Asserts.assertEquals(expectedMessage, message);
-            }
+            // OpenJ9 does not have the C1 mode, so we can just check for the c1ExpectedMessage.
```
keithc-ca, line 120: This seems self-contradictory: If OpenJ9 doesn't have a C1 mode, why would a c1 message be expected?

konark2010: I agree may be I can add a comment like this if it works for you: "// c1ExpectedMessage more precisely describes the failure in OpenJ9."

test/hotspot/jtreg/runtime/valhalla/valuetypes/NPEInPreviewTest.java
```diff
@@ -102,7 +108,7 @@ static void testNullRestrictedFieldStoredInNullError() {
     static void testActualNullFieldError() {
         String expectedMessage = "Cannot assign field \"nullVal\" because \"self\" is null";
         // In C1 Xcomp mode, the field or klass isn't resolved, so no idea which this is.
-        String c1ExpectedMessage = "Cannot assign field \"nullVal\" because \"self\" is null or \"nullVal\" is a null " +
+        String c1ExpectedMessage = "Cannot assign field \"nullVal\" because \"self\" is null or \"nullVal\" is a null-" +
```
keithc-ca: This change is not needed because `c1ExpectedMessage` is not used below.

test/hotspot/jtreg/runtime/valhalla/valuetypes/NPEInPreviewTest.java
```diff
@@ -155,24 +157,20 @@ static void testNullRestrictedArrayError() {
         try {
             var a = ValueClass.newNullRestrictedAtomicArray(MyValue.class, 10, new MyValue());
             a[4] = null;
-        } catch (NullPointerException npe) {
-            String message = npe.getMessage();
+        } catch (ArrayStoreException | NullPointerException e) {
```
keithc-ca: Please explain why `ArrayStoreException` must also be caught here and on line 172. Barring a good reason, OpenJ9 should throw the same exception class as the reference implementation.

keithc-ca (changes requested):

test/hotspot/jtreg/runtime/valhalla/valuetypes/NPEInPreviewTest.java
hangshao0: > Please explain why ArrayStoreException must also be caught here

This is from the inconsistent exception in
https://github.com/eclipse-openj9/openj9/blob/a6e82694d59dc5bbaf97093c0f862f8a0ec19e65/runtime/codert_vm/cnathelp.cpp#L1066
and
https://github.com/eclipse-openj9/openj9/blob/a6e82694d59dc5bbaf97093c0f862f8a0ec19e65/runtime/vm/BytecodeInterpreter.hpp#L7338

They are both for the same operation like `nullRestrictedArray[1] = null;`. which should throw the same exception. This needs and should be fixed by https://github.com/eclipse-openj9/openj9/pull/24789.

Regarding whether it should be ASE or NPE in this case, the draft [JEP](https://openjdk.org/jeps/8316779) says ASE:
`a2[1] = null; // ArrayStoreException`
However, the RI's behaviour is throwing NPE.

I had a vague impression that the JEP was saying NPE and it is changed to the current ASE, If my memory is correct, I guess likely the RI's behaviour will be changed to ASE.

konark2010: The ArrayStoreException catch is needed because OpenJ9 currently has inconsistent behavior for the same nullRestrictedArray[1] = null. 

I agree that OpenJ9 should match the RI. The current RI throws NullPointerException, although draft JEP 8316779 specifies ArrayStoreException for this case. The additional catch is therefore to account for the current OpenJ9 behavior while this discrepancy is being resolved.

keithc-ca: Thank you both for the explanation. `ArrayStoreException`, as proposed in the JEP seems more appropriate, I expect the RI to adapt.

